PR #14, opened 2015-04-01
 Subdirectory asset searching

JamzTheMan:
o Fixes FoW bug that overzealously resets FoW for all tokens. See http://forums.rptools.net/viewtopic.php?f=86&t=25833

o New Feature: Subdirectory searching. Checkbox toggles subdirectory
searches. 
  Notes: 
    - Currently AppConstants sets search result max to 1000.
    - File search is multi-threaded using SwingWorkers, please review
and test! Multi-threading is always tricky but I think I got it right.

### DIFF
--- a/maptool/src/main/java/net/rptools/maptool/client/AppConstants.java
+++ b/maptool/src/main/java/net/rptools/maptool/client/AppConstants.java
@@ -34,7 +34,8 @@ public class AppConstants {
 	public static final ImageBorder PURPLE_BORDER = new ImageBorder("net/rptools/maptool/client/image/border/purple");
 	public static final ImageBorder FOW_TOOLS_BORDER = new ImageBorder("net/rptools/maptool/client/image/border/fowtools");
 	public static final int NOTE_PORTRAIT_SIZE = 200;
-
+	public static final int ASSET_SEARCH_LIMIT = 1000;
+	
 	public static final FilenameFilter IMAGE_FILE_FILTER = new FilenameFilter() {
 		public boolean accept(File dir, String name) {
 			name = name.toLowerCase();
--- a/maptool/src/main/java/net/rptools/maptool/client/ui/assetpanel/AssetPanel.java
+++ b/maptool/src/main/java/net/rptools/maptool/client/ui/assetpanel/AssetPanel.java
@@ -40,6 +40,7 @@ import net.rptools.lib.swing.SelectionListener;
 import net.rptools.lib.swing.preference.SplitPanePreferences;
 import net.rptools.lib.swing.preference.TreePreferences;
 import net.rptools.maptool.client.AppConstants;
+import net.rptools.maptool.language.I18N;
 import net.rptools.maptool.model.Asset;
 
 public class AssetPanel extends JComponent {
@@ -137,7 +138,7 @@ public class AssetPanel extends JComponent {
 		top.add(BorderLayout.CENTER, getFilterTextField());
 
 		panel.add(BorderLayout.NORTH, top);
-//		panel.add(BorderLayout.SOUTH, getGlobalSearchField());
+		panel.add(BorderLayout.SOUTH, getGlobalSearchField());
 
 		return panel;
 	}
@@ -179,23 +180,43 @@ public class AssetPanel extends JComponent {
 	}
 
 	/**
-	 * Returns a checkbox that indicates whether the filter field applies to <i>all</i> images in all libraries or just
-	 * the currently selected image directory. Currently not implemented.
+	 * Returns a checkbox that indicates whether the filter field applies to <i>all</i> images in all libraries or just the currently selected image directory.
+	 * Currently not implemented.
 	 * 
 	 * @return the checkbox component
 	 */
-	public JCheckBox getGlobalSearchField() {
-		if (globalSearchField == null) {
-			globalSearchField = new JCheckBox("Search across all directories?", false);
-			globalSearchField.addActionListener(new ActionListener() {
-				public void actionPerformed(ActionEvent ev) {
+	private JCheckBox getGlobalSearchField()
+	{
+		if (globalSearchField == null)
+		{
+			globalSearchField = new JCheckBox(I18N.getText("panel.Asset.ImageModel.checkbox.searchSubDir1"), false);
+			globalSearchField.addActionListener(new ActionListener()
+			{
+				public void actionPerformed(ActionEvent ev)
+				{
 					updateFilter();
 				}
 			});
 		}
 		return globalSearchField;
 	}
-
+	
+	public void updateGlobalSearchLabel(int listSize)
+	{
+		if(getGlobalSearchField().isSelected())
+		{
+			globalSearchField.setText(
+					I18N.getText("panel.Asset.ImageModel.checkbox.searchSubDir1") 
+					+ " (" + listSize + "/" + AppConstants.ASSET_SEARCH_LIMIT + " " 
+					+ I18N.getText("panel.Asset.ImageModel.checkbox.searchSubDir2") + ")");
+		} else {
+			globalSearchField.setText(I18N.getText("panel.Asset.ImageModel.checkbox.searchSubDir1"));
+		}
+		
+		imagePanel.revalidate();
+		imagePanel.repaint();
+	}
+	
 	private synchronized void updateFilter() {
 		if (updateFilterTimer == null) {
 			updateFilterTimer = new Timer(500, new ActionListener() {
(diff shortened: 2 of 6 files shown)


rkathey:
Is the sub direstory seraching for the asset library?

Azhrei:
Sounds like it.  If it is, I’ll be curious to see the code.  Without rewriting a big part of the AssetManager (I think that was the class) it would be difficult to do since the cached information isn’t organized hierarchically in MT.

 <http://www.eeconsulting.net/>	
Edwards & Edwards Consulting, LLC <http://www.eeconsulting.net/>
Frank J. Edwards
sales@eec.com
Voice: +1 813.406.0604 (rings both office and mobile)

JamzTheMan:
It is for the image asset panel. I didn't rewrite the asset manager for that reason and I didn't use the cache. It does a recursive subdir listing directly.


That's why I put it out there now, so you guys can toke a look.  I didn't see any advantage of using the cache. Any gotcha you guys can think of?


-Jamz






Sent from Surface





From: Frank Edwards
Sent: ‎Wednesday‎, ‎April‎ ‎1‎, ‎2015 ‎10‎:‎20‎ ‎AM
To: RPTools/maptool
Cc: Lee Foster




Sounds like it. If it is, I’ll be curious to see the code. Without rewriting a big part of the AssetManager (I think that was the class) it would be difficult to do since the cached information isn’t organized hierarchically in MT.

<http://www.eeconsulting.net/> 
Edwards & Edwards Consulting, LLC <http://www.eeconsulting.net/>
Frank J. Edwards
sales@eec.com
Voice: +1 813.406.0604 (rings both office and mobile)

Azhrei:
Only a couple.

I have ~40k images in my asset library.  It seems like a scan is going to be slow (although doing it in a separate thread helps somewhat).  But what about people who use Dropbox or a USB stick for their library?  I think performance is going to be atrocious. :(

I'd rather see the asset class that caches this stuff changed to support additional information.  I think right now the key is the GUID (i.e. the SHA256 checksum of the image) and the value is the Asset, IIRC.  The GUID has to remain since it's how everything else in the program converts GUIDs to images.  But how much work would it be for the value to be an object consisting of image, pathname, image size, and image timestamp?  Whenever the existing code goes looking up the value there would need to be an additional method call to get the image.  Changing all existing code would be almost a global-search-and-replace.

Then the image loading would need to change to store the other info.  Once that stuff is stored, building a list sorted by pathname would make it easy to do subdirectory searches.  And the extra data could be used in the tooltips for the images.

Hm.  Maybe the relationship should be parent-child?  So when a directory is found that contains images, a "directory node" is created and all images within the directory are contained with it, including any subdirectory nodes. The directory name could become a tag and the user could associate other tags with images?  Then selecting a tag brings up the corresponding list of images?  This would mean a different "directory node" for each tag, where tags are automatically created from the directory names during the scan.

Sorry Jamz.  This is called "scope creep". ;-)  It's just that doing this right is a fair amount of work, but it's likely better to spend the time than it is to do it over _again_ in the future.  Has this been discussed much on the forums?  What would be the users' preference for organizing and displaying asset images?

JamzTheMan:
You lost me a little at the end there. I recommend trying it out first. SwingWorker can spawn up to 10 threads I believe, with my one folder I have 7.54 gb, 46,758 files over 1,247 folders.


It was taking only a few seconds and it was quicker on successive searches as it looked like Java did some internal caching.


We already have a set method for organizing assets in local repo? So not sure why that would be redone at this point?




As a user, using this to find a 'bed’ or 'crates’ was fast and seemed to work just fine.


And since it's coded already, it would be the same work to code now or recode later. 😊. Better to wait for asset classes to be updated with new types first like audio, etc.


Also, just fyi, the current search does not use the cache, so this just expands it to subdirectories.


Regarding dropbox/usb, sure, it will be slower but so would loading those assets and changing directories as that forces a rescan anyway. And drop box is cached locally to a hard drive? Maybe we are talking about 2 different things?





Sent from Surface





From: Frank Edwards
Sent: ‎Wednesday‎, ‎April‎ ‎1‎, ‎2015 ‎5‎:‎41‎ ‎PM
To: RPTools/maptool
Cc: Lee Foster





Only a couple.

I have ~40k images in my asset library. It seems like a scan is going to be slow (although doing it in a separate thread helps somewhat). But what about people who use Dropbox or a USB stick for their library? I think performance is going to be atrocious. :(

I'd rather see the asset class that caches this stuff changed to support additional information. I think right now the key is the GUID (i.e. the SHA256 checksum of the image) and the value is the Asset, IIRC. The GUID has to remain since it's how everything else in the program converts GUIDs to images. But how much work would it be for the value to be an object consisting of image, pathname, image size, and image timestamp? Whenever the existing code goes looking up the value there would need to be an additional method call to get the image. Changing all existing code would be almost a global-search-and-replace.

Then the image loading would need to change to store the other info. Once that stuff is stored, building a list sorted by pathname would make it easy to do subdirectory searches. And the extra data could be used in the tooltips for the images.

Hm. Maybe the relationship should be parent-child? So when a directory is found that contains images, a "directory node" is created and all images within the directory are contained with it, including any subdirectory nodes. The directory name could become a tag and the user could associate other tags with images? Then selecting a tag brings up the corresponding list of images? This would mean a different "directory node" for each tag, where tags are automatically created from the directory names during the scan.

Sorry Jamz. This is called "scope creep". ;-) It's just that doing this right is a fair amount of work, but it's likely better to spend the time than it is to do it over again in the future. Has this been discussed much on the forums? What would be the users' preference for organizing and displaying asset images?

Azhrei:
I thought there was some checksum calculating going on, but it wouldn't be possible to do that on 7.5GB of data in "a few seconds" (unless you're on an 8-core machine and all of the files are cached in RAM already by the OS!), so we must be talking about different things.

I'll look at the patch source code sometime. :)

JamzTheMan:
nope 4 core 2.3ghz q6600 processor on a 5400 rpm 500gb drive, so pretty old by current standards.


Simple File class, get fileList, iterate through looking for a .contains(String). Pretty standard stuff. Doing it in a SwingWorker allowed me to update the label with results (eg 56/1000) and asset images populate as it finds them while GUI remains responsive. As it finds each dir (as a dir is returned with all files in list) it recursively calls itself as another SwingWorker. So a few hundred may spawn but memory seemed to stay tight and I cancel out early if I reach the limit.


I started with a limit of a thousand. just in case someone put, say an ‘a’ for search, the imagePanel gets a little sluggish with 20k images and who needs that many results? 😊


Let me know how it performs on a Mac some time when we get another preview build.






Sent from Surface





From: Frank Edwards
Sent: ‎Wednesday‎, ‎April‎ ‎1‎, ‎2015 ‎8‎:‎58‎ ‎PM
To: RPTools/maptool
Cc: Lee Foster





I thought there was some checksum calculating going on, but it wouldn't be possible to do that on 7.5GB of data in "a few seconds" (unless you're on an 8-core machine and all of the files are cached in RAM already by the OS!), so we must be talking about different things.

I'll look at the patch source code sometime. :)

Azhrei:
Yeah, I was looking at the Gradle build rules for OS X.  I want to tweak things a little bit (should start the launcher, not MT, for example), but it'll need to wait until next week.

JamzTheMan:
oh, Speaking of assets, got another question.  I have a patch that allows you adjust the imagePanel thumbnail size via mouseWheel. I upped the thumb size from 50x50 to something like 100x100 iirc so you got enough details for maps, etc. It didn't seem to bloat the cache to bad.




Question, if we do this, it would not reflect on older cache’s.  Should we gave a clear cache option in MT? should we put a date limit on cache clearing anything older than x months? Other ideas on managing cache? 


Of course we also have a thumb cache as well as an asset cache.






Sent from Surface





From: Frank Edwards
Sent: ‎Wednesday‎, ‎April‎ ‎1‎, ‎2015 ‎8‎:‎58‎ ‎PM
To: RPTools/maptool
Cc: Lee Foster





I thought there was some checksum calculating going on, but it wouldn't be possible to do that on 7.5GB of data in "a few seconds" (unless you're on an 8-core machine and all of the files are cached in RAM already by the OS!), so we must be talking about different things.

I'll look at the patch source code sometime. :)

JamzTheMan:
hmm, haven't tested the test build if it launches launcher or not… Always in eclipse…






Sent from Surface





From: Frank Edwards
Sent: ‎Wednesday‎, ‎April‎ ‎1‎, ‎2015 ‎9‎:‎25‎ ‎PM
To: RPTools/maptool
Cc: Lee Foster





Yeah, I was looking at the Gradle build rules for OS X. I want to tweak things a little bit (should start the launcher, not MT, for example), but it'll need to wait until next week.

PR #15, opened 2015-04-04
Quick new functions, getZoom() and setZoom(x)

Jaggeroth:
This is task 15.7 on RPTroll's googledoc and from this thread here:
http://forums.rptools.net/viewtopic.php?f=86&t=25824


Azhrei:
I see that "/bin/" was added to the .gitignore but I don't think that's effective (unless something has changed in Git).  The .gitignore file only ignores files, not directories; directories are always ignored by git.  Proof:  create and empty directory and run 'git status" and you'll notice the directory doesn't show up in the "unadded" list.

If you want to ignore /bin and everything in it, change the .gitignore to include "/bin/**" (note the two trailing asterisks).

http://git-scm.com/docs/gitignore

I'll let Craig merge this since he's the one doing all the work in building this stuff lately. :)

JamzTheMan:
I believe both work….



If the pattern ends with a slash, it is removed for the purpose of the following description, but it would only find a match with a directory. In other words, foo/ will match a directory foo and paths underneath it, but will not match a regular file or a symbolic link foo (this is consistent with the way how pathspec works in general in Git).







Sent from Surface





From: Frank Edwards
Sent: ‎Saturday‎, ‎April‎ ‎4‎, ‎2015 ‎10‎:‎16‎ ‎AM
To: RPTools/maptool





I see that "/bin/" was added to the .gitignore but I don't think that's effective (unless something has changed in Git). The .gitignore file only ignores files, not directories; directories are always ignored by git. Proof: create and empty directory and run 'git status" and you'll notice the directory doesn't show up in the "unadded" list.

If you want to ignore /bin and everything in it, change the .gitignore to include "/bin/**" (note the two trailing asterisks).

http://git-scm.com/docs/gitignore

I'll let Craig merge this since he's the one doing all the work in building this stuff lately. :)

Azhrei:
Well, what do you know -- you're right.  A directory in the .gitignore removes all paths underneath the directory:

> If the pattern ends with a slash, it is removed for the purpose of the following description, but it would only find a match with a directory. In other words, foo/ will match a directory foo and paths underneath it, but will not match a regular file or a symbolic link foo (this is consistent with the way how pathspec works in general in Git).

I remember having a problem with it using Git 1.7 (which one of my hosting providers uses) and I had to add the double-star at the end.  However, looking at the documentation for both 1.7 and 2.3, it's almost identical with respect to this issue, so either I was having a different problem or there was a bug in the 1.7 version I was using.  I did test it using 2.3.3 and it does work as the documentation says it should.

Thanks!

Jaggeroth:
Awesome, thanks

PR #18, opened 2015-04-23
Add get and setNotes and get and setGMNotes

Jaggeroth:
Probably more work to merge than to write ;)


DarkAlf:
You can squash the 2nd commit into the first by running "git rebase -i 76808e3" (this will open your default text editor) and replacing 'pick' with 'squash'. That way the commit log will be cleaner.

Jaggeroth:
Thanks for the advice, but as a Git noob, rebasing is still too scary for
me ;)

PR #20, opened 2015-04-26
Improve random suffix logic

AlexWerne:
The new random suffix code is guaranteed to run in constant time. It
is also easier to use since most of the logic is now encapsulated in
a separate class.

Possible future improvements:
- Refactor the different types of suffixes using one cohesive design
  (for example by using an interface).
- Add a way to mark suffixes as 'in-use' during initialization of
  RandomSuffixFactory. (Not a priority.)

### DIFF
--- a/maptool/src/main/java/net/rptools/maptool/client/MapToolUtil.java
+++ b/maptool/src/main/java/net/rptools/maptool/client/MapToolUtil.java
@@ -22,6 +22,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import net.rptools.maptool.client.utilities.RandomSuffixFactory;
 import net.rptools.maptool.model.Asset;
 import net.rptools.maptool.model.AssetManager;
 import net.rptools.maptool.model.Token;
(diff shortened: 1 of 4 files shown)


mjstraw:
## Relayed Codex/ChatGPT review — changes requested

This review was performed in Codex Chat on 2026-09-08 and is being relayed here at the repository owner's request. Reviewed head: `03ce03df643e2bbc6917e80bff91302a9ffd9b65` (`randomsuffix`).

### Findings

- **P1 — Duplicate token identifiers can still be generated.** The retry loop uses `nameCollision && gmNameCollision`. It should retry when either configured/applicable display field collides. As written, a collision in only one field is accepted, including when numbering is stored exclusively in the name or GM name.
- **P2 — The uniqueness test does not test generated suffixes.** The test indexes and updates `uniqueSuffixes[i - 1]` instead of `uniqueSuffixes[suffix - 1]`, so it can pass independently of the generated value and would not catch duplicate suffixes.
- **P2 — The static factory retains every token base name indefinitely.** A permutation and generator are retained for each distinct name while the factory is process-global. This creates unbounded retained state; scope it to a zone/campaign or reset it when that context changes.
- **Verification limitation:** Tests were not executed because this historical checkout requires Gradle 2.1 and the wrapper download was blocked in the review environment. `git diff --check` reported two trailing-whitespace errors.

**Disposition:** Request changes due to the P1 correctness issue.

**Next action:** A maintainer should decide whether this closed PR should be revived or ported; if so, address the P1 collision condition, correct the suffix-indexing test, bound the factory lifecycle, and rerun the historical test suite.

PR #22, opened 2015-05-03
Set view area

Jaggeroth:
New function to set the viewable area.

Thread here: http://forums.rptools.net/viewtopic.php?f=86&t=25961

cheers,

Jagged


DarkAlf:
Looks good in my opinion. The only thing I would change is to put the code that tries to cast the parameters into integers in a private method. It reduces the code duplication and makes it easier to update the code in the future.

Jaggeroth:
Good suggestion.  Done and pushed

PR #23, opened 2015-05-06
Use standard mathematics for token facing

DarkAlf:
Token facing now follows standard mathematical definitions using
degrees. In other words the facing is always between 0 degrees
(inclusive) and 360 degrees (exclusive). 0° indicates the token is
facing towards the east, 90° towards north, 180° towards west and
270° towards south.

There is a new macro function (getTokenAngle) that gives scripters
access to these values. The old macro getTokenFacing still returns
values from -179° to 180°.
http://forums.rptools.net/viewtopic.php?f=86&t=25826

### DIFF
--- a/maptool/src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
+++ b/maptool/src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
@@ -46,10 +46,19 @@ public class TokenPropertyFunctions extends AbstractFunction {
 	private static final TokenPropertyFunctions instance = new TokenPropertyFunctions();
 
 	private TokenPropertyFunctions() {
-		super(0, 4, "getPropertyNames", "getAllPropertyNames", "getPropertyNamesRaw", "hasProperty", "isNPC", "isPC", "setPC", "setNPC", "getLayer", "setLayer", "getSize", "setSize", "getOwners",
-				"isOwnedByAll", "isOwner", "resetProperty", "getProperty", "setProperty", "isPropertyEmpty", "getPropertyDefault", "sendToBack", "bringToFront", "getLibProperty", "setLibProperty",
-				"getLibPropertyNames", "setPropertyType", "getPropertyType", "getRawProperty", "getTokenFacing", "setTokenFacing", "removeTokenFacing", "getMatchingProperties",
-				"getMatchingLibProperties", "isSnapToGrid", "setOwner", "getTokenWidth", "getTokenHeight", "setTokenWidth", "setTokenHeight", "getTokenShape", "setTokenShape",
+		super(0, 4, "getPropertyNames", "getAllPropertyNames",
+				"getPropertyNamesRaw", "hasProperty", "isNPC", "isPC", "setPC",
+				"setNPC", "getLayer", "setLayer", "getSize", "setSize",
+				"getOwners", "isOwnedByAll", "isOwner", "resetProperty",
+				"getProperty", "setProperty", "isPropertyEmpty",
+				"getPropertyDefault", "sendToBack", "bringToFront",
+				"getLibProperty", "setLibProperty", "getLibPropertyNames",
+				"setPropertyType", "getPropertyType", "getRawProperty",
+				"getTokenFacing", "getTokenAngle", "setTokenFacing",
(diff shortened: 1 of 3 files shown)


Jaggeroth:
Hi,

I am a little confused about the changes, to be honest.  Are you changing
the stored value from a value between -179 and 180 to a value between 0 and
360? If so that seems very dangerous. Apologies if I am miss-reading that.

Could I also ask that we don't format lines to 80 characters?  I know Craig
has talked about changing the 200 character length, but if we do change it,
I hope its something we can all agree on. Mostly I would rather only have
read through code changes and not format changes.

cheers

DarkAlf:
Yeah, that's exactly what the code is doing. But you're right, it will break too much stuff. I'm more or less going to start over on this one.

When I asked if there was a style guide I was told there was none. So therefore I have been using the standard Java conventions/formats.

JamzTheMan:
In regards to code formatting, I propose we use something like:


https://github.com/diffplug/spotless


That way all code would format to a standard eclipse style but we can code in our own style.  Thoughts?








Sent from Windows Mail





From: David
Sent: ‎Wednesday‎, ‎May‎ ‎6‎, ‎2015 ‎7‎:‎45‎ ‎AM
To: RPTools/maptool




Hi,

I am a little confused about the changes, to be honest. Are you changing
the stored value from a value between -179 and 180 to a value between 0 and
360? If so that seems very dangerous. Apologies if I am miss-reading that.

Could I also ask that we don't format lines to 80 characters? I know Craig
has talked about changing the 200 character length, but if we do change it,
I hope its something we can all agree on. Mostly I would rather only have
read through code changes and not format changes.

cheers

JamzTheMan:
I’m confused, I thought you were leaving the original function alone and only adding a new function?






Sent from Windows Mail





From: DarkAlf
Sent: ‎Wednesday‎, ‎May‎ ‎6‎, ‎2015 ‎11‎:‎12‎ ‎AM
To: RPTools/maptool





Yeah, that's exactly what the code is doing. But you're right, it will break too much stuff. I'm more or less going to start over on this one.

When I asked if there was a style guide I was told there was none. So therefore I have been using the standard Java conventions/formats.

cwisniew:
I just had a brief look at that link and it looks like it might fit the
bill perfectly.
I know Azrhrei had an eclipse format already set up so I will try it with
that this weekend and see how it looks.

I guess we will also have to take the painful medicine of a lot of changes
for format, I guess its better to do it in one commit and get it over and
done with just like a bandaid... maybe...

Jaggeroth:
Well, glad you spotted it.

Regarding styles, we don't have a style guide (wish we could use something
like CheckStyle) but what we do have is here:
http://forums.rptools.net/viewtopic.php?f=43&t=16663

JamzTheMan:
Ya, I was looking at another Gradle plugin but it had issues with Java 8.


I was watching this one and they finally put a 1.0 release out but it looked decent and easy. Was going to test it myself but RL work has been Crazy, but I don’t have to tell you that. 😊


It may be best to do it all at once. Although I have a dozen or more files in flux right now so it’ll be painful for me probably to remerge but we gotta do it at some point.






Sent from Windows Mail





From: Craig Wisniewski
Sent: ‎Wednesday‎, ‎May‎ ‎6‎, ‎2015 ‎9‎:‎50‎ ‎PM
To: RPTools/maptool
Cc: Lee Foster




I just had a brief look at that link and it looks like it might fit the
bill perfectly.
I know Azrhrei had an eclipse format already set up so I will try it with
that this weekend and see how it looks.

I guess we will also have to take the painful medicine of a lot of changes
for format, I guess its better to do it in one commit and get it over and
done with just like a bandaid... maybe...

Jaggeroth:
Yeah, if we are going to do this, best to agree a date in advance so people
can get their work submitted and merged.

Incidentally, is there a reason why you didn't pick one of the well known
plugins like Checkstyle? Checkstyle can be called from Gradle.

DarkAlf:
By the way, while redoing this one I noticed that I had accidentally made Eclipse apply its auto-formatting somehow. That was the reason why so many lines had nothing but format changes. I usually try to only change formatting if I'm actually changing the code as well.

I tried undoing this to the extent it was possible for the next pull request but in the end I got kind of lazy so some formatting changes will probably have slipped through. :-/

PR #24, opened 2015-05-11
Refactor SightType and Light to use same shape code from Grid class.

Jaggeroth:
Hi,

A relatively straight forward fix to issue 32.3 (http://forums.rptools.net/viewtopic.php?f=86&t=25968) I think.

SightType and Light classes shared the same code to produce circle, square and cone vision shapes, but a bug fix had been applied to SightType and missed from Light.  So I have refactored the two classes to get the shapes from the Grid class, which reduces code duplication and makes it more readable imo.

cheers

### DIFF
--- a/maptool/src/main/java/net/rptools/maptool/model/Grid.java
+++ b/maptool/src/main/java/net/rptools/maptool/model/Grid.java
@@ -272,13 +272,13 @@ public abstract class Grid implements Cloneable {
 	
 	/**
 	 * Called by SightType and Light class to return a vision area
-	 * based upon a specified distance.
+	 * based upon a specified distance
 	 * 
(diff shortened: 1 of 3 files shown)


DarkAlf:
There's no need to use hyphens in javadoc comments to separate the parameter name from the description. The Javadoc tool will automatically apply the correct formatting when generating documentation. So if there already is one then it will end up looking like this:
javdoc

Another even more minor note is that the first line of a git commit message is usually thought of as a headline. In other words there shouldn't be a period at the end. :-)

Apart from those two very minor things I can't see anything that seems out of place.

Jaggeroth:
I didn't know that, thanks :)

PR #25, opened 2015-05-14
Token facing

DarkAlf:
I've split this pull request into two separate commits since the second one might be somewhat controversial. Please have a look and say what you think. (Read the commit messages for more info as to what they entail.)

---
This change is [Review on Reviewable](https://reviewable.io/reviews/rptools/maptool/25)

### DIFF
--- a/maptool/src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
+++ b/maptool/src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
@@ -555,11 +555,12 @@ public class TokenPropertyFunctions extends AbstractFunction {
 			}
 		}
 
+		/* Returns the facing in degrees between 0 and 359. */
 		if (functionName.equals("getTokenAngle")) {
 			checkNumberOfParameters(functionName, parameters, 0, 1);
 			Token token = getTokenFromParam(resolver, functionName, parameters, 0);
 			if (token.hasFacing()) {
-				return new BigDecimal(token.getAngleInDegrees());
+				return new BigDecimal(token.getFacingInDegrees());
 			} else {
 				return "";
 			}
--- a/maptool/src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
+++ b/maptool/src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
@@ -46,11 +46,17 @@ public class TokenPropertyFunctions extends AbstractFunction {
 	private static final TokenPropertyFunctions instance = new TokenPropertyFunctions();
 
 	private TokenPropertyFunctions() {
-		super(0, 4, "getPropertyNames", "getAllPropertyNames", "getPropertyNamesRaw", "hasProperty", "isNPC", "isPC", "setPC", "setNPC", "getLayer", "setLayer", "getSize", "setSize", "getOwners",
-				"isOwnedByAll", "isOwner", "resetProperty", "getProperty", "setProperty", "isPropertyEmpty", "getPropertyDefault", "sendToBack", "bringToFront", "getLibProperty", "setLibProperty",
-				"getLibPropertyNames", "setPropertyType", "getPropertyType", "getRawProperty", "getTokenFacing", "setTokenFacing", "removeTokenFacing", "getMatchingProperties",
-				"getMatchingLibProperties", "isSnapToGrid", "setOwner", "getTokenWidth", "getTokenHeight", "setTokenWidth", "setTokenHeight", "getTokenShape", "setTokenShape",
-				"getGMNotes", "setGMNotes", "getNotes", "setNotes");
+		super(0, 4, "getPropertyNames", "getAllPropertyNames", "getPropertyNamesRaw",
+				"hasProperty", "isNPC", "isPC", "setPC", "setNPC", "getLayer", "setLayer",
+				"getSize", "setSize", "getOwners", "isOwnedByAll", "isOwner", "resetProperty",
+				"getProperty", "setProperty", "isPropertyEmpty", "getPropertyDefault",
+				"sendToBack", "bringToFront", "getLibProperty", "setLibProperty",
+				"getLibPropertyNames", "setPropertyType", "getPropertyType", "getRawProperty",
+				"getTokenFacing", "getTokenAngle", "setTokenFacing", "removeTokenFacing",
+				"getMatchingProperties", "getMatchingLibProperties", "isSnapToGrid", "setOwner",
+				"getTokenWidth", "getTokenHeight", "setTokenWidth", "setTokenHeight",
+				"getTokenShape", "setTokenShape", "getGMNotes", "setGMNotes", "getNotes",
+				"setNotes");
 	}
 
 	public static TokenPropertyFunctions getInstance() {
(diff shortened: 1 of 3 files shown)


Jaggeroth:
Unfortunately this is still changes the value stored against the token, which breaks backwards compatibility.  I also don't think we should be adding translations to functions such as Token.getFacing() since they are called thousands of times during zone rendering, where we need to be as efficient as we reasonably can.

I took a picture of a map created in 1.3 and loaded it into your branch and the result was a load of tokens and stamps pointing in different directions.

Before with 1.3:
before

After, with this branch:
after

DarkAlf:
God damn it, I really need to remember that Maptool is lacking unit tests. ^^ There was a bug in a line of code that messed things up. I've pushed the fix to the repo. Delete your old copy of the branch and try again please. :-)

Once people have had a look at this and said what they think my plan is to go through all the places that call the deprecated getFacing and setFacing methods and replace them with the new versions.

Jaggeroth:
I am still confused as to what this branch is trying to achieve?

It seems to be changing the value stored in the token facing parameter and adding code to the set and get methods, which I think is a bad idea because it will add unnecessary computation into the rendering pipeline and break backwards compatibility.

I am still getting some tokens and stamps facing wrong directions when I import an old map and it will simply error if I try to import a new map into 1.3.  Adding getFacingInDegrees() and getFacingInRadians() is a great addition, but I am against changing the getFacing() function.

DarkAlf:
In short, the old facing API (please note that I'm using the term "API" somewhat loosely here) of the token class was very poor and needlessly increased the complexity for any developer working with it. That is the main issue I'm trying to solve with this pull request. There were several other things wrong with it such as storing the facing in an Integer object which I fix at the same time since it's unnecessary to leave them as they are while improving the other aspects of the facing API.

Since there are still pieces of code relying on the old API I don't change the "public contract" of the setFacing and getFacing methods. They still accept null values and return values in the range of -179 to 180. However, I mark them as deprecated which means that they shouldn't be used at all. This includes future code as well as present one. The reason I don't change all of the code that call the deprecated methods is in case there is something that breaks while updating the API and the internal representation of facing. That way in case something breaks it will be easier to determine if it has something to do with either the code changes in Token or the code changes in the classes that call set and getFacing.

I have updated the commit message to try and convey this slightly better. The message still probably needs some more work, however.

Azhrei:
If it breaks compatibility with save files, is this a good time to change it?  On the one hand, this is a major new version and backward compatibility was never a factor.  On the other hand, you may los a lot of testers with a gratuitous change that prevents them from loading their campaigns...

If you still think it's a good idea, look at readObject() in the Token class so you can convert the internal values as the token is read in.

DarkAlf:
Personally I would rather not break backwards compatibility if possible so I'll look into it and see what can be done. How does Maptool currently save tokens? (I'll have a look at readObject() as well.)

We don't necessarily need to merge the second commit at this moment, especially if it breaks something. It was more or less that while adding the macro function getTokenAngle I realized that the facing API was pretty. . . "wonky". Hence the second commit. (There are some things I would like to improve on the first commit if we merge it by itself, though.)

Jaggeroth:
If we are going to break backwards compatibility, then now is the time.  But I do feel we should only do so for new functionality, not to make life easier for us Java Developers ;)

I think the main problem for backwards compatibility, is changing the type of the token facing property from an int to a float, which causes a 1.3 client to stack trace when it attempts to de-serialise a map built with these changes.

Perhaps if you kept the facing variable as is, and added a second facingRadians variable?  The two could be kept in sync and then perhaps a 1.4 map could still be imported into 1.3?  Also, since the facing property would be unchanged the risk of unintended consequence might be lower.

DarkAlf:
There doesn't seem to be a readObject() method in the Token class.

It seems like Maptool saves its tokens by serializing into XML and then zipping the result into a file? And that's why it "gets confused" when trying to unserialize the old XML into the refactored Token class. Anyways I've made some changes in the amended commit that avoids the worst of the pitfalls associated with this. (Changes such as the NO_FACING constant previously being -1 and the copy constructor now copies the facing using the set and getFacing methods.) The only backwards compatibility this seems to break is that a token saved in 1.3 without a facing gets assigned the default float value of 0 and hence will face towards the east when loaded in 1.4. I think this is something most users can live with? (But please check that there are no edge cases I have missed.)

Being unable to load a 1.4 map/token in 1.3 is what I would call breaking forward compatibility. Personally I view that as the natural order of things and I also think it is likely that other changes in 1.4 also will break forward compatibility.

rkathey:
Normally a version change indicates compatibility break. As a user I expect
this. In fact, this has happened with a good number of minor version
changes as well. You just need to notify the community so they can save off
an old version in case something in the new version is a deal breaker.

Azhrei:
Hm, must be that readObject() is in the Zone class instead then.  I know there's one somewhere (!) that handles compatibility changes.  It can also be handled at the XML layer as the data is read in but it's best to avoid doing it there as the XStream library forces such things to be done in Strings prior to parsing and that sucks up memory big time. :(

Jaggeroth:
Just pulled down the latest, and unfortunately I am still getting "some" tokens changing facing.  Not sure why the door, shown below, suddenly opens.  I tested the 1.4 master branch, just in case there was something strange going on, but it behaves the same as 1.3b91.

Map in 1.3:

master

Map in 1.4 tokenfacing:

tokenfacing

DarkAlf:
Hmmm, does the door have a facing set before being loaded?

Jaggeroth:
I should be able to take a look at it tonight

Jaggeroth:
Hi, the map I used for testing can be found here: http://hero-id.co.uk/Maptool/testMap.rpmap

DarkAlf:
I've had a look at the map and the door does indeed not have a facing set before being loaded.

cwisniew:
If I read the comments correctly this will cause a problem with the existing facing of certain objects when an old map is loaded? If so I would prefer to leave this one for 1.4.1 where a few incompatibilities can be introduced (I want to leave 1.4 for a release that doesn't break compatibility but works with a recent java version for people to move to).

JamzTheMan:
I think we should close this Pull Request.  If DarkAlf comes back and wants to support his code we can revisit this.

But I agree with Jagged, I think it's more complicated than it needs to be.  I have since added a new simpler method in Token class for internal use:

	public Integer getFacingInDegrees() {
		if (facing == null)
			return 0;
		else
			return -(facing + 90);
	}

It's the same math I used in my older VBL Macros.  To keep compatibility we can add a new matching macro function that returns this result (although, if you need degrees, you can also do the same simple math in a macro). Not sure if the request for a matching macro is on the list?

Jaggeroth:
I think the original request was for a macro (from Wolph42?) 

I see no justification for underlying changes.

rkathey:
Is this just making sure the getTokenFacing() returns a value between  0 and 360 and rotated so 0=East?

JamzTheMan:
I believe it was a new macro/option but yes, it returned 0-360.  All of our functions that take facing as input though expect the input to match facing, so you usually do not have to translate anything.

But if you want to take facing and do math on it (for some reason?) you may want to work in degrees instead.

rkathey:
I think my vote is to ditch it, then. I do things with token facing but do my own translation.

JamzTheMan:
If the request comes in again for a macro, we can easily put one in without changing token class as well.

PR #26, opened 2015-05-16
Default the "Rename Map" input dialog to old name

DarkAlf:
Small improvement here. When a user wants to change the name of a map using the Map>Rename... dialogue it now defaults to whatever the current name is.

It's not really something we have in our issue tracker but one poster mentioned it in this thread:
http://forums.rptools.net/viewtopic.php?f=26&t=24125


Jaggeroth:
Managed to get round to doing some testing today.  This looks good.  

Only comment I would make is that when you rename your map, the Maptool window bar does not get updated to reflect the change.  This is the same as current behaviour, so not a bug introduced here, but if its possible to change it, that would be a bonus :grinning:

PR #27, opened 2015-05-17
Add new macro function snapTokenToGrid

DarkAlf:
The new macro function duplicated some functionality that previously
resided in a GUI component. That functionality has now been refactored
to a method in the Token class in order to avoid code duplication.

### DIFF
--- a/maptool/src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
+++ b/maptool/src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
@@ -46,11 +46,20 @@ public class TokenPropertyFunctions extends AbstractFunction {
 	private static final TokenPropertyFunctions instance = new TokenPropertyFunctions();
 
 	private TokenPropertyFunctions() {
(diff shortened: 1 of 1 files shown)


Jaggeroth:
This works, but is inconsistent with other functions.  To set snap to true you must code something like:

    [h: setTokenSnapToGrid('true')]

But to be consistent with other functions, your value for true and false should be a number, true(1) or false(0).

FluxIX:
'true' and 'false' are more contextually correct, although the real determining factor should be if the given value can be converted into a boolean value, since the boolean value is the desired value.

Jaggeroth:
Not sure what you mean by contextually correct.  The values expected by the function are just text literals which is not a good idea.  This is Maptool's scripting language remember, so there are no strongly typed variables and certainly no boolean.

Plus there's the consistency issue, where all other script function use 1 or 0 in similar situations.

cwisniew:
I don't mind if they accept "true"/"false" as I believe there are some functions that already do this, but they should accept 1/0 as true false too as all other functions that expect true/false values accept these.

PR #28, opened 2015-07-07
Isometric

Jaggeroth:
HI All,

I think the basics of the Isometric Grid is ready to be merged.  the functionality is described here:
http://forums.rptools.net/viewtopic.php?f=3&t=26036

**Key Features:**
* Isometric Grid
* Figure Token Type
* Token Image Tables
* Flip to Isometric Pane

cheers,

Jagged


Jaggeroth:
Just for reference I will add that once this is merged you won't be able to build a map in 1.4, export it and use it in 1.3.  But 1.3 stuff will open in 1.4 without problem.  This is due to the "Flip to Isometric Plane" option, which adds a new attribute to maps.

JamzTheMan:
We can fix the save compatibility.

Jaggeroth:
So I did a "quick hack" which I did not commit, to remove the new token properties if they were not used i.e. if set to false.  This allowed a map or token which did not use the new features to be exported from 1.4 and imported into 1.3b91.  

So this can be done, but I would appreciate some suggestions for the best way to code this neatly ;)  I will create a thread in the Maptool forum.

Probably best to wait to merge this until thats sorted.

JamzTheMan:
I have a piece of code that basically let's you 'save as' an older version,
eg b91. You set classes/fields via xstream to omit during save and it will
stop those out.

I believe there may be changes in rplib. I wish I could get some patches
out but with the summer here I have had zero free time.

After GenCon I will be running a game again so may put some patches out
then, I like to mod MT when I can evaluate it myself during game...

Jaggeroth:
That would be cool.  Shame rplib isn't in GitHub yet.

Should we go ahead and commit this and do a "Save As 91" later?

JamzTheMan:
Sure. My patch isn't too complicated and should be easy enough to
accommodate this. It touches a few classes so just needed to dig it all out.

My main issue is I'm in the process of putting a patch together of some
things Lee did last year and was surprised it never made it in the official
build. It touches a lot of classes though so trying to find time to test it
better before I submit...

Then I can bring forward the next wave of patches.

Jaggeroth:
I am more than happy to play the role of test-monkey.  If you can point me at a branch on GitHub and let me know what needs testing, I will find the time.

Jaggeroth:
The other thing I was wondering was whether I should be using the "Boolean" class instead of "boolean" primitive?  That way tokens that don't make use of the new features would have the value of null instead of false, which naturally wouldn't appear in the exported xml

Jaggeroth:
I believe this can be merged now without any compatibility issues, forward or back :)

PR #29, opened 2015-11-08
Update Readme with Markdown formatting, build commands and JDK install directions

joshsziegler:
I just took directions available on the RPTools forums and put them in the README and doc/ files for easier reference. I also formatted the existing Readme to make it more readable on Github's web interface. Some may not like how it makes the raw text look however.

Note the `doc/How_To_Install_JDK.md` file is linked to twice in the Readme.


joshsziegler:
Crap. I didn't notice the `build-resources/README.md` file. Perhaps the Build section of the main README.md should point to that file? Could we move it to the doc dir so people have one place to look for this type of thing? `build-resources/README.md` -> `doc/How_To_Build_Using_Gradle.me`

PR #32, opened 2015-11-08
Add UPnP Discovery Timeout preference to UI

joshsziegler:
I noticed the **UPnPUtil** code had a comment about needing to change the **discoveryTimeout** member to a preference, so that's what I did. I added the setting to the AppPreferences.java file and corresponding Dialog.

The setting can be found at: **Edit > Preferences > Discovery Timeout**

I'm not sure why there are so many differences in the `preferencesDialog.xml` file. I tried using `git add -p` to only add what was needed for the commit.

Hopefully I did this correctly as this is my first change to the code. :\

**FYI**: I removed the getter/setter in the UPnPUtil file because they weren't being used (as best I could tell). All code should use the AppPreferences.java getter/setter now.

### DIFF
--- a/maptool/src/main/java/net/rptools/maptool/client/AppPreferences.java
+++ b/maptool/src/main/java/net/rptools/maptool/client/AppPreferences.java
@@ -388,6 +388,9 @@ public class AppPreferences {
 	private static final String KEY_TYPING_NOTIFICATION_DURATION = "typingNotificationDuration";
 	private static final int DEFAULT_TYPING_NOTIFICATION_DURATION = 5000;
 
+	private static final String KEY_UPNP_DISCOVERY_TIMEOUT = "upnpDiscoveryTimeout";
+	private static final int DEFAULT_UPNP_DISCOVERY_TIMEOUT = 5000;
+
 	public static void setTypingNotificationDuration(int ms) {
 		prefs.putInt(KEY_TYPING_NOTIFICATION_DURATION, ms);
 		MapTool.getFrame().setChatNotifyDuration(ms);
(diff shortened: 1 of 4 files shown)


joshsziegler:
After seeing the code guidelines, I now know I shouldn't have used the English header/string for this setting in the UI. That said, I'm not sure yet how to fix it. I'll look into it.

Azhrei:
Abeille is fine for what it does, but there's no built-in integration with localization efforts.  Trevor wrote a binding object so that Abeille forms could contain strings that started with an at-sign ("@") and then the code would locate the corresponding variable in an associated object using Reflection and bind the instance variable to the associated form control.  I've never used it because I couldn't get it to work in testing; perhaps he never finished it?

Something like that should be done for all form strings.  I can see two ways of handling it.  The first is for the form to contain the resource key, such as "@maptool.formUPnP.title" to represent the title of a UPnP popup dialog.  The form class would then scan the fields and replace any field with an at-sign with a corresponding string read from the properties file.

The other possibility is to embed the English string in the form fields and use the entire string as the property key, leaving it alone if the key isn't found in the properties file.  I see this as a little clunky, but it's more readable when the form is first opened.  Note that the English string can still be replaced with a different string by populating the properties file, just as it can be with other languages.  (Might be confusing to some people:  open the Abeille form and see a particular label, only for something different to appear on the screen because it was read from an external file when actually running the app!)

Probably best to skip Abeille entirely and start learning how to build dialogs in JavaFX.  I think it has some localization support built-in, IIRC.  Be nice if we had an expert on JavaFX to help out with this stuff.  I've played with Scene Builder a little bit but never integrated it with code.

Also keep in mind that a few dialogs, like the Start a Server dialog, should ideally be restorable after being serialized.  That would let a user fill in the fields the correct way, then ship the settings to another player for them to actually start the server.  For those times when the GM is running the game from a hotel room and doesn't have UPnP control over the hotel wireless!  It would also help in supporting a "lobby" app for players to join games -- just ship them the data to unserialize into the Connect to Server dialog...

joshsziegler:
I see. So perhaps I didn't muck it up as bad as I thought. 

Also, you mention the **Start a Server** dialog. I actually put this setting in the tool-wide settings: **Edit > Preferences > Discovery Timeout**. Is that ok, or does it need to be in the **Start a Server** dialog?

Azhrei:
It's probably better in preferences.  I can't imagine people will want to change it very often, certainly not every time they start a server.  I only mentioned that other dialog as an example of things to keep in mind (serialization) when figuring out how to handle forms...

PR #33, opened 2015-11-09
Remove auto generated comments

joshsziegler:
this is just my opinion, but I don't find these comments helpful.  They just clutter the search results whenever I search for "TODO" looking for places that need work.


Jaggeroth:
Agreed

PR #34, opened 2015-11-10
Add code style and guidelines documentation

joshsziegler:
This is from Azhrei and this forum thread:
http://forums.rptools.net/viewtopic.php?f=43&t=16663

It just lists a few rules for the code and links to an Eclipse settings
file to enforce some of them. However I'm not sure all of this is up-to-date. The bit about submitting patches to the forum (vs GitHub pull request) for example. 

Also I personally don't use Eclipse (or at least haven't yet) when working on MT. Should we provide an tool-neutral description of what we want for those rules?


Azhrei:
I'm all for a tool-neutral approach (as though my opinion matters much any more!).  The best place to put it is probably in the docs directory alongside the overall developer doc.

We might also consider linking to an online standards guide (like Google's) and then list exceptions that we prefer for RPTools projects.  For example, Google lists a bunch of silly things in their C++ style guide that I would change if I were the one making edicts about code style.  I'm sure their Java document has similar things.

(See, I always read commit message threads and I sometimes even reply. :))

joshsziegler:
Good suggestion! Google's Java style guide is here for reference:
https://google.github.io/styleguide/javaguide.html).

It might dissuade people from reading it, given how long it is, but at the
very least it's well defined and thought out.

Jaggeroth:
I think there is some confusion about our Style Guide. I use (I think) the style guide from the thread you linked, but others use a different style, possibly taken from the repository.  So I would be keen to see a definitive statement about the code style from Azhrei or Craig.

I think I can adapt to any style, just as long as it isn't limited to 80 columns :D

joshsziegler:
I took care of the conflicts arising from the merge of some of my other pull request into master. 

I think there is agreement that the code style guide could be improved, but perhaps we add that to the task list for now? I mean the existing guide is still useful.

PR #36, opened 2016-01-19
View drawing

Jaggeroth:
Hi All,

This is a pull request is for my "Draw Explorer" feature discussed here:
http://forums.rptools.net/viewtopic.php?f=86&t=26382

It contains a "group" function, which if you use will make maps incompatible with 1.3, but there are no other compatibility issues.

regards,

Jagged


Azhrei:
I think grouping the shapes would be extremely useful.  How much effort would it be for them to be reloaded into 1.3 as individual members without the grouping?

If you're using the Composition pattern, it would mean writing out the XML in a different way.  I think you'd need to write out a "group" marker, then the items in the group.  When reading it into 1.3, the group marker could be ignored by the other individual shapes could still be processed.  It would require modifying the 1.3 readObject() for Zone because I don't think unrecognized elements will be ignored otherwise (although that might be a good thing to add as a separate feature since it might allow more backward compatibility).

Jaggeroth:
That's a great idea. The "Group" object is a Drawable itself and is little more than a list of Drawables. So it should be doable.

I'd rather attempt it in a different branch though ;)

Azhrei:
Heh-heh, of course use a different branch! :)

I'm not sure how best to do this, though.  One way is to edit the XML data stream as it's read in so that unrecognized elements are stripped out.  But that means teaching it which elements are valid and that will be constantly changing (possibly).

Another way would be to create an XStream unmarshalling function that unwraps the group element itself, but again, there could be many such functions needed in the future (although when MT moves to a different IO scheme, a lot of the XStream stuff will likely be rewritten anyway).  This technique does have one benefit:  the XML code could check a particular class path location and just load all available classes, thus allowing new classes to be added in the future without modifying the base code.

Probably too much work, though. :-/

JamzTheMan:
FYI: Xstream has a built in function to  omit fields/classes. It's only a
2-3 lines of code.

I did it as a POC to strip a new field from the Lights class on a custom
"Save as" function.

You can also easily translate incoming/outgoing fields to new names.

If that was what you were looking for?

Azhrei:
Yes, but the problem is not omitting an element (and all of its subelements), but eliminating the start element and end element WITHOUT eliminating the content between the two.

I'm wondering if it wouldn't be better for the Group object to be inserted into the XML using a custom marshaling function that embeds a single Drawable called Group which then contains an attribute specifying how many of the following Drawables belong inside it.  That way XStream can be taught to completely ignore the Group element because the other components are not embedded inside it.

Or at least, that's what I remember about XStream.  I will admit it's been quite awhile since I've looked at the docs for it.

Jaggeroth:
Perhaps we should move these comments to a new thread in the forum?  There are other things that would need to handled as part of a "Export Map to 1.3" function, like Figure Tokens and Isometric Maps and no doubt, more will come as we add things to 1.4.

Azhrei:
Sure.  I'm not even convinced that "export to 1.3" is even a worthwhile goal, given how many accommodations would need to be made...

Jaggeroth:
It might also be useful for MOTE compatibility.

PR #45, opened 2016-05-02
Parameter check changed to accept 3 parameters for table image

Jaggeroth:
Mea culpa

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/45)


Azhrei:
I started work on an MTscript test file at one point.  Something that would exercise all MTscript functions with a variety of parameters.  The obvious goal being to run the script during the test phase and determine whether anything fails.

Easier said than done.  There are a LOT of script functions!  And we were talking about switching to JS, so given the size of the job and the trajectory for the scripting language at the time, I put it off.

I'll see if I can find that code and get it checked into to Github...

Jaggeroth:
That would help :)

ps: Are you going to sign up for the slack?

PR #80, opened 2016-05-27
Updates to wave 2

JamzTheMan:
- New checkRegion method to do a more relaxed bounds check, for use
with VBL Tokens.
 - Maptool installing spash updated to match MT splash (added shadow
under gear logo)
 - VBL Tokens now only show when token can "see" them. Currently 2/9ths
of token need to be seen, that is 2 out of 9 "regions".
 - jWrapper.7z removed, was a backup artifact
 - Updated gradle build to reflect new launch.properties config to
replace mt.cfg
 - launch.properties is now updated and stored as a resource in the JAR
to be used as a default for new installations and to update JAR/JRE path
values on new installations
 - New property MAPTOOL_VERSION added to track launch.properties and
update properly when new MapTool version detected
 - Fixed Logging Bug, logging selections now persist properly across
multiple launches
 - Default configuration (launch.properties) should properly install
itself across all OS types
 - If jWrapper JRE is used (default unless changed) and a new version is
detected, the config property will be updated
 - copyXmlFiles method commented out, default logging xml files stored
in .maptool/logging directory 
 - launch.properties in the same directory as the JAR will be used
instead of what is stored in .maptool/config directory to allow
mutltiple versions to be installed/ran manually

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/80)


JamzTheMan:
The gradle.build file probably has conflicts but the changes are minor, shouldn't be an issue to fix.

PR #136, opened 2017-02-05
Enhancements for 1.4.1.8

JamzTheMan:
o Gradle fix to properly copy & update launch.properties file during
configuration.
 o jWrapper jar updated to 0049034758
 o Gradle enhancement to rename and copy completed jWrapper
installations to target directory
 o Gradle enhancement to run new class CreateVersionedInstallSplash
which creates a matching "Installing" splash image which jWrapper embeds
and uses in it's installation.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/136)


Azhrei:
Jams, have you found a way to use relative pathnames in the CreateSplashImage class?  I think you were going to look at changing it so that instead of "D:\..." it uses a relative path; wasn't the problem not knowing the directory Gradle was in when running it?  That's pretty easy to check (create File(".") and print .getCanonicalPath), so I'm thinking it must be something else.

Also, I was going to create a branch to merge this into (in fact, I had done that) and then saw that Travis is executed on the PR _as though_ it had been merged, so there's no reason to make a separate branch in the code repo.  Cool. :)

I'm going to see about finishing the download page, then I'll do some builds and upload them to see how everything looks.  Oh, and I'll copy the old stuff from Craig's site so that everything is on rptools.net.

Thanks for the PR!

JamzTheMan:
I haven't finished it yet. I was going to pass parameters to it instead (IIRC it is already set to accept command line parameters).

In then Gradle can pass the required info and customized via gradle.properties if needed. Is that cool?

It will have to wait until at least Monday though if that is cool?  Figured in the mean time you can test it as is and make sure new JRE/jWrapper and all that works without a hitch?

Azhrei:
Sounds like a good plan. :)

And yes, I plan to test the JW stuff quite a bit.  I need to get JW updated on my Mac, then also inside my Linux VM.  I just upgraded to Neon in both places, so I'm sure something has broken (you can count on that with Eclipse ☹️ ).

Since I've got two traveling gigs coming up, it will easily be a few days before I get to that, but I expect the gradle build will fail when I attempt to run it on any of my Un*x boxes; I can change the filenames pretty easily though.

JamzTheMan:
FYI I did check in the latest jar so you should be able to build just from my PR. JRE update is outside that though.
And FYI for now I just checked in the 
Manifest files so we have em. We'll automate that step later...

Any thoughts on automating downloading the JRE as a Gradle task?

JamzTheMan:
And regarding Neon, honestly it was the smoothest eclipse to date. I didn't upgrade though, that is what bites you. But I don't believe I had to install any plugins or do any special set up this time. Install, import Gradle project, done!

Azhrei:
Good to hear.  Thanks.

It would be nice to have a gradle task for updating the JRE, but Sun/Oracle tends to change the URLs regularly, an there's the whole "accept the license" thing.

A quick Google search wasn't helpful; too many bogus hits.  But if you across an existing configuration, we should certainly look at it to see how they take care of the above concerns.

JamzTheMan:
That's what I kind of figured as well. If you run the jwrapper app they have download links bland u accept the license before it downloads but I'm sure keeping up the links can be a pita...

Issue #164, opened 2018-01-16
Token Facing Broken in 1.4.1.8


rkathey:
The current build does not show the facing arrow for Round and Square token shapes. Top down tokens still face properly. 

This is broke regardless of whether you are using the mouse wheel or popup menu option.

http://forums.rptools.net/viewtopic.php?f=21&t=27432

JamzTheMan:
Ok, so traced this back to here: a71a2cf8beccc6f931357d9274a46dc39cf2ce4e

So not a "bug" per sae but a change to the use of the App Preference (originally used only for Top-Down forcing of arrows I believe).

Not sure if it was changed as the developer thought it was a bug or felt it should do this? Personally I don't want arrows on my "Top Down" tokens but now I have to have it that way if I want it on my Circle tokens? At the very least it should be another App Preference, otherwise this change should be backed out.

Phergus:
Fixing this shouldn't force arrows onto Top-Down tokens.

It sounds like the preference got misused as some kind of setting for ISO tokens with image tables.

JamzTheMan:
This option was added in 1.4.4.0. and per the tool tip, was only used to force arrows on Top-Down and tokens with Table Images (Figure tokens).  

And ya, now it's being misused to group Circle and Square into it as well.  @Jaggeroth would probably need to fill in the need for it but I'm ok with the original App Preference use.

Azhrei:
I presume your mention of 1440 is for Nerps.  I also agree -- it seems that it should only apply to Top-Down (and if it includes tokens with table images, I guess that makes sense; I don't know, but someone like Wolph2 would know).

I'm fine backing it out, if there's consensus that it needs to revert back.  Whoever is going to look at this, should also investigate to see where else the "force" flag is used and determine if there need to be two options (which would depend on the use cases).

JamzTheMan:
sorry, corrected. I meant 1.4.0.0 (sorry those numbers transpose so easily. Which is why I try and always add Nerps + version less we both have the same version # at some point and there is confusion). If I don't say Nerps assume RPTools version/repo.

PR #177, opened 2018-03-05
Reverse change that causes facing arrows to only show when the "Force…

Jaggeroth:
https://github.com/RPTools/maptool/issues/164
Correct behaviour is:
* Facing Arrows should appear when a facing is set for Circle, Square and Figure tokens (without Image tables).
* Facing arrows should not appear on topdown tokens or Figure tokens with image tables.
* Facing arrows should always show if the "Force token facing arrow" preference is set

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/177)


Azhrei:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/rptools/maptool/177)*

Issue #194, opened 2018-07-14
Player clients logging in with drawing layer other than Token can draw on that layer.


Phergus:
If a player has selected a layer other than Token in the drawing tools just prior to connecting to a server their drawings will appear on that layer.  i.e.  A player could be drawing on the Hidden, Object or Background layers!

Seen in 1.4.0.5 and current Nerps.

Azhrei:
Wow, good catch!  Luckily it should be a one or two line fix, I would think.  Jagged will likely get to it before I have his current PR merged.  (@Jaggeroth Please file a separate PR and refer to this issue.  Thanks!)

Phergus:
I would think a simple setCurrentDrawingLayer("Token") or whatever at client connection init would do it.

Jaggeroth:
See Pull Request #195

Phergus:
Fixed with PR #195

PR #195, opened 2018-07-15
Player drawing layer - Issue 194

Jaggeroth:
Fix for #194 
Prevent players from being able to draw on layers other than TOKEN

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/195)


Jaggeroth:
*sigh*

Accidentally started this branch from my GetNearestVertex() branch instead of the dev branch :(

Can you live with that or should I resubmit?

Azhrei:
I'd like to keep things separate, if possible.  Are you saying that this builds on the PR for #193?  In that case, I can apply #193 first and, assuming that it applies cleanly, apply this one, and the result will be an incremental patch for this issue.

Jaggeroth:
You are correct. I accidentally branched from the wrong point. 
One of those times when tiredness did not make for the wise choice ;)

PR #202, opened 2018-08-07
Update i18n_it.properties

jappyjoker:
Updated and fixed lots of traductions from English to Italian

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/202)

### DIFF
--- a/maptool/src/main/resources/net/rptools/maptool/language/i18n_it.properties
+++ b/maptool/src/main/resources/net/rptools/maptool/language/i18n_it.properties
@@ -450,13 +450,13 @@ component.tooltip.macro.sortPrefix=Determina l'ordine nel gruppo in cui questa m
 component.tooltip.macro.hotKey=Imposta un tasto di scelta rapida per la macro. Questi tasti sono comuni in tutti i pannelli macro.
 component.tooltip.macro.includeLabel=Seleziona per includere l'etichetta della macro con il suo output nella chat. Deseleziona per avere solo l'output della macro in chat.
 component.tooltip.macro.autoExecute=Seleziona per avviare la macro appena il suo bottone è premuto. Deseleziona per impersonare (se applicabile) e preparare il comando della macro nel box dei comandi.
-component.tooltip.macro.applyToSelected=Seleziona per avviare la macro applicata al token attualmente selezionato. Deseleziona per avviare la macro fuori dal contesto di qualsiasi token selezionato.\nSe la macro fa riferimento a una o più proprietà dei token, dovrebbe essere selezionato.
+component.tooltip.macro.applyToSelected=Seleziona per avviare la macro applicata al token attualmente selezionato. Deseleziona per avviare la macro fuori dal contesto di qualsiasi token selezionato. Se la macro fa riferimento a una o più proprietà dei token, dovrebbe essere selezionato.
 #TODO: Write tooltip for this.
 component.tooltip.macro.buttonColor=Seleziona il colore del bottone di questa macro.
 component.tooltip.macro.fontColor=Seleziona il colore con il quale l'etichetta del bottone di questa macro sarà visualizzata nel pannello.
 component.tooltip.macro.fontSize=Seleziona la grandezza del carattere con il quale l'etichetta del bottone di questa macro sarà visualizzata nel pannello.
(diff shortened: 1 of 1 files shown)


jappyjoker:
Yes I tryed to be consistent. Some of the underscored letters displayed a different shortcut than the one shown on the right of the word. And then, keyboard shortcut are displayed, so the ampersand seems like a repetition :)

Azhrei:
First, thanks for contributing!  The more people who contribute, the better the software becomes!

@jappyjoker wrote:  "Yes I tryed to be consistent. Some of the underscored letters displayed a different shortcut than the one shown on the right of the word. And then, keyboard shortcut are displayed, so the ampersand seems like a repetition :)"

I don't understand this and it may be a terminology problem.

The & is interpreted by the builtin menu system in Java AWT/Swing.  It causes the following character to appear with an underscore underneath it and is called a "mnemonic".  If the menu is being displayed, that letter can be typed to activate the corresponding menu option.  This allows the letter "S" to be underlined on multiple different menus, for example.  (Very few platforms might require a modifier key, like Meta-S, to activate the menu option, but the menu must still be displayed.  Some platforms might not implement mnemonics at all, or implement them without visual feedback, such as macOS.)  So the "S" in "Save" might be underlined on the **File** menu, but the "S" might also be underlined in "Search" on a fictional **Navigate** menu, too.

A shortcut is different.  Shortcuts are keystrokes that are available without a menu being displayed, so the user might be in the middle of some other operations elsewhere in the application, yet the shortcut still works.  They are typically activated using a modifier key, frequently with multiple modifier keys, such as Meta-S for **Save** and Shift-Meta-S for **Save As...** Some applications might change shortcuts depending on what mode the application is in (edit mode vs. play mode), but this is generally frowned upon in user interface guidelines as very confusing for the user.  (And yet **vi** does this. Go figure.)

So this means that a menu option can have a mnemonic of one letter and a shortcut of another letter.  The ubiquitous cut/copy/paste are typically Meta-X, Meta-C, and Meta-V on most platforms, but that doesn't mean you'll see an underlined X in the word **Cut** when it appears in a menu!  (Most applications choose to underline the **t** in this particular case.)

What's the result of all this?  The & should still be there, even in cases where the shortcut key is different.  I realize that most users and programmers aren't familiar with all the nuances, but I'm afraid we'll need the & added back in again before this is likely to be merged.

jappyjoker:
Thank you for your explanation @Azhrei, it is very useful. I'll fix this issue and upload the file again

jappyjoker:
You are definitely right! I was focusing on the lines after and I forgot the main menu headings... Now it should be fine. I also fixed some traductions. Many things are still not internationalized and they aren't in the English file, so I couldn't translate them... I will need some one more expert than me to do that

Azhrei:
@jappyjoker "I also fixed some _traductions_."

I think you mean _translations_.  (Honestly, if you had to try to read my attempts at Italian, it would make your eyes bleed; your English is great, though!)

"Many things are still not internationalized and they aren't in the English file, so I couldn't translate them..."

Yes, there are still strings in the source code that have not been externalized.  There are tools to help with that job, but it's tedious.  And it's not always simple, either; some strings, mainly debugging and logging messages, we want to leave in English so the developers will be able to read them if there's a problem.  That means a volunteer can't just move **all** of them to external files without knowing their purpose...

Azhrei:
Yes, the tool we use for designing dialog panels leaves a lot to be desired.  We can add text to a dialog in the tool fairly easily, but it doesn't have a simple way to replace strings from a properties file.  I had looked into this awhile back and it became pretty tricky.  Given that we'd _like to_ leave that tool behind and move to JavaFX, there's not a lot of incentive to spend time on it at this point.

I'm looking at way to apply this to the dev branch instead of `master`, but it _will_ be applied. :)

Azhrei:
@jappyjoker I'm having a problem isolating just the properties files when I try to apply this patch against `dev-branch-1.4` instead of `master` and I could use your help.  I'm going to close this PR, but would you resubmit your changes against the dev-branch and only include the properties file?  Worst case, you could copy your file somewhere else, checkout the dev-branch in your fork, then copy the file back in.  When you commit your change, it will be only that one file and the next PR will be clean.  (There are other ways to do it as well, depending on how well you know Git.)

Thanks.

PR #215, opened 2018-12-02
Changes to get the Java FX up and running in Mac OS

cwisniew:
**Changes to get the Java FX up and running in Mac OS**

JIDE DockManager fails on Mac OS as there is no WindowLookAndFeel class available, so MapToolFrame no longer extends JIDE DockingManager. All calls to DockingManager class are now routed to a temporary stub class which will pop up a dialog as all this functionality will be replaced as part of the JavaFX port anyway.

Also removed the MetalSliderUI look and feel for the slider in the AssetPanel as this causes a null pointer violation under Mac OS as the apple settings for look and feel do not provide all the required default values.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/215)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/MapToolFrame.java
+++ b/src/main/java/net/rptools/maptool/client/ui/MapToolFrame.java
@@ -10,8 +10,6 @@ package net.rptools.maptool.client.ui;
 
 import java.awt.BorderLayout;
 import java.awt.Color;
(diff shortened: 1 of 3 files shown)


JamzTheMan:
No issues for me.  MapToolFrame will be removed eventually anyhoo and the panels redone so this gets it working for development reasons, thanks!

PR #224, opened 2019-01-07
Fix invalid variables in calculated token property

Darinth:
Invalid variables in calculated token properties will no longer cause maptool to spam up dialogs and possibly lock up.
MapToolVariableResolver now has a property called autoPrompt, which (if set to false) will prevent automatic prompting for variable values.
Token.getEvaluatedProperty can now take a MapToolVariableResolver to use in resolution.
PointerTool now creates a MapToolVariableResolver, disables autoprompt on it, and passes it into the token.getEvaluatedProperty during display of the statsheet.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/224)


Darinth:
Issue JamzTheMan/MapTool#96 actually touches on that question a bit. The statsheet is the location where this code was needed the most, but it can also cause some issues when you have things like a very long looping macro that keeps prompting because of a mistyped variable name. I don't know that there's a lot of use in providing a default when auto prompt is turned off, but it might be useful to look into doing something along the lines of 'when a macro input is canceled, it disabled autoprompt' to let a macro crash end that the user probably has no interest in seeing the results of anyways. But honestly, at that point we're probably better off to have the entire macro just get terminated.

Darinth:
And I forgot that my pushing those commits to master was going to also put them into this pull request.

JamzTheMan:
@Darinth This PR was submitted against Master, we will be building our next Developer Release based on merge-from-nerps branch.

If you could fix the conflicts and redo based on this branch, we can look to merge this into a future Developer Release.

Darinth:
Closing and creating with new PR off of correct branch in my fork.

PR #228, opened 2019-02-06
Pen play from nerps

Jaggeroth:
Replacement for PR 174 Pen Play

Contains changes to allow square line ends on drawings.
MT Script functions to access drawing functions (see wiki)
Fix for npe caused by maps with ghost drawings.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/228)


JamzTheMan:
Looks good. But... :-)

Now that we have the Code Editor with Syntax Highlighting, we need to add the new Macro Functions manually (although I think there is a way to automate/dynamically do this in code but it's not there yet).

There is a external jar that can be ran to generate the proper files but I think we can edit them manually as well (adding function names is easy).

Do you want to take a stab at it (with guidance) or do you want me to handle it? I'm all for granting learning opportunities lol

Jaggeroth:
Point me at it!

JamzTheMan:
Hmm, looking at it again, I think we still need to read the jar.  You can edit MapToolScriptTokenMaker.flex to add the new function names (the flex file was initially created but TokenTokenMaker jar, a GUI to make the flex file) but then there's still another process that generates MapToolScriptTokenMaker.java from that flex file.

It's been a while so I'll have to look at it at home.  Feel free to edit the flex file and add it to the PR and I'll see if I can document the steps from there.  I may or may not have already added the external jar to build resources or such, I'll have to check from home.

JamzTheMan:
Ok, I'll check this in and work on the MapToolScriptTokenMaker to see if we can get macro functions automated for the editor.

PR #229, opened 2019-02-07
New function for HOME key

Jaggeroth:
Replacement for PR 196
Allow HOME key to rotate between visible PC tokens owned by player

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/229)


JamzTheMan:
Easy one! :+1:

PR #176, opened 2018-03-04
Whisper function

Jaggeroth:
Whisper command added to the "Connections" panel.
**Warning:** The internationalisation has been done via Google Translate. But it only had to translate "Whisper" and "enter text here" so hopefully its okay.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/176)

PR #235, opened 2019-02-11
Whisper command

Jaggeroth:
Replacement for PR Whisper function #176

Whisper command added to the "Connections" panel.
**Warning:** The internationalisation has been done via Google Translate. But it only had to translate "Whisper" and "enter text here" so hopefully its okay.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/235)


JamzTheMan:
Approved!

Issue #240, opened 2019-02-12
Macro Editor Auto-Completion


JamzTheMan:
Turn on RSyntaxTextArea's auto-completion which adds word completion for Macro function names and display a short description + summary of the function.

Also add a few short hands for if/for/switch

JamzTheMan:
The framework is done and ready to add descriptions.  Will create another issue to track adding in descriptions.

Azhrei:
Periods are treated as non-identifier characters so that `son.sort` doesn't autocomplete properly.  @JamzTheMan already found the fix, so I'll submit a PR shortly.

JamzTheMan:
Thanks!

cwisniew:
Moving this to post Nerps fix as it isn't critical for the release hasn't been addressed for 4 days

Azhrei:
This has been merged.

Issue #239, opened 2019-02-12
Extend MapToolScriptTokenMaker to dynamically pull in function names


JamzTheMan:
Current method of using TokenMakerMaker to generate Flex file and run JFlex to generate source code is cumbersome and often missed when adding new macro function names.

Should extend this class to pull in macro function names dynamically.

JamzTheMan:
done!

PR #241, opened 2019-02-12
Extend MapToolScriptTokenMaker

JamzTheMan:
PR for Issue #239 & #240

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/241)


JamzTheMan:
I'm going to merge this for now but consider it WIP and may change as we solidify how to handle macro descriptions.

JamzTheMan:
Ok not sure why Travis reports yellow, it completed successfully. I will return job and see what happens.

Azhrei:
I18N might be fine place for it, but then some reorg is needed.

The long line will freak out some tools (like GH Reviewable!) and makes it difficult to see changes to one word (because the whole line is in the diff).

No big deal from a functionality standpoint.

JamzTheMan:
What reorg do you have in mind?

And all valid points re long line, makes sense, I can make that happen.

PR #276, opened 2019-03-06
Add support to catch an abort and a new return function

oliver-szymanski:
To exit a function/macro use the new "return" function.
First parameter is same as in "abort" function, so the return will only
do anything if first parameter is 0. Second parameter is an optional
return value, which will be put in macro.return and in output text of
the macro using the "return" function.

E.g.:
[return(0, "returned hello world")]
Flow will end and leave the macro continuing in the calling macro once the return function is evaluated.
To not stop all macro execution when some macro in the callstack is
calling "abort(0)" you can set a variable named "macro.catchAbort=1".
All macros in the callstack will still be aborted, except the one with
"macro.catchAbort=1" in scope.

E.g.:
[h: macro.catchAbort = 1]
[h: resultText = "defaultValue"]
[h: resultText = abort(0)]
[r: resultText]
Will not abort but instead have "defaultValue" in resultText. The abort can happen anywhere above in the callstack, so could also happen in another macro that we call.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/276)


oliver-szymanski:
OK, if you guys don't want it I close this now cause I'm making more changes on my side. Happy to provide the catch of abort and the new return function again if someone wants it.

rkathey:
The problem is we're in the middle of getting a release out so everyone's too busy to evaluate at the moment. You can leave it open to be incorporated into the next release if you like.

cwisniew:
👍  to leave open/resubmit.

Happy to take this for next minor release, but at the moment it seems we are losing the war against scope creep for the current build and trying to concentrate on just getting the build out

oliver-szymanski:
The problem is I pushed much more on that branch, including new ways to add macros/functions as framework bundles. So much more than you guys probably need on top of the abort/return. I will create a new clean branch for another PR

rkathey:
That sounds great. We have some MTScript projects coming up you may be interested in contributing to as well. You can join in the Discord server to participate in the discussions. https://discord.gg/crpk7FM

oliver-szymanski:
I made a new branch and another PR https://github.com/RPTools/maptool/pull/283

If anyone is interested I've also implemented on a way to import (Java based) functions as bundles to support different frameworks for my role playing systems. Main work is done and working with nearly no core changes by https://github.com/oliver-szymanski/MapTool-nerps/blob/jadebringer/src/main/java/de/jadebringer/maptool/functions/AddFrameworkFunction.java

Thanks everyone.

PR #283, opened 2019-03-09
Return and catch abort func

oliver-szymanski:
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/283)


oliver-szymanski:
Hey finally its green... never work on a fork of a fork again ;)

PR #286, opened 2019-03-10
Fix not working syntax highlighting for new defined functions

oliver-szymanski:
If you use defineFunction these new functions to not get syntax
highlighted straight away when manually calling the onCampaignLoad
macro. Only when the campaign is reloaded. This fixes that behaviour.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/286)


JamzTheMan:
So this is highlighting UDF's now? cool. I wanted to get to that...

I sort of wanted them to have their own color though so they stand out...can be enhanced later though.

oliver-szymanski:
Yeah, that's for the UDFs. I need it for my plugin frameworks as well, but this is for the UDFs

Issue #287, opened 2019-03-10
PR #283 merge breaking existing macros that use abort()


Azhrei:
It seems that macros that call `abort()` are causing a popup from the MapToolVariableResolver when the Java code tries to reference the `macro.catchAbort` variable.  Since the variable resolver has a parameter now to turn off those prompts, the code should probably be using that.

JamzTheMan:
Beat me lol

@oliver-szymanski I opened up my existing campaign and got a 
![image](https://user-images.githubusercontent.com/8051654/54087350-f1208c00-431f-11e9-98b7-9fbda28d1d48.png)

oliver-szymanski:
Will check and fix

oliver-szymanski:
Fix in https://github.com/RPTools/maptool/pull/290

Azhrei:
PR merged

Azhrei:
This merged code is throwing NPE when the `macro.catchAbort` is set to 1 and the `abort` function is called with an empty roll option.  It appears that the code is trying to create a tooltip from the output and there isn't any, hence the NPE.  Put this into a macro and run it; when the popup appears, enter `1` and it works, but enter `0` and it throws an NPE.
```
[macro.catchAbort = 1]
[t: abort(var)]
<br>
More text...
```

oliver-szymanski:
on it

oliver-szymanski:
Sadly I have to raise a new PR as I deleted the nerps fork and so cannot reuse the original PR.
Fix in https://github.com/RPTools/maptool/pull/294

JamzTheMan:
Are we reverting this PR for this release or merging #294? I honestly don't know where we are with the testing?

Azhrei:
AFAIK, aliasmask was happy with the proposed API but didn't have a build that he could use to test it.  IMO, it should be merged and s build done so he can try it out.  It's pretty easy to click the Revert button if it doesn't work out...

JamzTheMan:
k, I'll merge it and create a build so testing can be done

PR #294, opened 2019-03-11
Catch abort

oliver-szymanski:
Ensure catchAbort for UDF, straight macros and roll options Issue #287
As discussed with Azhrei I deleted the nerps fork, so cannot reuse the old PR.

Previous catchAbort ignored direct macro calls and had an issue with
some roll options. This fixes that.

Sample of calls (with raiseAbort@this being a macro doing an abort(0)
and raiseAbort, raiseAbortNotIgnoreOutput 2 UDFs pointing to the macro
wit/without output ignored):

`[h: var = 0]
[h: macro.catchAbort = 1]
[h: result = "default value"]
abort_r: [r: result = abort(var)]

result_r: [r: result]
abort_t: [t: result = abort(var)]

result_t: [r: result]
abort_t_code: [t, code: {
[result = abort(var)]
}]

result_t_code: [r: result]
abort_r_code: [r, code: {
[result = abort(var)]
}]

result_r_code: [r: result]
abort_f: [raiseAbort()]

result_f: [r: result]
abort_f_with_output: [raiseAbortNotIgnoreOutput()]

result_f_with_output: [r: result]
abort_macro: [macro("raiseAbort@this"): ""]

result_macro_code: [r: result]
More text...`

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/294)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
+++ b/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
@@ -1312,7 +1312,19 @@ public class MapToolLineParser {
                 String callName = result.getValue().toString();
                 result = parseExpression(resolver, tokenInContext, rollBranch);
                 String macroArgs = result.getValue().toString();
-                output_text = runMacro(resolver, tokenInContext, callName, macroArgs);
+                
+                try {
+                	output_text = runMacro(resolver, tokenInContext, callName, macroArgs);
+                } catch (AbortFunctionException e) {
+                    // required to catch abort that are not 
+                	// in a (UDF)function call
+                	// but in a real "macro(...)" call
+                	log.debug(e);
+                    boolean catchAbort = BigDecimal.ONE.equals(resolver.getVariable("macro.catchAbort"));
+                    if (!catchAbort) throw e;
+                    output_text = "";
+                }
+                
                 if (output != Output.NONE) {
                   expressionBuilder.append(output_text);
                 }
@@ -1425,7 +1437,13 @@ public class MapToolLineParser {
       log.debug(e);
       boolean catchAbort = BigDecimal.ONE.equals(resolver.getVariable("macro.catchAbort"));
       if (!catchAbort) throw e;
-      return null;
+      
+      // return an empty result to not collide with tooltips
+      // when catching an abort
+      Result result = new Result("");
+      result.setDetailExpression("");
+      result.setValue("");
+      return result;
     } catch (AssertFunctionException afe) {
       log.debug(afe);
       throw afe;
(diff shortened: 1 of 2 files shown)


oliver-szymanski:
I'm not 100% about the tooltip. The error is fixed but not sure if that "[t:" should evaluate to anything specific as tooltip. Feel free to tell me or ping on discord if you want me to amend to that ;)

Thx everyone

Azhrei:
It looks like it returns an empty string, then?  I guess since `abort` usually just kills everything, some discussion around what the various roll options should do seems inevitable.  This change was to fix the NPE, but is the tooltip text ever going to be used anywhere if the abort is being caught?

oliver-szymanski:
I don't think it ever comes up cause there is no output in the end. But as
said in the comment i'm not sure what's expected cause I do not use t: a
lot for such function.

If anyone could clarify that would be great. As far as I can see running
the example only t: had a problem

Azhrei:
Yes, I only had an issue with **t:** as well.  I don't do much macro writing though, and I'm not up on current best practices so there may be other roll options that trigger exceptions.  I'm hoping **aliasmask** on Discord can provide some input.

aliasmask:
Can we get assert() to work the same with catchAbort as abort() does? Assert is an abort that posts a message to the chat.

`
[H: macro.catchAbort = 1]
[H: result = "subMacro was aborted"]
[H: result = subMacro()]
[R: result]
[H: macro.catchAbort = 0]
[R: "End Main"]


[H, for(i,1,10), code: {
   [H, if(i == 5): assert(0,"Failed on index "+i,0)]
}]
[H: macro.return = "End Sub"]`

Output would be:
Failed on index 5
subMacro was aborted End Main

oliver-szymanski:
Possible, just not so sure if that is really wanted? Or at least should
that be another flag? like macro.catchAssert?

I left it out before cause I think there is a conceptional diff between
assert and abort and I did not want to treat both the same. Wilk be on
later in discord

aliasmask:
I would be okay with catchAssert. I would probably use that more than catchAbort because there is a message option since I'm already focused on error checking with a poor mans try/catch.

oliver-szymanski:
working on catchAssert now, lets see what I can achieve there

oliver-szymanski:
Added catchAssert based on my discussion with @aliasmask

oliver-szymanski:
@aliasmask  are you fine with the catchAbort? I think @Azhrei is waiting for your feedback before we go further with this PR

aliasmask:
If it works as specified, then it's okay with me. I can't actually test myself. I don't have a build of it.

oliver-szymanski:
Sure, this is how it looks, so as you have defined :) 

image

aliasmask:
Yep, that looks correct. It has my approval.

oliver-szymanski:
@Azhrei If you want me to change anything else related to this please tell me. Sounds complete to me otherwise. Would be great to have this in 1.5.0 so that all this catch stuff is consistent.

PR #305, opened 2019-03-15
Fix #292 wrong URLs in Help menu

### LINKED ISSUE #292: Update URLs in the Help menu
The **Documentation** and **Tutorials** URLs used in the **Help** menu point to locations that are no longer correct and don't redirect to the right place (and they're about 8+ years old, so that's no surprise).

Verify all of these URLs and update as necessary (URLs are stored in the **i18n.properties** file; perhaps some day they should be read from our server and cached for future use, that way we could update them, if needed).

oliver-szymanski:
As far as I can see only the first 2 links are wrong (to general documentation and tutorials are wrong

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/305)


JamzTheMan:
Some minor conflicts need to be resolved.

Phergus:
@oliver-szymanski See Jamz comments above.

JamzTheMan:
We're going to have to figure out how to better manage i18n properties...

Can you resolve this with one as well?

oliver-szymanski:
I think the problem with auto merging was about line ending (windows/unix style). should be better now

JamzTheMan:
It still shows conflicts for i18n.properties.

PR #310, opened 2019-03-15
Fix #189 amend transferVBL function to really transfer

### LINKED ISSUE #189: transferVBL macro function is misnamed
The transferVBL() function is somewhat misnamed.  If you transfer something it is moved from one place to another.  The function is really a copyVBL() function.

Changing this might break existing macros but it would be useful if it did what it said.

Alternatively a second parameter that determined transfer or copy could be added.  This might also break existing macros as the current second parameter, token ID, would need to be pushed to the third slot.

oliver-szymanski:
introduces an optional delete (boolean 0/1) parameter to not just copy but move the VBL.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/310)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/VBL_Functions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/VBL_Functions.java
@@ -47,6 +47,8 @@ import net.sf.json.JSONObject;
  * in VBL
  *
  * <p>getVBL(jsonArray) :: Get the VBL for a given area and return as array of points
+ *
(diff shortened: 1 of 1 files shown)


JamzTheMan:
cool beans. yer on fire :)

oliver-szymanski:
lol, just trying to help. hope it also helps me later to code more rules for my RPG ;)

JamzTheMan:
Some minor conflicts need to be resolved.

PR #319, opened 2019-03-16
Fix #50 unit per cell as double

### LINKED ISSUE #50: Enhancement: Decimal digits in map properties (distance per cell)
[Forum request: Decimal digits in map properties (distance per cell)](http://forums.rptools.net/viewtopic.php?f=26&t=26636&p=262974#p262974)

> It would be great if you could use decimal digits (like 3.5 or 4.7) at "Distance per cell" in the map properties. It would make MapTool more international. A little example:
> 
> In Pathfinder one cell on the map is 5 foot. 5 foot are 1.5 metres in the metric system, which is used all over the world outside the USA (and so in the rpg rulebooks of course).
> 
> Not a big thing, but it's implemented in other virtual tabletops like Roll20, and it would make things just a little bit easier and more compfortable for non-americans.

Should be doable, duplicates another forum request.

oliver-szymanski:
Include to calculate all distance in double (except when calculating pixels of course)

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/319)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/TokenLocationFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/TokenLocationFunctions.java
@@ -331,7 +331,7 @@ public class TokenLocationFunctions extends AbstractFunction {
       }
 
       // Get the distances from each source to target cell and keep the minimum one
-      int distance = Integer.MAX_VALUE;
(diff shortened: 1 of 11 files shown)


JamzTheMan:

PR #320, opened 2019-03-16
Fix #313 add macro.catchAbort and macro.catchAssert to syntax

### LINKED ISSUE #313: Macro Editor Help needs update for macro.catchAbort & macro.catchAssert
As it says.  The editor isn't recognizing `macro.catchAbort` & `macro.catchAssert`.

oliver-szymanski:
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/320)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
+++ b/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
@@ -24,6 +24,8 @@ import java.util.Set;
 import java.util.Stack;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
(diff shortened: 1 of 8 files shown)


Azhrei:
I would much rather this be done dynamically.  Is there a method call that the code could make that lists all of the variables at runtime instead of hard-coding them?

oliver-szymanski:
Currently the static way is implemented for all special chars,  so I just added them. I like a dynamic approach more anyway, so I will change that (at least for these new special vars)

oliver-szymanski:
Done. Functions can now define specials variables that are picked up in syntax highlighting and auto completion.

Also fixed a Maptool issue, that existing special vars and reserved keywords showed up in syntax highlighting but not in auto completion. Not sure if there was ever a Jira for this latter bug.

JamzTheMan:
Fixes #313

PR #325, opened 2019-03-17
Fix #324 getViewArea was returning an wrong formatted string.

### LINKED ISSUE #324: getViewArea is returning a useless value
It appears that the macro function getViewArea is return a string containing the Java object representation. If I do this: 
```
[getViewArea(0)]
```
I get this returned: 
```
java.awt.Rectangle[x=40,y=50,width=75,height=39]
```
Which I cannot do much with directly. I have to parse it somehow to extract the relevant values. It would be much preferable to have the method return a json object. 

Edit: This is in 1.5.rc.4.

oliver-szymanski:
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/325)


oliver-szymanski:
Fix for https://github.com/RPTools/maptool/issues/324

JamzTheMan:

rkathey:
I'm not sure the numbers add up.
getViewArea1 5 2

I ran this script
<br> Nothing [r: getViewArea()]
<br> 0 [r: getViewArea(0)]
<br> 1 [r: getViewArea(1)]
<br> 0 json [r: getViewArea(0, "json")]
<br> 1 json [r: getViewArea(1, "json")]
<br> 0 , [r: getViewArea(0, ",")]
<br> 1 ; [r: getViewArea(1, ";")]

Phergus:
The numbers that are output agree with each other but the width and height in cells are way off.
Too funny.  It is dividing the viewport size in pixels by grid size to get the size in cells and ignoring zoom level.

Phergus:
It was broken this way in Nerps 1.4.5.4 as well.

oliver-szymanski:
I only changed the formatting of the output, as this was reported as an
error, not the way it counts.

If you tell me what needs to be changed to fix the numbers I can add that
as well

Phergus:
Yeah.  It's not something you did.  It's been like this from the beginning.

The question is what is the correct output in cell sizes for a view like above where partial cells are shown on all sides?  Do you include the partials in the count or only those cells fully visible?   Or do you divide the size in pixels by grid size and factor in zoom?  Wouldn't think folks would want this but maybe.

PR #351, opened 2019-03-20
Change gradle root project name to match the case of the git hub project

Jaggeroth:
Title says it all.
Change needed to simplify Eclipse Gradle Import process.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/351)

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -15,4 +15,4 @@ include 'api'
 include 'services:webservice'
 */
 
-rootProject.name = 'MapTool'
+rootProject.name = 'maptool'


JamzTheMan:
@Jaggeroth and @cwisniew 

Any reason to not just change the project name to MapTool? RPTools is already title cased and looks better and less changes required?

JamzTheMan:
FWIW since we had to revert this change, It doesn't look like we can reapply the PR? It needs to be opened again?

Jaggeroth:
Where are we with this change?
We reverted it because it caused non-windows OS to lose their icon.
What change needs to happen to make sure that doesn't happen?

PR #366, opened 2019-03-23
Catch exceptions when MapTool is executed by Java 11

Azhrei:
#350 describes MapTool crashing badly when executed by Java 11.  This PR cleans up the exception handling so a nicer error message is displayed.  Exception handling is added to cover MapTool.main(), but also inside that method there is handling just for JavaFX splash screen creation so that the old Swing approach can be used instead.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/366)


Azhrei:
Way too many commits shown above.  I'll resubmit.

Issue #102, opened 2016-06-25
Change the current campaign file format to a directory


rkathey:
Some benefits include
Ability to store database files
Ability to work on macros (both JS and MTScript outside MapTool

Darinth:
Is the interest here in removing the usage of zip files, or just splitting content.xml into multiple small files within the zip?

If we are we talking about entirely eliminating the usage of zip files so that a single campaign is actually multiple files rather than a single zipped file, then while I can see benefits to it, I'd actually disagree with the change on the grounds of 'keeping it simple for the average user'. One of the benefits of a single zip file is that A: users who don't know what they're doing can't make mistakes. B: Transfering the campaign file to another person is trivially easy.

The campaign files are already zip files and should be able to store database information and once you know that the campaign files are just zip files, you can actually open them up and work on macros outside of MT. What might make more sense (at least to me) is to start pulling some of the content out of content.xml (I believe that's the file that contains all of the campaign details) and move some of that into it's own files inside of the zip.

kayila:
I see no solid reason why we can't have the directory format and then zip the folter to get a single file campaign. Both can probably be supported

Jaggeroth:
Yes, I think the desire is to have both. But if you can explode a campaign file into a directory structure then the option of storing said campaign in a source control system becomes a very practical option.

Darinth:
I don't know that I ever would've considered storing a campaign file in a source control system, even though it makes perfect sense and is honestly a compelling reason to have the option of working on maptool campaigns as directories.

It doesn't seem to make a huge amount of sense to be always doing both all of the time, but at a minimum I acknowledge the benefits of having an option to save maptool campaigns into a directory structure over having them zipped.

oliver-szymanski:
Started coding this, will ping when I have a workable/testable solution

JamzTheMan:
I know @cwisniew had done some Proofs on this.

@cwisniew do you have any design requirements for this?

oliver-szymanski:
I got the save/load to work already.

Just need to check how/if to split content to multiple files.

kayila:
I thought the whole point of this was to split the content into different
files?

oliver-szymanski:
Sure, but first I had to start with saving/loading/exporting it as a directory. Works now. 

Now I'm at splitting in files.

oliver-szymanski:
All done, will test it a bit before creating the PR

oliver-szymanski:
[...]

PR #371, opened 2019-03-23
Load save split directories 102

oliver-szymanski:
This enhancement #102 allows you to:

- save files as a directory structure instead of a cmpgn (zipped) file
; therefore you use git or another source control system to manage them and you can open/check files with an editor (only experts should do the latter)
- it splits the big content.xml file in multiple files per campaign/zone/token/macros; make it easier when using a source control system to detect/manage changes
- it puts the actual macro code in a ".macro" plain text file, so you can edit the code easily with any editor if you like
- allows to copy whole zones or just tokens or macros between campaigns; e.g. have a base map with all your macros and use that from a linked shared folder. so you can have lots of campaign sharing code

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/371)


oliver-szymanski:
I know I could have rewrote the whole load/save mechanism, but I tried to have as less impact as possible to have no issues in export/save/save as/autosave/backup. So this solution is based on the existing save/load mechanisms

JamzTheMan:
@cwisniew Since we have an extra week (next release end of April?) do we want to merge this now and start reviewing/testing? Pending code review of course...

Azhrei:
Since Craig implied that revert wasn’t as clean as he thought it’d be, should we create a new branch off develop to test such a big change?  It would make it easier to remove later (just delete the branch).  We can still tag it so the CIs generate builds...

JamzTheMan:
Yea, good idea. I'll create a feature branch so we can at least run from source and test but not block/disrupt other 1.5.2 dev work.

@rkathey or @phergus Let me know if you need a RC for this but I think you guys can checkout the branch and run ok.

oliver-szymanski:
> Reviewed 14 of 14 files at r1.
> Reviewable status: all files reviewed, 3 unresolved discussions (waiting on @Azhrei, @cwisniew, and @oliver-szymanski)
> 
> 
> src/main/java/net/rptools/maptool/client/AppConstants.java, line 67 at r1 (raw file):
> 
>   public static final String CAMPAIGN_FILE_EXTENSION = ".cmpgn";
>   public static final String CAMPAIGN_DIRECTIORY_EXTENSION = "";
> 
> Do we want to use a extension of some sort? It may be helpful to denote everything in the campaign dir is for save only (but I assume you ignore other content added, like if we dumped a pdf or something in the directory?)
> Oh, ok, I see later you have the file filter as .cmpgndir so probably should set it here and use the constants in MapToolFrame below...
> 
> src/main/java/net/rptools/maptool/client/ui/MapToolFrame.java, line 236 at r1 (raw file):
> 
>       new MTFileFilter("cmpgn", I18N.getText("file.ext.cmpgn"));
>   private final FileFilter campaignDirectoryFilter =
>       new MTFileFilter("cmpgndir", I18N.getText("file.ext.cmpgnDirectory"), true);
> 
> We should use the defined constants here...
> 
> src/main/java/net/rptools/maptool/util/PersistenceUtil.java, line 557 at r1 (raw file):
> 
>     fileName = fileName.replace("'", "_");
>     fileName = fileName.replace("#", "_");
>     return fileName;
> 
> We don't have something like this already? Maybe in the lib? Could be a single regular expression.  There could also be one in Apache lib?

No idea. If you know I'm happy to change :)

Phergus:
@JamzTheMan No need for RC for initial testing.  Checkout and build from shell is fine.

JamzTheMan:
In MapToolFrame you have the filter set/hard coded to cmpgndir.  Although, is there a reason to just not use .cmpgn? It's the same extension BUT it is a dir vs file so it shouldn't confuse anything.

The downside of dir is we can never to file association with it (unless we through a file inside the dir that we associate with?)  Not a major concern, just popped in my head.

oliver-szymanski:
> In MapToolFrame you have the filter set/hard coded to cmpgndir.  Although, is there a reason to just not use .cmpgn? It's the same extension BUT it is a dir vs file so it shouldn't confuse anything.
> The downside of dir is we can never to file association with it (unless we through a file inside the dir that we associate with?)  Not a major concern, just popped in my head.

See answer in my response above. So basically the one in MTFileFilter is not used at all. I'm not sure if a suffix for directories is good or not, still a bit undecided. I like it on Mac, but there its much better supported to look as a file

Azhrei:
TL;DR on my phone, but an extension for top-level dir would be good on Mac since we can later set a file association for it (we don’t do that now, but I had plans for it).  However, not sure what happens if that extension is then used for a regular file!  I’ll test and reply here.

JamzTheMan:
> TL;DR on my phone, but an extension for top-level dir would be good on Mac since we can later set a file association for it (we don’t do that now, but I had plans for it). However, not sure what happens if that extension is then used for a regular file! I’ll test and reply here.

Yea, wasn't sure if association would work with directories?
If it does work, then technically we would want both files and dirs associated to MapTool and load appropriately anyway.

Javapackager supposedly supported associations but I couldn't figure it out (documentation sucks) and it may only work if you used ant config.  I was going to wait and see if the new jpackager supports it.

Azhrei:
On the Mac at least, there’s a script we can supply that the packager will use when building.  The script can do the work.  Worst case, a similar PowerShell script on Windows and Bash script on Linux...

Issue #299, opened 2019-03-12
Implement a cursor for dragged tokens


rkathey:
Currently, the cursor vanishes while dragging a token around the map.

At the very least it would show you where MapTool thinks the cursor is, which might make solving the drift a bit easier

Jaggeroth:
Making the cursor visible while draging has more implications than I expected.

If snap-to-grid is off, then its fine.

If snap-to-grid is on, then the token needs to move in sync with the cursor rather than jump from cell to cell. Otherwise it looks just plain odd. Which means you probably want a grayscale version of the token, showing where the snap-to-grid token will drop.

Anyone know of examples of good drag UI in other systems?

JamzTheMan:
Wait, doesn't it do that now?

JamzTheMan:
In other news, for 2.0 I'd like to change/add a new method of movement, and instead let you click a token the click where you want it to go. It helps the AI out and you can add in a "GM approval" stage.

I'd like to do a little more of a MMO style system like Pathfinder Kingmaker.

Azhrei:
I _really_ like the current drag scheme because I get to see the distance traveled.  If you want to switch to a click-to-start and click-to-end approach, I'll need mouse movement events to show the distance that will be traveled.

Phergus:
Don't we effectively have click-to-move already?  Right-click on a token.  Select Move and then position pointer where you want to go and click.  Shows path and distance without the normally spazzy behavior from using the mouse.

JamzTheMan:
Eh? Hmm, um. [Needs to load up MapTool]

And ya, there's a few options we code do. We could even turn on a "show cost on every cell" up to your "movement" so you could see where u could get in 1 move action equiv, accounting for Terrain.

But getting off topic, sorry.

Azhrei:
Lol, good point, Phergus.  But yes, off topic for "show cursor" issue...

Jaggeroth:
I intend to submit a change that does the following to token drag behaviour:
1) Cursor no longer disappears.
2) Snap-to-grid tokens will follow the cursor just like free position tokens. Users will know where the token will drop by the path shading.

This will make it easier ti diagnose Token Drift #300 
People thought this was how it behaved anyway ;)
I will leave finding a nice cursor to later (as most of ours are a bit long in the tooth now).

JamzTheMan:
There is code that does the "greyscale/shaded token" in the drag n drop event unless you want a generic cursor then? Are you going to hide the token why moving?

[...]

PR #372, opened 2019-03-24
Change token drag behaviour

Jaggeroth:
Change for Issue #299 
Cursor remains visible while dragging token.
Snap-to-grid tokens don't jump between cells.
Path indicates where token will be placed.

Please test. I think this is a case where we over-thought the behaviour. Its far easier to use like this imo.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/372)


Jaggeroth:
No code added. Deletes only :)

Issue #386, opened 2019-03-25
Command line java version no longer needed in debug info window


Phergus:
**Describe the bug**
As MapTool going forward will always be 64-bit the extra info provided by the command line `java -version` call is no longer needed and leads to confusion for users.

**To Reproduce**
1. Start MapTool
2. Go to Help -> Gather Debug Information...

**Expected behavior**
No longer will show the unneeded information.

Will also be slightly faster at showing the info window without the system call.

**Screenshots**
![image](https://user-images.githubusercontent.com/34379254/54954670-d0a42480-4f10-11e9-9a7e-01ba3f493a94.png)

**MapTool Info**
- Version: 1.5.1
- Install: Any

**Desktop (please complete the following information):**
 - OS: Any
 - Version: Any

Phergus:
@Azhrei pointed out that the max memory value being given was incorrectly formatted due to a bug introduced in Nerps fork.

JamzTheMan:
Hey I didn't introduce no bugs! I'm pretty sure I didn't introduce a bug. Did I introduce a bug?
"Hello bug!"

Joking aside, I never touched gather info other than to add the encoding section as far as I can remember.  If it changed it could be due to JVM changes from java 1.6 through java 10?

Phergus:
Somebody touched MemoryStatusBar.java and made a change to how the value was being saved but didn't notice that there was a public method used to access it.  ;)

No worries.  Easy fix.

JamzTheMan:
Man, who did that? You gotta watch that guy lol 😜

PR #387, opened 2019-03-25
Update SysInfo.java

Phergus:
Removed method to get the command line java -version info as it is no longer needed since all MT versions going forward are 64-bit.  Issue #386

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/387)


Azhrei:
While you’re in there, can you check the formatting of the memory sizes?  I’ve seen some second entries (Mem Used) with an M after them and three fields (like ###,###,###M).  Seems high for “used”.

If there’s a way to get the command line args, that would be nice (not JVM, just for the app).

JamzTheMan:
FYI: If you need it, there is an Apache lib util to format numbers in k/m/g (and it's smart enough to pick k/m/g for you IIRC).  I use it in the statusbar updates (search for that) where I get the Asset Cache size and display on the, well, status bar. :)  

If it just returns a long or something for the memory sized, this will be handy I think.

Phergus:
@Azhrei Roger that.
@JamzTheMan   That will be handy.

Issue #395, opened 2019-03-26
The macro editor is still broken


juan0286:
**Describe the bug**
The macro editor is still broken when try Edit any Macro.

**To Reproduce**
1. Open Map Tool
2. Edit any Button Macro

**Expected behavior**
View the MAcro Editor Windows

**MapTool Info**
- Version: 1.5.1
- Install: Upgrade [maptool-1.4.5.4]

**Desktop (please complete the following information):**
 - OS: [Windows and Linux Ubuntu ]
 - Version [W7, Ubuntu 14.04 LTS]
- Language: Spanish

**Additional context**
Only agree that I use the Spanish language both in OS, and in the map tool.

juan0286:
Error Detail: 

java.lang.RuntimeException: Didn't find the 'Edit' menu?
	at net.rptools.maptool.client.ui.MacroButtonDialog.createMenuBar(MacroButtonDialog.java:506)
	at net.rptools.maptool.client.ui.MacroButtonDialog.<init>(MacroButtonDialog.java:126)
	at net.rptools.maptool.client.ui.macrobuttons.buttons.MacroButtonPopupMenu$EditButtonAction.actionPerformed(MacroButtonPopupMenu.java:164)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2308)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:405)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
	at java.desktop/javax.swing.AbstractButton.doClick(AbstractButton.java:369)
	at de.muntjak.tinylookandfeel.TinyMenuItemUI.doClick(Unknown Source)
	at de.muntjak.tinylookandfeel.TinyMenuItemUI$MouseInputHandler.mouseReleased(Unknown Source)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6589)
	at java.desktop/javax.swing.JComponent.processMouseEvent(JComponent.java:3342)
	at java.desktop/java.awt.Component.processEvent(Component.java:6354)
	at java.desktop/java.awt.Container.processEvent(Container.java:2261)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:4966)
[...]

PR #396, opened 2019-03-26
Fix non-English Edit menu issue.

Phergus:
Fixes Issue #395 - Should have been using I18N call to get menu name.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/396)


Azhrei:
The menu updates were supposed to go away completely and be replaced with toggle buttons.  Is that what's in here?  Might as well do the tooltip field size increase since the abeille form will need editing anyway.  I think there was one other thing on the abeille form to update; should be in the issue list somewhere.

Phergus:
This only addresses the two exceptions.

Issue #416, opened 2019-04-02
Create a movedOverDrawing function


Jaggeroth:
**Is your feature request related to a problem? Please describe.**
The second part of @rkathey request to create a script method to determine if a token has moved over a drawing.

**Describe the solution you'd like**
Create a movedOverDrawing function that when given a drawing reference and a "Last Path" array of coords, returns a list of coords (if any) that cross the drawing.

This should cover instances of drawing that cannot be plotted by a simple list of coords, such as grouped drawing, ovals, thick lines etc.

**Describe alternatives you've considered**
We have produced a getDrawingInfo() function that returns coords for simple polygons that can be passed into the movedOverPoints() function.

**Additional context**
Linked Issue #407

Jaggeroth:
Pull Request #418

PR #418, opened 2019-04-04
Add movedOverDrawing() function

Jaggeroth:
Pull Request for Issue #416 

This creates a movedOverDrawing(mapName, drawingId, points) function.

`[h: lp = getLastPath()]<br><br>
[h: id = findDrawings(getCurrentMapName(),"fig5")]
[r: movedOverDrawing(getCurrentMapName(),id,lp)]`

This works as per the movedOverPoints() function and returns a JSON array with coordinates of all cells in the path that overlap with the area defined by the specified drawing. 

If the drawing happens to be a group, the function respects "cuts" and which won't be included in the drawing's area.

If the drawing is a fat line, the function triggers if the path point crosses the line's width.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/418)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/DrawingMiscFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/DrawingMiscFunctions.java
@@ -14,9 +14,16 @@
  */
 package net.rptools.maptool.client.functions;
 
+import java.awt.BasicStroke;
+import java.awt.Point;
+import java.awt.geom.Area;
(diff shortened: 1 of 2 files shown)


Jaggeroth:
Example campaign for testing. Remove the .zip
[movedOver.cmpgn.zip](https://github.com/RPTools/maptool/files/3045485/movedOver.cmpgn.zip)

image

Issue #422, opened 2019-04-07
Issue in used jgit lib locks files during Maptool build


oliver-szymanski:
As of issue https://github.com/ajoberstar/grgit/issues/288 that I just raised in the grgit library MT uses during build the gradle daemon process that stays alive in the background even after a build can lock ".pack" file in the ".git" repository.

This can be fixed in side the MT build by closing the underlying jgit repository that the grgit plugin uses. I will shortly commit a PR for that

oliver-szymanski:
PR in https://github.com/RPTools/maptool/pull/423

PR #423, opened 2019-04-07
Fix Issue in used jgit lib locks files during Maptool build #422

oliver-szymanski:
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/423)


oliver-szymanski:
Fixed #422

PR #439, opened 2019-04-16
Macro editor changes

JamzTheMan:
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/439)


Phergus:
Hotkey visibility checkbox is erratic.  If a hotkey has already been assigned, unchecking it doens't remove the hotkey from button label.    If I uncheck it and then assign a hotkey, the button is labeled correctly but the checkbox does not then work to enable display of the hotkey.

JamzTheMan:
Thanks, I will take a look.

PR #448, opened 2019-04-20
Change to JSONObject and JSONArray

Jaggeroth:
Switched to JSONObject and JSONArray to preserve attribute order.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/448)


Jaggeroth:
Well it makes it match the code of the other functions and match the attribute order.
:)

Issue #473, opened 2019-05-16
MapTool 1.5.2 NullPointerException when try to shutdown server.


adventuremagic123:
**Describe the bug**
I tried to shutdown the MapTool 1.5.2 server and got the following exception:
```
java.lang.NullPointerException
	at net.rptools.maptool.util.UPnPUtil.closePort(UPnPUtil.java:225)
	at net.rptools.maptool.client.MapTool.disconnect(MapTool.java:1139)
	at net.rptools.maptool.client.MapTool.stopServer(MapTool.java:1016)
	at net.rptools.maptool.client.AppActions.disconnectFromServer(AppActions.java:2280)
	at net.rptools.maptool.client.AppActions$61.execute(AppActions.java:2271)
	at net.rptools.maptool.client.AppActions$ClientAction.actionPerformed(AppActions.java:3089)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
        ...
```

**To Reproduce**
1. Start the MapTool 1.5.2 server with shared FoW.
2. Try to shutdown the server.
3. Get the above exception.

**Expected behavior**
Expected no error message.  Under the same scenario with a copy of the same campaign, MapTool 1.4.1.8 has no trouble.

**Screenshots**

**MapTool Info**
- Version: 1.5.2
- Install: New

**Desktop (please complete the following information):**
 - OS: Windows
 - Version 10

**Additional context**
==== MapTool Information ====
MapTool Version: 1.5.2
MapTool Home...: C:\Users\mark\.maptool-rptools
Max mem avail..: 3 GB
Max mem used...: 1 GB
Custom Property: -DMAPTOOL_DATADIR=.maptool-rptools

==== Java Information ====
Java Vendor.: Oracle Corporation
Java Home...: C:\Users\mark\AppData\Local\MapTool\runtime
Java Version: 10.0.1

==== OS Information ====
OS Name........: Windows 10
OS Version.....: 10.0
OS Architecture: amd64
PATH...........: C:\Users\mark\AppData\Local\MapTool\app;C:\Users\mark\AppData\Local\MapTool
Number of Procs: 4

==== User Information ====
User Name: mark
User Home: C:\Users\mark
User Dir.: C:\Users\mark\AppData\Local\MapTool\app

==== Network Interfaces ====
Display Name..: Software Loopback Interface 1
Interface Name: lo
Address...: 127.0.0.1
Address...: 0:0:0:0:0:0:0:1

[...]

PR #475, opened 2019-05-17
Fixed NPE when unmapping port

Phergus:
IGD was being removed from iterator while it was still being used.  Fixes NPE seen in #473.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/475)

### DIFF
--- a/src/main/java/net/rptools/maptool/util/UPnPUtil.java
+++ b/src/main/java/net/rptools/maptool/util/UPnPUtil.java
@@ -219,10 +219,10 @@ public class UPnPUtil {
           // NewPortMappingDescription=MapTool
           boolean unmapped = gd.deletePortMapping(null, port, "TCP");
           if (unmapped) {
-            iter.remove();
             count++;
             if (log.isInfoEnabled())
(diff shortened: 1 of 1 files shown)


Azhrei:
From looking at the code (not using Eclipse), I can't tell if calling `.remove()` on the iterator will invalidate it (it depends on the type of iterator, which relies on the type of the container).  I'm going to figure you've done that. :)

Phergus:
@Azhrei  Yup.  Stepped through the code and as soon as the remove() was called igds.get(gd) evaluates to null.

Azhrei:
Right, but does it invalidate the operator such that the loop then immediately terminates?  If that’s true, then the remove cannot be done inside the loop at all; the items to be removed must be accrued in a separate list and removed after the iterator  is no longer is use.

I’ll take a look at this but it’ll likely be tomorrow before I get a chance.

Phergus:
No, that doesn't invalidate or terminate the loop.  However, I don't know why the remove() is there at all.

Azhrei:
My mistake.  It's the Enumeration type that doesn't guarantee well-defined semantics on removal via an iterator.  "Nothing to see here...  Move along."

PR #486, opened 2019-06-08
[build] bump gradle to 5.4.1

grimreaper:
This also removes old instructions to run `gradlew wrapper` as the
wrapper is a part of the repository.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/486)


Azhrei:
Why do this?  I’m not a fan updating a version requirement without a reason.  I would prefer to leave the Java 9 references in the README as well, as we don’t currently _require_ a later version of Java to build and devs with gradle already installed need to be aware of the version dependency.

grimreaper:
> Why do this? I’m not a fan updating a version requirement without a reason. I would prefer to leave the Java 9 references in the README as well, as we don’t currently _require_ a later version of Java to build and devs with gradle already installed need to be aware of the version dependency.

This isn't really a version requirement so much as a dependency bump. Users with gradle installed shouldn't be affected by this _providing they are using the wrapper_ - which they should be anyways. Its not generally safe to run the global gradle installation on a project with a specific wrapper in mind.

As for why do this: mostly to pick up the performance and bugfixes in newer versions of gradle.

Azhrei:
> As for why do this: mostly to pick up the performance and bugfixes in newer versions of gradle.

Heh.  “Better the devil you know than the one you don’t.”

In other words, we’re not currently affected by any gradle bugs now (that I know of), and updating means finding new bugs that we’d then need to work around.

I’ll let someone else review this and decide whether they want to merge it.  I’m clearly biased. :)

cwisniew:
@grimreaper can you resubmit this pull request against the develop branch please.

JamzTheMan:
FYI: I created the createWrapper task so you can easily download and update the gradle wrapper to a new version. Also, makes it easy to see what wrapper version the build is currently packaging/using...

Issue #487, opened 2019-06-15
Restore previous token dragging behavior


Phergus:
**Describe the bug**
PR #372 changed the token drag behavior such that the token image doesn't snap to grid and was visibly off from the mouse pointer position by a significant amount.  While the movement path highlighting does indicate where the token would end up, this new behavior still causes confusion.  It also made it very easy to accidentally drop a token inside of VBL.

Token dragging needs to be restored to previous behavior.

**MapTool Info**
- Version: 1.5.2
- Install: Any

**Desktop:**
 - OS: Any
 - Version: Any

Jaggeroth:
I am against this change.
While some people might not like it, I have players that found the previous drag style very difficuilt to use.

Phergus:
Preference settings added to select between snapped and non-snapped dragging modes.  Code tweaked to keep token image better synced with mouse position.

Phergus:
No further feedback or comments.  Closing.

Issue #479, opened 2019-05-18
Token set to Snap to Grid no longer snaps to grid when moving


leowalter:
**Describe the bug**
Token set to Snap to Grid no longer snaps to grid when moving.  When moving the token, the token is moving like it is not snap to grid.  Accidental drop beyond VBL is common.

**Expected behavior**
When moving a snap to grid token, the token should snap to grid when moving.

**MapTool Info**
- Version: 1.5.2
- Install: New

**Desktop (please complete the following information):**
 - OS: Windows
 - Version [10, multiple versions.  All players experienced it.]

Phergus:
@leowalter  Please test with 1.5.3-rc3 (or later).

Phergus:
Fix is in 1.5.3.rc3.  Closing.

PR #488, opened 2019-06-15
Restore previous token dragging behavior.

Phergus:
Restores previous token dragging behavior by restoring code removed by PR #372 .  Commented out code which was causing tokens to frequently end up offset by one cell up and left.  Overall tracking of token image to mouse pointer was also improved by the removal of those lines.

Completes #487

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/488)


Phergus:
This also restores the snap-to-grid behavior for #479 .

Phergus:
New commit adds preference settings for _Snapping Token while dragging_ and _Hide Mouse Pointer while dragging_.
Additional tweaks to ZonePoint location for different grid types and snap VS non-snap.

PR #491, opened 2019-06-18
Update for current build requirements.

Phergus:
Removed references to older Java versions and 32-bit builds.
Included URL for downloading Java 10 JDK from Oracle.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/491)

### DIFF
--- a/doc/How_To_Install_JDK.md
+++ b/doc/How_To_Install_JDK.md
@@ -1,29 +1,15 @@
 How to Install Java Development Kit (JDK)
 =======================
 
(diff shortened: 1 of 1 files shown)


Azhrei:
If you can make it generic so it doesn't list specific filenames (like something ending with `.exe`), then it can apply to Linux and macOS downloads as well.  Either that, or the other platforms should be listed explicitly.

(Man, I can't believe the document was _that_ old!)

Phergus:
It was pretty old.  I had never looked at it before.  Changed to be more system agnostic.

Issue #504, opened 2019-07-24
getInfo("Server") missing some Server Policy info


Phergus:
**Describe the bug**
Two of the server settings are missing from getInfo("server"):
- Auto Reveal on Movement
- Individual FoW

**To Reproduce**
1. Start MapTool and start a server.
2. Run `<pre>[r: json.indent(getInfo("Server"),2)]</pre>` from chat command line
3. Note above settings are missing.

**Expected behavior**
Expect to see all settings from Start Server dialog.

**MapTool Info**
- Version: 1.3.b69+
- Install: Any

**Desktop (please complete the following information):**
 - OS: Any

Phergus:
@JamzTheMan  The code is done for this.  Do you want it in 1.5.3 or 1.5.4.

PR #515, opened 2019-07-29
Get info504

Phergus:
Adds the Auto Reveal on Movement and Use Individual FoW settings to the getInfo("server") results.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/515)


Phergus:
Resolves issue #504

PR #518, opened 2019-07-29
Fix #513 Null Pointer Exception on apply change to token macro

### LINKED ISSUE #513: Applying change on a macro from another map throws a null pointer exception
**Describe the bug**
Applying changes to a token macro from another map throw a null pointer exception. The error originates from save() in [MacroButtonProperties](https://github.com/RPTools/maptool/blob/develop/src/main/java/net/rptools/maptool/model/MacroButtonProperties.java), called by [MacroButtonDialog](https://github.com/RPTools/maptool/blob/develop/src/main/java/net/rptools/maptool/client/ui/MacroButtonDialog.java). [Full error message](https://textuploader.com/1148b).

A similar error occurs if the token is deleted when the edit macro window is open.

**To Reproduce**
1. Edit a macro on a token.
2. Change map or delete token.
3. Click "APPLY" or "OK" on the macro opened in 1.

**Expected behavior**
Changes are applied with no null pointer exception.

**MapTool Info**
- Version: 1.5.3
- Install: New,

 - OS: Windows
 - Version: 10

**Additional context**
Regardless of the error, all changes appear to be saved.

Merudo:
Fix issue caused by two functions related to macro edit window that were looking for token only on current map.

Token macro edit window can remain open even if the token is no longer on the current map. Fix so applying changes works for a token regardless of the map it is on.

EDIT: second commit also fixes Null Pointer Exception when applying changes to a token macro after the token is deleted.

Close #513

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/518)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/MacroButtonProperties.java
+++ b/src/main/java/net/rptools/maptool/model/MacroButtonProperties.java
@@ -31,6 +31,7 @@ import net.rptools.maptool.client.ui.MacroButtonHotKeyManager;
 import net.rptools.maptool.client.ui.macrobuttons.buttons.MacroButton;
 import net.rptools.maptool.client.ui.macrobuttons.buttons.MacroButtonPrefs;
 import net.rptools.maptool.client.ui.zone.ZoneRenderer;
+import net.rptools.maptool.language.I18N;
 import net.rptools.maptool.util.StringUtil;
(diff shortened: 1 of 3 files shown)


Merudo:
Added a new commit to correct a Null Pointer Exception when attempting to apply change to a deleted token.

PR #554, opened 2019-08-09
Fix opening&closing Edit Token Dialog turning on Visible over FoW

### LINKED ISSUE #551: Opening and closing the Edit Token dialog on Player Client turns on Visible over FoW for that token.
**Describe the bug**
The act of simply opening and closing the Edit Token dialog on a player client will cause the token to be set to Visible over FoW.  This can be seen on the GM side.

**To Reproduce**
1. Start a server and place a PC token on default map.
2. Connect with second MT instance as player.
3. Give player client ownership of token.
4. Verify that token does not have Visible over FoW set via the right-click token menu.
5. Open and then close (with OK button) Edit Token dialog on player client.
6. On server observe that token is now flagged as Visible over FoW

**Expected behavior**
Certainly didn't expect that to happen.

**MapTool Info**
- Version: 1.5.3
- Install: New

 - OS: Windows
 - Version: 10

**Additional context**
Was able to duplicate this effect several times as well with a combination of the use of setTokenSnapToGrid(1|0) from the server side which is not propagating the change to the clients causing them token to be out-of-sync between the two clients.  Appeared to be triggered by the use of the right-click menu on the Player side and setting Snap-To-Grid to on but couldn't do so reliably.  Could be a matter of forcing a token update from the client side.

Merudo:
Fix so that opening and closing the Edit Token dialog on Player Client no longer turns on Visible over FoW for that token.

Close #551.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/554)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/token/EditTokenDialog.java
+++ b/src/main/java/net/rptools/maptool/client/ui/token/EditTokenDialog.java
@@ -289,7 +289,6 @@ public class EditTokenDialog extends AbeillePanel<Token> {
     if (MapTool.getPlayer().isGM()) {
       tabbedPane.setEnabledAt(tabbedPane.indexOfTab("VBL"), true);
       getTokenVblPanel().setToken(token);
-      getAlwaysVisibleButton().setSelected(token.isAlwaysVisible());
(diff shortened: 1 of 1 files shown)


JamzTheMan:
Nice!

Issue #555, opened 2019-08-09
"set" functions to send only the relevant data to other clients (not the whole token)


Merudo:
**Is your feature request related to a problem? Please describe.**
Each time a function operating on a token is called, the whole token is sent to other clients through MapTool.serverCommand().putToken().

This can cause heavy slowdowns on tokens with a large number of properties.

Examples of functions affected: setProperty(), setState(), setPC(), etc.

**Describe the solution you'd like**
At least some of these functions should not send the whole token, and instead send the modification to be done on the token.

For example, setProperty() should only send the property and its new value, instead of the entire token.

JamzTheMan:
@CoreDevs should I create a feature branch to merge all the function/putToken changes into or just go straight into develop for testing? I don't think there is much other dev going on but if anyone is uneasy with these changes we can branch off for testing.

Phergus:
I have no significant objection to bring it straight into develop.

Azhrei:
Agreed.

Any structural changes (such as modifying how token updates work) should probably be done separately as I can see lots of little issues cropping up over time.

Phergus:
Following functions tested and are updating clients as expected.
- bringToFront
- moveToken
- removeTokenFacing
- resetProperty
- resetSize
- sendToBack
- setAllStates
- setGMNotes
- setLayer
- setLibProperty
- setNotes
- setNPC
- setOwnedByAll
- setOwner
- setPC
- setProperty
- setPropertyType
- setSize
- setState
- setTokenDrawOrder
- setTokenFacing
- setTokenHeight
- setTokenShape
- setTokenSnapToGrid
- setTokenWidth

JamzTheMan:
So these functions have all been updated? I'm trying to run some tests but it's hard with only 1 or 2 PC's connected.  I really need to do a test on Sunday with a bunch of friends for a full effect but I'll see what I can test.

JamzTheMan:
Having issues getting properties (getting prompt for missing vars in my campaign, not sure what is broken but something is...)

Phergus:
Those were all updated.

re: getting properties - I know changes were made to various get functions to allow getting properties from tokens on other maps.  Haven't had time to test any of that yet.

Merudo:
I found the bug regarding the missing vars.

I forgot a **break;** statement in my code, which results in the token getting a bogus property type when **setstate** is called. This would result in the default properties no longer being defined, resulting in the promp for missing vars.

[...]

PR #556, opened 2019-08-09
Change Functions to only send relevant data instead of whole token 

### LINKED ISSUE #552: Functions not propagating to other clients
**Describe the bug**
The following functions do not update the tokens on other clients:

- setGMNotes()
- setNotes()
- setTokenShape()
- setTokenWidth()
- setTokenHeight()
- setTokenSnapToGrid()
- resetSize()  (also not documented on Wiki)

**To Reproduce**
Use any of these functions on a token, then check the properties of the token from another client. 

The token will be unchanged on their client.

**Expected behavior**
The token is changed for every client.

**MapTool Info**
- Version: 1.5.3
- Install: New

 - OS: Windows
 - Version 10

**Additional context**
There was extensive discussion in #545 regarding a fix. The concern is that currently, tokens can only be updated by sending the entire token to other clients. This can cause performance issues.

It might be preferable to not fix these functions until a better system to update token changes is implemented.

Merudo:
- Stop the functions in TokenPropertyFunctions from sending the entire token to other clients
- Same for setState and setAllStates
- Same for setTokenDrawOrder and moveToken
- Add server commands to update token properties by sending only the Token method to call and its new value
- Fix #552, and partially fullfill #555
- The following functions are affected:

    setPC
    setNPC
    setLayer
    setSize
    resetSize
    resetProperty
    setProperty
    sendToBack
    bringToFront
    setPropertyType
    setTokenFacing
    removeTokenFacing
    setOwner
    setOwnedByAll
    setTokenWidth
    setTokenHeight
    setTokenShape
    setGMNotes
    setNotes
    setTokenSnapToGrid
    setLibProperty
    setState
    setAllStates
    setTokenDrawOrder
    moveToken

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/556)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ClientCommand.java
+++ b/src/main/java/net/rptools/maptool/client/ClientCommand.java
@@ -28,7 +28,7 @@ public class ClientCommand {
     getAsset,
     removeAsset,
     putToken,
-    setTokenProperty,
+    updateTokenProperty,
     removeToken,
     draw,
     clearAllDrawings,
--- a/src/main/java/net/rptools/maptool/client/ClientCommand.java
+++ b/src/main/java/net/rptools/maptool/client/ClientCommand.java
@@ -28,6 +28,7 @@ public class ClientCommand {
     getAsset,
     removeAsset,
     putToken,
+    setTokenProperty,
     removeToken,
     draw,
     clearAllDrawings,
--- a/src/main/java/net/rptools/maptool/client/ClientMethodHandler.java
+++ b/src/main/java/net/rptools/maptool/client/ClientMethodHandler.java
@@ -230,6 +230,17 @@ public class ClientMethodHandler extends AbstractMethodHandler {
                 MapTool.getFrame().refresh();
                 return;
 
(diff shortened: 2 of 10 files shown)


JamzTheMan:
I like what I'm seeing so far! If no one else does I'll look deeper into this over the weekend and merge if ok.

Time to add some logging to my campaign and start collecting some data!

Phergus:
A quick glance at the code but it looks good.

JamzTheMan:
@Merudo I think I found a bug testing today (btw, state changes are def faster! and overall looks to be an improvement)

For any functions that move/rotate/resize a token need to check/account for VBL on that token. The FoW is no longer updating if, say, moveToken is used to move said token with token VBL. Could check if the token has VBL (think there is a hasVBL method?) and if so then call the relevant FoW update server command?

Merudo:
Thanks a lot for the testing @JamzTheMan - I don't typically use macros on tokens with VBL so this bug flew right by me.

Issue #539, opened 2019-08-04
Need a getTableEntry() to go along with addTableEntry() and setTableEntry()


Phergus:
**Is your feature request related to a problem? Please describe.**
Currently no method exists to retrieve a specific table entry's details and get the min, max, value and image assetID for the entry.

**Describe the solution you'd like**
A getTableEntry(tableName, entryNum) macro function that would return the indicated entry as a JSON object.  Perhaps an optional Boolean parameter to indicate absolute entry versus the entry matching a roll of that value.

- getTableEntry("Animals", 3) - Would return entry 4 as a JSON object or an empty string if the requested entry doesn't exist.  An entry with no image returns an empty string for assetid.
  - Example JSON: {"min":3, "max":5, "value": "duck", "assetid":""}

Merudo:
Added requested function.

Phergus:
Wiki page added for [getTableEntry](http://www.lmwcs.com/rptools/wiki/getTableEntry)

Phergus:
Tested - working as requested.

Issue #538, opened 2019-08-04
Change tableImage() function to return an empty string if no image is found


Phergus:
**Is your feature request related to a problem? Please describe.**
Currently tableImage() throws a parser exception if there isn't an image for the requested entry. This aborts the current macro and gives the user no method to recover gracefully.

**Describe the solution you'd like**
Instead, tableImage() should return an empty string which can easily be checked for by the calling macro.

From discussion with Aranginor on Discord server.

Merudo:
aranginor, are you going to reopen your Pull Request?

aranginor:
No.

Phergus:
Tested. Change working as intended.

PR #559, opened 2019-08-10
Fixes for tableImage and getTableEntry

aranginor:
Issue #538 
Issue #539

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/559)


Phergus:
@aranginor  It looks like you didn't update the imports for the file.

Also Issue #539 is about adding a new macro function called _getTableEntry_ and it is supposed to return the details about the table entry not just a result from the entry.  See the issue description and note that to add a macro function you have to add the command name to the list at the top in the constructor.

Also you should run `gradlew spotlessApply` before committing code.

Issue #569, opened 2019-08-15
getPropertyDefault() and setLayer() undocumented parameters


Merudo:
**Describe your question**
The functions getPropertyDefault() and setLayer() have undocumented parameters.

getPropertyDefault() usually returns the default value for the specified property of the current token. However, if the second argument is specified, it instead returns the default value for the given token type (such as "Basic"), and the current token is ignored.

setLayer() takes a third argument (after token) called "forceShape". If set to 0, the setLayer() will not change the shape of the token.

I was wondering what I should do with these extra parameters?

Bug:  NPE error if getPropertyDefault() is used with one parameter and there is no current token

Phergus:
These should get documented on the wiki including examples of their use.

Merudo:
Sure.

About setLayer(), should the extra parameter be moved around? Right now, it's after tokenId, which goes against the convention for the other functions.

Phergus:
Oooh, that's unfortunate.   I suspect that change would break quite a few existing macros so we'll leave it that way.

Merudo:
Updated the documentation:

https://lmwcs.com/rptools/wiki/getPropertyDefault
https://lmwcs.com/rptools/wiki/setLayer

PR #572, opened 2019-08-15
Fix minor bugs in setSnapToGrid() and getPropertyDefault()

Merudo:
- Fix null exception error if getPropertyDefault() is used with one parameter and there is no current token.  Issue #569 
- Fix setTokenSnapToGrid() returning true/false instead of 1/0

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/572)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
@@ -388,11 +388,15 @@ public class TokenPropertyFunctions extends AbstractFunction {
      */
     if (functionName.equals("getPropertyDefault")) {
       checkNumberOfParameters(functionName, parameters, 1, 2);
-      Token token = resolver.getTokenInContext();
       String name = parameters.get(0).toString();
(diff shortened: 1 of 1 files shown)


Phergus:
@Azhrei , @JamzTheMan , @aliasmask , @Jaggeroth 
Any problems with changing the return value of setTokenSnapToGrid() from "true"|"false" to 1|0?

Need to tie the change to an issue.

JamzTheMan:
Other than break existing campaigns I'm ok with it. It should be minor, probably not a lot of checks for the return value of a set function I would think.

aliasmask:
Not sure if I even had the need to use that function. I doubt there will be much if any impact.

Phergus:
Yeah.  Pretty sure I've never had reason to use it and I expect the impact will be minimal if noticed at all.

Merudo:
I just realized setTokenSnapToGrid() doesn't actually return anything in 1.5.3, the true/false return is actually something I added in one of my commits.

Phergus:
Then it definitely won't have any impact.  :)

PR #607, opened 2019-08-22
Fix token names showing under FoW when Vision or Auto-Reveal are off

### LINKED ISSUE #553: NPC token names visible over Fog of War.
**Describe the bug**
The token names for GM controlled tokens under FoW show up if a player client mouse pointer hovers over a token's position or by enabled Show Token Names in View menu.

Reported by Dorpond via Discord.

**Edit for additional detail:**

- Note that it shows up most easily with the Vision for the map set to Off.
- It will also show up with Vision set to Day, if there is no VBL in the way and the player token is in sight range.
- For maps set to Night, the same rules as Day applies plus there needs to be light sources by the tokens that are under the fog.

**To Reproduce**
1. Load attached campaign. (At bottom of issue.)
2. Start Server - Strict Token Ownership, Use Individual Views, Players Can Reveal Vision
3. Start player instance of MT and connect with name Hero.
4. Hover mouse pointer over coordinates:  (2,4), (2,7) and (4,6) and see token names appear.
5. Select Show Token Names from View menu and all the hidden token names appear.

**Expected behavior**
Expect that players would not be able to see the names of tokens that they can't actually see.

**Screenshots**
GM View
Bug GM View
Player View hovering over token with mouse
Bug Mouse Hover
Player View - View Token Names enabled
Bug View Token Names

**MapTool Info**
- Version: 1.5.1, 1.5.3, Nerps 1.4.5.3
- Install: Any

 - OS:  Windows
 - Version:  10

**Additional context**
MapTool 1.5.3 campaign - remove ".zip" ending.
[VisibleUnderFogTest153.cmpgn.zip](https://github.com/RPTools/maptool/files/3484691/VisibleUnderFogTest153.cmpgn.zip)

Merudo:
- Add a check with zone.isTokenVisible()
- The operation is slow, but only runs if "Show Token Names" is enabled or the token is under the mouse
- Only affects cases where either Vision or Auto-Reveal are off
- Close #553

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/607)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
@@ -3439,6 +3439,15 @@ public class ZoneRenderer extends JComponent
 
       // Token names and labels
       boolean showCurrentTokenLabel = AppState.isShowTokenNames() || token == tokenUnderMouse;
+
+      // if policy does not auto-reveal FoW, check if fog covers the token (slow)
+      if (showCurrentTokenLabel
+          && !isGMView
+          && (!zoneView.isUsingVision() || !MapTool.getServerPolicy().isAutoRevealOnMovement())) {
+        if (!zone.isTokenVisible(token)) {
+          showCurrentTokenLabel = false;
+        }
+      }
       if (showCurrentTokenLabel) {
         GUID tokId = token.getId();
         int offset = 3; // Keep it from tramping on the token border.


Azhrei:
If the operation is slow, and if the visibility of the token was checked previously, can it be cached and reused?  I looked at the Reviewable code but hard to do on a phone!

Merudo:
I found a better way. Will force push an update in the next few minutes.

Merudo:
I must admit I don't really know how the cache system works. Hopefully someone more familiar can look at the code.

Phergus:
There is code around 3464 that does a check in the labelRenderingCache.  Maybe that's not being updated properly somewhere else.

Merudo:
Good idea! There could be a check to skip the zone.isTokenVisible() when 

`view.equals(lastView) && !labelRenderingCache.containsKey(tokId)`

Azhrei:
Depend on what the "status" change is that you're talking about.  If the token changed from PC to NPC that would require the cache be invalidated for that token.  But the cache is probably created fresh on ever pass of the renderer, so likely not an issue anyway.

Merudo:
I don't feel confident implementing the caching, so I will stop working on this for now.

If someone finds a better way, they can amend my PR or make their own PR.

JamzTheMan:
Lets give this a go and test. I'll build a RC today.

PR #610, opened 2019-08-23
Debounce/throttle repaint() calls in ZoneRenderer

pnichols04:
This PR would modify the several locations in ZoneRenderer that currently call repaint().  Often these get called in very rapid succession, which tends toward a cascade of event dispatches and conquest switches that can make the program unresponsive.

The PR adds a class, DebounceExecutor, that debounces or throttles serial calls to a initiate a task by scheduling the task to run at most once, after a specified cool-down or aggregation interval.  Calls made to initiate that task during the interval are ignored.

With the change here, ZoneRenderer waits 33ms (the duration of one frame at 30fps) from the time repaint() is first required until it is executed.  During that time, any other changes to the backbuffer can accumulate.  All changes are then painted to the screen at once, opening the performance bottleneck caused by having each part of the rendering code call repaint() as that part becomes ready.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/610)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/DebounceExecutor.java
+++ b/src/main/java/net/rptools/maptool/client/DebounceExecutor.java
@@ -1,0 +1,131 @@
+/*
+ * This software Copyright by the RPTools.net development team, and
+ * licensed under the Affero GPL Version 3 or, at your option, any later
+ * version.
+ *
+ * MapTool Source Code is distributed in the hope that it will be
+ * useful, but WITHOUT ANY WARRANTY; without even the implied warranty
+ * of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * You should have received a copy of the GNU Affero General Public
+ * License * along with this source Code.  If not, please visit
+ * <http://www.gnu.org/licenses/> and specifically the Affero license
+ * text at <http://www.gnu.org/licenses/agpl.html>.
+ */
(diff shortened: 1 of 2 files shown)


Phergus:
You need to run `gradlew spotlessApply` on your code before submitting and it should be against the `develop` branch.

Azhrei:
I have a couple of comments, although I want to start by saying "good job", as I think performance tweaks like this are a great use of developer/contributor time!

First, can the Executor thread pool (only a single thread in this case) be give a name?  One of my pet peeves with MT code is that there are a bunch of threads created and none of them are named.  This makes it difficult to look at VisualVM or other tools and determine _why_ the cpu is being used (a half-dozen threads, none of them named, each consuming varying amounts of cpu — why?).  So it's strictly from a debugging standpoint, but I'd definitely appreciate it.  I'm not familiar enough with the concurrency APIs enough to know off the top of my head.

Second, it was a good catch to notice that repaint() is called from all of these various locations.  But maybe all that's needed is to change the call to invalidate() instead?  Then Swing decides when to repaint — probably by calling it when there's nothing else in the event dispatching queue — and then the complexity of using a timer goes away.

I also didn't see that the Executor runs its work in the EDT — does it?  (This is just me being ignorant. :))

Phergus:
@Azhrei  - He closed this PR and opened #611

PR #611, opened 2019-08-23
Debounce/throttle rapid ZoneRenderer calls to repaint()

pnichols04:
This PR would modify the several locations in ZoneRenderer that currently call repaint(). Often these get called in very rapid succession, which tends toward a cascade of event dispatches and conquest switches that can make the program unresponsive.

The PR adds a class, DebounceExecutor, that debounces or throttles serial calls to a initiate a task by scheduling the task to run at most once, after a specified cool-down or aggregation interval. Calls made to initiate that task during the interval are ignored.

With the change here, ZoneRenderer waits 33ms (the duration of one frame at 30fps) from the time repaint() is first required until it is executed. During that time, any other changes to the backbuffer can accumulate. All changes are then painted to the screen at once, opening the performance bottleneck caused by having each part of the rendering code call repaint() as that part becomes ready.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/611)


JamzTheMan:
Thanks for the PR!  We're wrapping up 1.5.4 right now and going to do a RC today so I'll post pone this change to after that. There's a lot of changes already for this release and it will be better if this goes into the next release IMO. In case there are issues that come up, it'll be easier to isolate.

We do ~2 week sprints with a sprint for development then a sprint for QA/bug fixes/release.

Phergus:
There is some weird stuff from Mar 30th being pulled in with this.  Leftovers from the master branch?

Azhrei:
(Sigh.  Posted in the wrong PR.)

I have a couple of comments, although I want to start by saying "good job", as I think performance tweaks like this are a great use of developer/contributor time!

First, can the Executor thread pool (only a single thread in this case) be give a name?  One of my pet peeves with MT code is that there are a bunch of threads created and none of them are named.  This makes it difficult to look at VisualVM or other tools and determine _why_ the cpu is being used (a half-dozen threads, none of them named, each consuming varying amounts of cpu — why?).  So it's strictly from a debugging standpoint, but I'd definitely appreciate it.  I'm not familiar enough with the concurrency APIs enough to know off the top of my head.

Second, it was a good catch to notice that repaint() is called from all of these various locations.  But maybe all that's needed is to change the call to invalidate() instead?  Then Swing decides when to repaint — probably by calling it when there's nothing else in the event dispatching queue — and then the complexity of using a timer goes away.

I also didn't see that the Executor runs its work in the EDT — does it?  (This is just me being ignorant. :))

JamzTheMan:
@pnichols04 if you get a chance, tcan you do a pull from develop to your local and clean up any conflicts then push again to update this PR?

pnichols04:
> First, can the Executor thread pool (only a single thread in this case) be give a name? . . . it's strictly from a debugging standpoint . . . 

You don't have to say "from a debugging standpoint" twice to somebody who probably just given himself new gray hairs trying to figure out how ZoneRenderer works.

> Second, it was a good catch to notice that repaint() is called from all of these various locations. But maybe all that's needed is to change the call to invalidate() instead? Then Swing decides when to repaint — probably by calling it when there's nothing else in the event dispatching queue — and then the complexity of using a timer goes away.

I'm investigating this, though I fear that it would have surfaced in the research I did to implement this kind of solution.

> I also didn't see that the Executor runs its work in the EDT — does it?

There are two pieces of work that you might be referring to.  The actual, eventual call to repaint() (via the Runnable that DebounceExecutor gets initialized with) occurs on a thread received from the SingleThreadScheduledExecutor.  The logic in dispatch(), which decides to either execute, defer, or ignore, happen on the thread they were sourced from.  This represents the vast majority of the calls, and these, generally speaking, come from the application-initiated repaint notifications (the ones at the end of the various renderXXX() methods in ZoneRenderer).  Of the few calls that run on the EDT, they arise from system-initiated repaint requests (restore from minimized, resize window, and handleDragAndDrop are representative examples).

pnichols04:
I'll make a new commit with those conflicts cleared.

pnichols04:
> There really shouldn't be any changes to CHANGE_LOG.md, README.pod or CommandPanel.java. None of the changes are part of the work being done for debounce.
> 
> > _[Reviewable](https://reviewable.io/reviews/rptools/maptool/611#-:-Ln2nbEMBBWU21LYTZhG:b-x8qk7y)_ status: 0 of 5 files reviewed, all discussions resolved (waiting on @Azhrei and @cwisniew)

When I committed 4363de1 (bringing my local up to develop) only ZoneRenderer and DebounceExecutor showed differences.  Those are the only two files that should be getting touched here.  I'll get it straightened out.

Phergus:
That's what I would have expected.  I think somehow these other "changes" got pulled over from the Master branch.

JamzTheMan:
You should be able to do a ```git pull upstream develop``` or such depending on you setup. (I have origin as my fork and upstream defined as RPTools)

It's always good to pull from our develop branch before you do your push and create a PR so it's up to date and the CI will be happy.

Merudo:
Yeah, with git I would do the following:

```
git fetch upstream develop
git rebase upstream/develop
git push -f
```

JamzTheMan:
@pnichols04 Can you update and clean up this PR? (for instance, why was comandPanel changed?)

Once cleaned up we can merge it into develop branch. I've tentatively slated this for the 1.5.5 release.

Merudo:
I would also recommend changing the magic number 33 to some explicitly defined constant.

Azhrei:
I can't find an issue related to this at the moment, but I thought I had commented somewhere that perhaps using `invalidate()` instead of `repaint()` might have accomplished the same thing but without the extra code requiring a timer.  (Was it my imagination or did I type that in somewhere?)  My thinking is that invalidate() will flag a repaint as needed, but it won't do it until the EDT becomes idle.

Anyway, you've identified where repaint is called — if you leave out the rest of your changes and just put in a call to invalidate, how does it perform?

Thanks, btw.  I said this above, but I really like these kinds of tweaks. :)

JamzTheMan:
Would def be worth testing with invalidate...

Phergus:
@JamzTheMan   Looking at the history at the top - unless I'm misinterpreting what it says - it looks like you are the one that brought in the changes to CommandPanel.java.

No idea how he ended up with changes to the Changelog which are clearly wrong and should just be reverted back to the current develop branch version.  Same for the ReadMe.pod file.

@Azhrei  You made your comments about invalidate() here in this pr.  And the associated issue is #610

JamzTheMan:
Looks like it was changed by @kayila in 1.5.2. I only did the merge.

I suspect he built this PR off old code and this undoes that change...

The change log is probably also messed up for sale reason...

Probably should just test invalidate first why way.

CHANGE_LOG.md
```diff
@@ -253,6 +255,7 @@ Enhancements
 [i429]: https://github.com/RPTools/maptool/issues/429
 [i441]: https://github.com/RPTools/maptool/issues/441
 
+>>>>>>> upstream/develop
```
Merudo:
```suggestion
```

CHANGE_LOG.md
```diff
@@ -1,3 +1,5 @@
+<<<<<<< HEAD
+=======
```
Merudo:
```suggestion
```

CHANGE_LOG.md
```diff
@@ -1,3 +1,5 @@
+<<<<<<< HEAD
```
Merudo:
```suggestion
```

src/main/java/net/rptools/maptool/client/ui/commandpanel/CommandPanel.java
```diff
@@ -527,7 +527,7 @@ public void cancelCommand() {
   /** Inserts a newline into the chat input box. */
   public void insertNewline() {
     String text = commandTextArea.getText();
-    commandTextArea.replaceSelection("\n");
+    commandTextArea.setText(text + "\n");
```
Merudo:
```suggestion
commandTextArea.replaceSelection("\n");
```

pnichols04:
@Azhrei 
> I can't find an issue related to this at the moment, but I thought I had commented somewhere that perhaps using `invalidate()` instead of `repaint()` might have accomplished the same thing but without the extra code requiring a timer. (Was it my imagination or did I type that in somewhere?) My thinking is that invalidate() will flag a repaint as needed, but it won't do it until the EDT becomes idle.

Calling invalidate() in all those places takes things would be problematic.  First, since invalidate() is meant for layout changes, it walks up the component tree firing it on each parent, grandparent, etc. component it finds, until and unless it finds one that's already invalid.  So we could be asking the whole window to repaint, repeatedly and in rapid succession, when most of that window hasn't changed.  Second, invalidate() enters a critical section to do its thing.  The code in that section is minimal, but our repaint() fires so often and from enough threads that I think it would run a risk of livelocking the whole component tree.

That said, my debouncer only treats a symptom; what really needs to be done is to see why those 26 calls to repaint() are all even there and why they fire so fast.  Maybe some kind of non-locking monitor is a better solution.

Azhrei:
(On my phone; excuse typos, please!)

Hm, good point on the layout thing.  I’m not sure it matters too much, though, since ZoneRenderer doesn’t use Swing’s layout managers anyway.  And if ZoneRenderer were made a “root validator” (or whatever it’s called; maybe just an empty LayoutManager with stubbed out methods?), then the request wouldn’t go any higher in the tree.

Swing is a complex beast and working within the framework seems better to me than trying to impose order onto it from the outside...

pnichols04:
> (On my phone; excuse typos, please!)
> 
> Hm, good point on the layout thing. I’m not sure it matters too much, though, since ZoneRenderer doesn’t use Swing’s layout managers anyway. And if ZoneRenderer were made a “root validator” (or whatever it’s called; maybe just an empty LayoutManager with stubbed out methods?), then the request wouldn’t go any higher in the tree.
> 
> Swing is a complex beast and working within the framework seems better to me than trying to impose order onto it from the outside...

I'm happy to code it up and see what happens.  The more I look at the code, though, the more I think the answer is behind door number 3: figure out why repaint() gets called so often and, in each case, if it really should.  I mean, the whole thing doesn't really become an issue anyway until the map starts using more features (fog, exposed areas, VBL, number of tokens), at which point those features far and away dominate the matter.

pnichols04:
Using invalidate(), it's impossible to load a campaign file.  Some time after the window becomes disabled and before it would become re-enabled, the window just stops updating.  Everything is still working behind the veil, because invalidate() is still being called, but paintComponent() never gets called again.

Azhrei:
> Using invalidate(), it's impossible to load a campaign file. Some time after the window becomes disabled and before it would become re-enabled, the window just stops updating. Everything is still working behind the veil, because invalidate() is still being called, but paintComponent() never gets called again.

Yeah, as postulated above, the docs for these calls (`invalidate()`, `revalidate()`, and `validate()`) are based on component layout and ZoneRenderer doesn't define a layout manager.

One solution would be to have a layout manager stub that sets a flag whenever `invalidate()` is called, then when the event queue is empty, call `repaint()`.  We already hook into the dispatch queue to add exception handling for logging purposes.

But this doesn't feel any better than your solution with the timer.  Sigh.

So I guess I'm fine with the original approach of replacing repaint with a custom call and using a timer to call repaint instead.

Thanks again!

pnichols04:
It's lipstick on a pig, I admit.  I mean, the program's not a first-person shooter; how many independent things can be happening in 33ms that it generates five and six calls to repaint over that period of time?  It's a tough situation, though, because that one component is doing so much by itself that it's difficult to even trace the dependencies.

Merudo:
@pnichols04 any hope you can clean this a little bit? Remove the changes to `CHANGE_LOG.md`, `README.pod` and `CommandPanel.java`, rebase the branch, then squash the commits into 1.

pnichols04:
> 
> 
> @pnichols04 any hope you can clean this up? Remove the changes to `CHANGE_LOG.md`, `README.pod` and `CommandPanel.java`, rebase the branch, then squash the commits into 1.
> 
> Right now it's one of the most messy PRs I have ever seen.

I think I've got it cleaned up now--1 commit, 2 files changed.  Sorry for the mess.

PR #641, opened 2019-09-01
Change wrong nb of parameters error messages to display function name

### LINKED ISSUE #629: More helpful error messages for incorrect # of parameters
Error messages for incorrect number of parameters can be unnecessarily cryptic. Example:

`listGet()` and `input()` both give the same message:

`Invalid number of parameters 0, expected at least 1 parameters`

which doesn't tell the user the name of the problematic function. Similarly, `hasSight(1,2,3,4)` returns

`Invalid number of parameters 4, expected between 0 and 3 parameters`

The error message gives the name of the function that caused the error. For example, `listGet()` would return

`Function "listGet" requires at least 1 parameters; 0 were provided.`

which is immensely more helpful for debugging.

**Additional context**
Issue is caused by [checkParameters](https://github.com/RPTools/parser/blob/9d8e339428adec1d6827cba756289c1cc84a92eb/src/main/java/net/rptools/parser/function/AbstractFunction.java#L50) in the Parser not displaying the name of the function.

I think we should either modify `checkParameters` so that it displays the function name, or remove the call entirely and let each individual function deal with checking the number of parameters.

The later may make more sense, since each aliases in a function class needs a different number of parameters.

Related issue: https://github.com/RPTools/maptool/issues/613

Merudo:
- Change wrong number of parameters error messages to always display name of function
- Needs https://github.com/RPTools/parser/pull/3 to work
- Close #629
- Example: for `[input()]`

Old error message:

> Invalid number of parameters 0, expected exactly 1 parameter(s)

New error message:

> Function 'input' requires exactly 1 parameters; 0 were provided.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/641)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -177,7 +177,7 @@ dependencies {
     implementation 'net.rptools.decktool:decktool:1.0.b1'
 
     implementation 'net.rptools.maptool.resource:maptool.resource:1.0.b18'
-    implementation 'net.rptools.parser:parser:1.4.0.+'
+    implementation 'com.github.RPTools:parser:master-SNAPSHOT'
 
     implementation 'jide-common:jide-common:3.2.3'
     implementation 'jide-components:jide-components:3.2.3'
--- a/src/main/java/net/rptools/maptool/client/functions/InputFunction.java
+++ b/src/main/java/net/rptools/maptool/client/functions/InputFunction.java
@@ -1181,8 +1181,9 @@ public class InputFunction extends AbstractFunction {
   }
 
   @Override
-  public void checkParameters(List<Object> parameters) throws ParameterException {
-    super.checkParameters(parameters);
+  public void checkParameters(String functionName, List<Object> parameters)
+      throws ParameterException {
+    super.checkParameters(functionName, parameters);
 
     for (Object param : parameters) {
       if (!(param instanceof String))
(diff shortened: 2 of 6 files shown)


Merudo:
Before https://github.com/RPTools/parser/pull/3 gets merged, you can test out this branch by getting it, then changing in `build.gradle` the line

`implementation 'net.rptools.parser:parser:1.4.0.+'`

to

`implementation 'com.github.Merudo:parser:1.5.5'`

Merudo:
@cwisniew approved the parser merge, so this works fine now.

Merudo:
I'll ask @cwisniew to release a new version of the parser, so it will be cleaner.

Phergus:
It may be that the way you did it is more flexible for future builds.  @cwisniew ?  @JamzTheMan ?

JamzTheMan:
I created a 1.5.4 release and is now available, give it a try.

@Phergus All we need to do is a GitHub "Release" off master for Parser. JitPack does the rest and you can vew them here: https://jitpack.io/#RPTools/parser/1.5.4

It's actually pretty easy. There's no CI right now for Parser. And ya, we should only use MASTER--Snapshot for testing.  Since @cwisniew approved it I assume we are all good to go.

Phergus:
Roger that.   Merging on it then.

JamzTheMan:
Oh oh, um, sorry, guess I wasn't clear.  @Merudo (or anyone) still need to update the gradle file here to point to the parser-1.5.4 release vs master snapshot...

I only did a "Parser" release which created a new release to be used... Another PR will need to be created to update the gradle.build

Merudo:
PR #659 changes the parser to 1.5.4

Phergus:
My bad.  I misread your post and didn't catch the "for testing".

Issue #612, opened 2019-08-23
Functions json.get.path and json.set.path


Merudo:
**Is your feature request related to a problem? Please describe.**
It can be slow and cumbersome to deal with nested jsons.

For example, to get a "Level" field from a json, I might need

```
[att = json.get(json, "PrimaryAttributes")]
[st = json.get(att, "ST")]
[stLevel = json.get(st, "Level")]
```
The situation is worse when I need to set a property deep inside a nested json, in which case I need to update each level of the json.

**Describe the solution you'd like**
Functions json.get.path() and json.set.path() to automate the process.

For example, in my example I would only need

```[stLevel = json.get.path(json, "PrimaryAttributes",  "ST", "LEVEL")]```

Similarly for set.path, to change the Level to 5 (say),

```[stLevel = json.set.path(json, 5, "PrimaryAttributes",  "ST", "LEVEL")]```

Azhrei:
This is a great proposal, but I'd like to see good error messages for when the user messes it up!  Suppose the JSON looks like:

    { name: { first: "Bob", last: "Smith" },
      "class": [ "cleric", "fighter", "cleric", "cleric", "fighter" ],
      hp:  [ 5, 9, 6, 4, 8 ],
      ...
    }

To get to the number of HP for the third level of cleric it would be:

    JSON.get.path(json, "hp", 3)

Because the first level is an object and the second level is an array.

Giving a number instead of a string can work for an object, but not for an array, so the error message on arrays is pretty easy:  "cannot access array element using a string in JSON.get.path() in parameter 3", for example.

Otherwise, I think this is great.  It would make accessing JSON data much simpler (at least, until such time as JSON becomes a first class data type).

JamzTheMan:
I approve. I know a lot of JSON libs have a json.path built in. I haven't checked to see if the lib(s) we use have it or not.

They may or may not have standard error responses for errors.

Merudo:
I don't think the package net.sf.json has json.path, unfortunately.

Regarding error messages, are there cases where we would prefer to return "" instead of an error?

aliasmask:
This is the code I use in MapTool to json.set.path(), json.append.path() and json.get.path(). I was hoping if this was done in java there would be an easier method or at the very least be quicker in java.


`@@ @json.set.path

[H: sub_jsonName = arg(0)]
[H: sub_path = arg(1)]
[H: sub_key = arg(2)]
[H: sub_value = arg(3)]

[...]

PR #651, opened 2019-09-03
Add function json.get.path to return element from nested json

Merudo:
- First half of #612
- json.get.path returns appropriate error messages if path is invalid
- json.get.path can take two forms:

1. json.get.path(json, path)

where json is a json object or array, and path is a json array where each element specify a subpath (ex: ["Equipment", 1, "Name"])

2. json.get.path(json, subpath1, subpath2, ...)

for example, json.get.path(json, "Equipment", 1, "Name")

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/651)


Merudo:
Closed. I will do a PR for a different implementation using jayway.

PR #668, opened 2019-09-08
Add playStream, stopStream, and editStream functions to stream audio

### LINKED ISSUE #667: Add playStream / stopStream functions to stream audio from URL
Music can be immensely useful for setting the right atmosphere. It would be great to have a function to stream music from an URL or a local file.

New function: `playStream(uri, cycleCount, volume)` where:

`uri`: the url or uri path to the sound, example: 

`[playStream("http://mywebsite/mysong.mp3")]`
`[playStream("file:///C:/mysong.mp3")]`

`cycleCount`: how many times to play the resource. If -1, repeats indefinitely
`volume`: value from 0 to 1 determining the volume level

New function: `editStream(uri, cycleCount, volume)` to change the cycleCount or volume of stream.

Same parameters as `playStream()`

New function: `stopStream(uri, remove)` to stop a stream

`uri`: like above, if not provided or if set to "*" stops all sounds
`remove`: if 1 will remove stream from memory

There are other ways to implement sound (check #615 for an extended discussion), but regardless of what other features we implement, streaming sound from a server should always be useful.

**Additional context**
A more narrow version of #615

Merudo:
- Add playStream(uri, cycleCount, volume) to play audio from url or local file.
- cycleCount: number of times to play the resource. -1 for infinite loop. Default: 1.
- volume: volume value from 0-1. Default: 1.
- Add stopStream(uri, remove) to stop a stream. If remove is set to 1, unload the stream from memory. Default: 1. If no uri specified, stop all sounds.
- Add editStream(uri, cycleCount, volume) to change the cycleCount or volume of stream.
- Add AppPreference playStreams to enable/disable playing streams
- Ex:

[playStream("https://www.fesliyanstudios.com/musicfiles/2019-05-01_-_Undercover_Spy_Agent_-_David_Fesliyan.mp3", -1, 0.8)]

- Close #667

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/668)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppPreferences.java
+++ b/src/main/java/net/rptools/maptool/client/AppPreferences.java
@@ -354,6 +354,9 @@ public class AppPreferences {
   private static final String KEY_PLAY_SYSTEM_SOUNDS = "playSystemSounds";
   private static final boolean DEFAULT_PLAY_SYSTEM_SOUNDS = true;
 
+  private static final String KEY_PLAY_STREAMS = "playStreams";
+  private static final boolean DEFAULT_PLAY_STREAMS = true;
+
   public static void setHaloLineWidth(int size) {
     prefs.putInt(KEY_HALO_LINE_WIDTH, size);
   }
@@ -574,8 +577,16 @@ public class AppPreferences {
     prefs.putBoolean(KEY_PLAY_SYSTEM_SOUNDS, play);
   }
 
+  public static void setPlayStreams(boolean play) {
+    prefs.putBoolean(KEY_PLAY_STREAMS, play);
+  }
+
   public static boolean getPlaySystemSounds() {
     return prefs.getBoolean(KEY_PLAY_SYSTEM_SOUNDS, DEFAULT_PLAY_SYSTEM_SOUNDS);
+  }
+
+  public static boolean getPlayStreams() {
+    return prefs.getBoolean(KEY_PLAY_STREAMS, DEFAULT_PLAY_STREAMS);
   }
 
   public static void setPlaySystemSoundsOnlyWhenNotFocused(boolean play) {
--- a/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
+++ b/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
@@ -83,6 +83,7 @@ public class MapToolLineParser {
               PlayerFunctions.getInstance(),
               RemoveAllFromInitiativeFunction.getInstance(),
               ReturnFunction.getInstance(),
+              SoundFunctions.getInstance(),
               StateImageFunction.getInstance(),
               StringFunctions.getInstance(),
               StrListFunctions.getInstance(),
--- a/src/main/java/net/rptools/maptool/client/functions/SoundFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/SoundFunctions.java
@@ -66,23 +66,13 @@ public class SoundFunctions extends AbstractFunction {
       String strUrl = psize > 0 ? args.get(0).toString() : "*";
       int cycleCount = psize > 1 ? FunctionUtil.paramAsInteger(functionName, args, 1, true) : 1;
       double volume = psize > 2 ? FunctionUtil.paramAsDouble(functionName, args, 2, true) : 1;
-      if (strUrl.equals("*")) {
-        for (HashMap.Entry mapElement : mapStreams.entrySet())
-          editStream((String) mapElement.getKey(), cycleCount, volume);
-      } else {
-        editStream(strUrl, cycleCount, volume);
-      }
+      editStream(strUrl, cycleCount, volume);
       return "";
     } else if (functionName.equalsIgnoreCase("stopStream")) {
       FunctionUtil.checkNumberParam(functionName, args, 0, 2);
       String strUrl = psize > 0 ? args.get(0).toString() : "*";
       boolean del = psize > 1 ? FunctionUtil.paramAsBoolean(functionName, args, 1, true) : true;
-      if (strUrl.equals("*")) {
-        for (HashMap.Entry mapElement : mapStreams.entrySet())
-          stopStream((String) mapElement.getKey(), del);
-      } else {
-        stopStream(strUrl, del);
-      }
+      stopStream(strUrl, del);
       return "";
     }
     return null;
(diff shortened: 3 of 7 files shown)


Azhrei:
Three questions:

It looks like the `mapStreams` hash is being modified on two separate threads, but I didn’t see any locking.  Does it need some?

If a stream is playing and the Preferences checkbox is turned off, existing sounds are not stopped.  (I think they should be.)  Is that correct?

Does `editStream` start a stream if it’s not currently playing?  If so, it needs to check the Preferences setting.

(Sorry, I was trying to look through the code on my phone!)

I’d like to see an overall volume control for stuff coming from MT, although I realize it never had one before. 😐  Is that why you asked about a slider control in Abeille in the Discord channel?

Azhrei:
Ah, great.  I see you also refactored some other pieces.

Um, you’re going to hate me for this, but...  if `mapStream` is only accessed by the JavaFX thread now, shouldn’t it — and the code that accesses it — be placed into a separate class with the variable marked private?  It just seems like proper encapsulation to do that.

Otherwise, this looks good!  Thanks!

Merudo:
Thanks for the comments, @Azhrei !

Yeah I changed some things.

I've fine with creating a new class, in fact I almost did.

Regarding the volume slider, where do you think it should go? It could be in the Preferences, but @Jamz suggested it could also go in the toolbar.

rkathey:
I like either the toolbar or the bottom bar with asset cache, garbage collection, and other odds and ends reside.

Phergus:
A volume slider on the toolbar sounds like a good idea.

JamzTheMan:
FWIW the stop button should either stick until you hit play again. If the framework sends the macro link to autosteam audio to me, I should be able to press stop to stop current and future streams.

Maybe three way toggle? Play -> Pause -> Stop where pause pauses current stream but if another stream starts it resumes, as where stop stays stop until you press play again.

I assume these would be universal and used to control future audio as well?

It probably should stick per preferences too? (Some users may never want audio)

Lastly, if in toolbar, should it collapse or stay always available like the PC/NPC view buttons?

rkathey:
> Maybe three way toggle? Play -> Pause -> Stop where pause pauses current stream but if another stream starts it resumes, as where stop stays stop until you press play again.

I think the pause would be useful in face-to-face games. Not sure what the pause button would be used for in remote gaming sessions. Seems like you'd want it either on or off. Maybe the Pause is only on the GM's screen?

Phergus:
So in a remote session maybe the player has the MT sounds going through speakers when he gets a phone call.  Which is better, clicking on a pause/stop button or dragging a volume slider down to 0?

rkathey:
I think for a player, the stop button is the mute button and should potentially be labeled that way. In fact, they may only need a volume control button and mute.

Phergus:
So a stop/mute button plus volume on the toolbar for both players and GM.  And the preferences checkbox to disable entirely.

That sound right?

rkathey:
I was thinking differently about it. The play/pause would bt on the GM to initiate and pause the stream on everyone's machine. Stop/Mute and Volume would be on everyone's screen to control volume, mostly. In this case, the only preferences setting I can think of is to not play any music which would keep the other controls greyed out.

JamzTheMan:
OK, lets not flood this issue but...

Right now, it's only a "function". However, said function could be auto-ran via macro code to other clients. And since clients don't have a way to "don't run this macro" or "stop this macro" then they need a way to stop this particular macro (or rather the effect of the macro). 

I don't think (for now) the GM should be able to control the play/pause/stop buttons on client machines. Per previous concerns, some people may not want any sound (for audio reasons or limited internet reasons), and if you have a mix of players, a global stop/start/pause button control by the GM should probably be held off for a later release.

JamzTheMan:
Oh, and I guess, if that means we only do a master volume control for now, so be it. But there should still be a way to prevent the stream 100% for bandwidth reasons, which is mainly why i proposed a stop button (rather than a preference option)

Azhrei:
> since clients don't have a way to "don't run this macro" or "stop this macro" then they need a way to stop this particular macro (or rather the effect of the macro).

And the Preferences checkbox does that for people that always want it off.  I think of the Preferences checkbox as a "default" setting:  should it play by default, or be off by default/

> I don't think (for now) the GM should be able to control the play/pause/stop buttons on client machines.

Agreed.  Certainly not yet.  I think there are a lot of other things to work out and this will be a good test of the UI — I'm sure we'll get some good feedback on this.

So create that new class you were (almost) going to do and then I think it's ready for a merge.  This should be cool!

Azhrei:
Oh, @Phergus already merged it?!  I guess the new class can be a new PR, then.

Merudo:
@Azhrei do you have a recommendation for the name of the new class?

Phergus:
@Merudo  Were you still intending for `stopStream("*")` to stop all streams?  Not supplying any parameters works to stop everything but the description in the original issue still has "*" for stopping all streams.

Merudo:
Yeah if you don't put any parameter it stops all streams. You can also put "*", both work.

Phergus:
`playStream()` and `editStream()` will both take a cycleCount of **0**  and will play (or continue playing) the requested stream.  I would expect `playStream()` to not play at all but perhaps go ahead and download the file for later playback?    Would expect `editStream()` to stop playback.

Azhrei:
> @Azhrei do you have a recommendation for the name of the new class?

No, not really.  Maybe MediaPlayerAdapter or something?  It sort of fits the Adapter pattern...

Merudo:
> playStream() and editStream() will both take a cycleCount of 0  and will play (or continue playing) the requested stream. 

That behavior is [straight from JavaFX](https://docs.oracle.com/javafx/2/api/javafx/scene/media/MediaPlayer.html#cycleCountProperty) - only arguments greater or equal than 1 are recognized.

We could either have 0 returns an error, or we could have it stop the stream.

> I would expect playStream() to not play at all but perhaps go ahead and download the file for later playback?

I believe "streaming" media typically doesn't involve downloading the file.

> Would expect editStream() to stop playback.

Not sure about that. One use of editStream() is to adjust the volume, which I don't believe should stop the stream..

Phergus:
> That behavior is straight from JavaFX - only arguments greater or equal than 1 are recognized.

And yet playStream() is playing the stream when given a cycleCount of 0 which is counter-intuitive.  Obviously we could just document that 0 and 1 both mean 1.

> > I would expect playStream() to not play at all but perhaps go ahead and download the file for later playback?

> I believe "streaming" media typically doesn't involve downloading the file.

But it's not really streaming is it?   You pass in a http based URI and you can watch the network activity to see that it immediately downloads the whole thing.   Your comment regarding remove parameter suggests that stopping leaves the media stream in memory.  I was hoping that passing a cycle count of 0 would essentially just get it buffered up but not play it.

Merudo:
> But it's not really streaming is it?   You pass in a http based URI and you can watch the network activity to see that it immediately downloads the whole thing.   Your comment regarding remove parameter suggests that stopping leaves the media stream in memory.  I was hoping that passing a cycle count of 0 would essentially just get it buffered up but not play it.

It does stream in the sense that it starts playing before the whole file is downloaded. You can see that plainly if you start to stream a large file and immediately disconnect from the internet - you'll only get a few seconds of sounds before it stops.

About having cycle count 0 pre-buffer the file - I love that idea!

PR #672, opened 2019-09-10
Json, store "null" "false" and "true" as their revelant java type internally

### LINKED ISSUE #670: Unable to send json values true, false and null
MT 1.5.4
There's is no way to send the following json object:
```
{
  "key0": false,
  "key1": true,
  "key2": null
}
```

to the external world using Maptool REST apis.
Best scenario, you can send strings containings "true", "false", "null" but that may clash with the structure the api is expecting.

Proposed solution: json.set and json.append should jsonify the strings "false", "true", and "null" into the appropriate Java type so the JSON library properly construct the json object.

That way we gain the ability to send any JSON supported data type.
We loose the ability to properly encode the **actual strings** "false", "true", "null", however I humbly think it's still better that the inability to send any bool or null value.

bobifle:
Fix #670 proposal

In the same way json getters stringify incoming json data type for values, the setters now jsonify the maptool valule into the proper data type so the json library properly generates the Json string.

I have tested this against a RESTfull app that just resend to received json as a raw json string.

The following MT code
<pre>
[h: baseURL = "http://whatever:5123/api"]
[h: jarray = json.append("[42]", "true")]
[h: jarray = json.append(jarray, "plouf")]
[h: jarray = json.append(jarray, "false", "null", "_null_")]
[h: response = REST.post(baseURL + "/test", jarray,"application/json; charset=utf-8", 0)]
A: [response]
[h: jobj = json.set("{}", "key1", "true")]
[h: jobj = json.set(jobj, "key2", "plouf")]
[h: jobj = json.set(jobj, "key3", "true")]
[h: jobj = json.set(jobj, "null", "null")]
[h: response = REST.post(baseURL + "/test", jobj,"application/json; charset=utf-8", 0)]
B: [response]
C: [json.get(response, "null")]
D: [json.get(response, "key3")]
</pre>

generate the following output.

<pre>
A: [42,true,"plouf",false,null,"_null_"] 
B: {"key3":true,"key2":"plouf","key1":true,"null":null}
C: null 
D: true
</pre>

As you can see in outputs A and B, true false and null can now be encoded without string quotes.
They can still be accessed and MT getters convert them to strings (outputs C and D)

I have not tested for any regression using all other json functions. I'm super new to this, is there any test suites that run regression tests on the json functions ?

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/672)


JamzTheMan:
Do we have the opposite problem now and I can't sent a literal String of "true"? (What if the other end of the system can't can't booleans?"

bobifle:
Yes we have the opposite problem, but its magnitude is far lower.

- you gain the ability to send the very common true, false and null json values
- you loose the ability to send the very uncommon strings "false", "true" and "null" (because ppl use the actual json datatype)
- The other system must support json that's a hard requirement, hence it has to support true, false and null. How it does it is probably not Maptool's problem (for instance, Maptool converts everything into strings, that's one way of handling it)

Phergus:
This is against the wrong branch.  It should be against `develop`.

Phergus:
To answer your question above, no, there are no regression suites for the JSON functions.

Any resolution to this can't break existing macro code that handles JSON.  Such as:
```
[h: char = '{"name":"Bob","isAlive":"true"}']
Bob [r: if(json.get(char,"isAlive") == "true", "is Alive", "is Dead")]
```

bobifle:
I'm confident that any getter setter code in macros will work. I've tested some.

Where I'm less confident is for the builtin json macro functions that work at the java level. Because I changed the java data type of these values.
For instance the function JSONCount may be broken, I'm not sure, as it may now encounter boolean values.

PR #677, opened 2019-09-12
Add propertyType option for getTokens(), improve getTokens()

### LINKED ISSUE #676: Add "PropertyType" option for getTokens

It would be useful to augment `getTokens` to return only tokens of a given property type as defined in Campaign Properties (ex:"Basic").

For example if I get set a property type of "Trap", I could have `getTokens` only return trap tokens.

Not only would it be helpful for getting the right list of tokens, but it could improve the speed of `getTokens` when it is possible to narrow down the property type of the desired tokens.

An additional field to the `condition` parameter of `getTokens` called `propertyType`, which can be a JSON array of property types, or a single type as a string

Merudo:
- Add new option "propertyType" to conditions in getTokens()
- Value can be a single type as string, or an array of types
- Simplify getTokens() code by removing the second loop. Now include/exclude is a filter property ("match")
- Improve performance of getTokens() by trimming the list continuously instead of constantly filtering the list of all tokens
- Close #676

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/677)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/FindTokenFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/FindTokenFunctions.java
@@ -15,11 +15,7 @@
 package net.rptools.maptool.client.functions;
 
 import java.math.BigDecimal;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 import net.rptools.maptool.client.MapTool;
 import net.rptools.maptool.client.MapToolVariableResolver;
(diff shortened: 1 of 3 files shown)


Merudo:
I rebased and force pushed the project - now Travis is fine with it :)

The code is exactly the same as what was approved.

Phergus:
Something is stuck now.  Isn't recognizing approval.

Merudo:
Should I open another PR?

JamzTheMan:
I saw a error popuo in reviewable, something about not able to connect to GitHub API 

I will merge this after CI is done. Admin privileges FTW.

PR #689, opened 2019-09-15
Imagepanel improvements

pnichols04:
Improvements to ImagePanel
181f1747
* Previously, `ImagePanel` requested every image in its `model` from
`ImageManager`. Now it requests only those images that are actually
inside the viewport. This improves rendering time in proportion with
the number of items outside the viewport; improvements are substantial
with more than a few dozen off-screen items, such as for large asset
folders.

* Previously, a call to `setGridSize(n)`, where n was the existing value
of `getGridSize()`, would raise a `PaintEvent`, even though the visible
surface of the `Component` had not changed. This has been modified
so that a `PaintEvent` will only be raised if the new property is
unequal to the existing value. Similar changes are made to
`SetSelecionMode(...)`, `SetShowCaptions(...)`, `setShowBorders(...)`
and `setModel(...)`.

* Spacing was increased, particularly around captions. Previously,
captions ran very close to each other and to any image below them,
creating a cluttered appearance.

* Images are now rendered with bilinear interpolation when they are
being scaled down and bicubic interpolation when they are being scaled
up, to eliminate the artifacts and muddiness associated with the nearest-neighbor
interpolation used previously. No interpolation is applied if the image
is rendered at its natural size.

* Text is now rendered with antialiasing, using the label font specified
by the theme.

* Mouse event bounds now include the caption, so that captions will
trigger tooltips and click handlers, consistent with similar controls
(such as file browsers) on all three target platforms.

* Truncated item names are now determined heuristically. Previously,
truncation used an iterative brute-force approach that allocated a
number of strings on the heap and then discarded them.

KNOWN ISSUES.

* Tooltips are not triggered until the `Component`, or
sometimes the application, loses and regains focus. This is a
pre-existing issue that was not reprodicible with sufficient reliability
to diagnose and correct.

* The `ImagePanel` used by `LookupTablePanel` does not refresh its
scrollbar visibility when the model changes (i.e., when a campaign is
loaded). This is a side-effect of the way that `LookupTablePanel`
creates its `JScrollPanel`, which does not preserve the `Component`
hierarchy from within the scope of the `ImagePanel`, and as such, leaves
no parent to be notified when the `ImagePanel` revalidates. This is a
pre-existing issue.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/689)


Azhrei:
Wow, this modifiers 45 source files!  Um, silly question, but can we break something like this up into separate issues/PRs next time!!  (I understand some things will require updates to a lot of files, such as changing params to a method in an interface, but this seems to be multiple unrelated issues.)

Thanks for the updates, though.  Looking forward to trying this out.

pnichols04:
> Um, silly question, but can we break something like this up into separate issues/PRs next time!! 

I think next time I'll just try to PR into the right branch, the first time, instead.

See #691, #694, #695

Phergus:
And PRs should always reference an open issue for tracking purposes.

PR #691, opened 2019-09-15
ImagePanel QOL improvements

pnichols04:
***Behavioral Improvements***
- **Obtain only images in view from `ImageManager`.**  Previously, `ImagePanel` obtained all of the images in its model from `ImageManager`.  By obtaining only those that are (at least partly) in view, rendering speed is improved.  When more than a few dozen items are out of view, the speed-up is by an order of magnitude.

- **Trigger repaint() from paint-affecting properties only if property actually changes.**  Previously, the mutators `setGridSize(int)`, `setShowCaptions(boolean)`, `setShowImageBorders(boolean)`, `setSelectionMode(SelectionMode)` and `setModel(ImagePanelModel)` would call for a repaint even if the passed-in value was the same as the existing value (*e.g.*, `setGridSize(getGridSize())`).  These methods are modified so that they compare the passed-in value to the existing value and react only to value changes.

- **Calculate truncated caption heuristically.**  Previously, truncated text in item captions was determined using a brute-force approach that allocated a number of strings equal to the number of characters to be removed.  This PR would determine the text heuristically, by determining the average character width of the string, calculating how many of those average-character-widths will fit, and truncating to that.  It is conceivable that certain combinations of characters, such as a number of W's followed by a number of I's, could produce a string that is still longer than desired, but this is unlikely in practice and the short width of the '.' character mitigates it.

- **Include caption in mouse area.**  Captions are now included in the mouse event bounds, so clicking on a caption will select its associated item and hovering over a caption will trigger the item's tooltip.

***Visual Improvments***

image

- **Improve spacing.**  Previously, captions often ran very closely into one another, even with truncation.  Also, in a panel with multiple rows of items, captions rendered nearer to the item below than they did to the item above, for which they actually are the caption.  Spacing has been improved to avoid these issues.  

- **Render scaled images with interpolation.** Images displayed at less than their native size are rendered with bilinear interpolation.  Images displayed at greater than their native size are rendered with bicubic interpolation.  This produces less muddy images.

- **Render caption text with antialiasing.**

- **Use theme label font for captions.**

***Known Issues (all pre-existing).***

- **Tooltip triggering.**  Sometimes tooltips will not trigger unless the application or the component loses and regains focus.  This could not be reproduced with sufficient regularity to correct it.

- **Scrollbar after loading campaign.**  If the user loads a campaign with the lookup tables window open, and the new campaign renders with a different number of rows in its lookup table window than the previous campaign, the scrollbar does not update (*i.e.*, it does not scale its scroll height, or become visible or invisible, as the case may be).  This appears to be a by-product of the `replaceComponent(...)` call in `LookupTablePanel.java`, which creates the `JScrollPanel`.  That call removes the `ImagePanel`'s reference to its parent, and thus to its hierarchy, so the `ImagePanel`'s `revalidate` messages are not received by anyone, and it has no reference to the `JScrollPanel` to trigger it directly to revalidate.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/691)

### DIFF
--- a/src/main/java/net/rptools/lib/swing/ImagePanel.java
+++ b/src/main/java/net/rptools/lib/swing/ImagePanel.java
@@ -14,38 +14,15 @@
  */
 package net.rptools.lib.swing;
 
-import java.awt.Color;
-import java.awt.Cursor;
-import java.awt.Dimension;
-import java.awt.FontMetrics;
-import java.awt.Graphics;
-import java.awt.Graphics2D;
-import java.awt.Image;
-import java.awt.Paint;
-import java.awt.Rectangle;
+import java.awt.*;
 import java.awt.datatransfer.Transferable;
-import java.awt.dnd.DnDConstants;
-import java.awt.dnd.DragGestureEvent;
-import java.awt.dnd.DragGestureListener;
-import java.awt.dnd.DragSource;
-import java.awt.dnd.DragSourceDragEvent;
-import java.awt.dnd.DragSourceDropEvent;
-import java.awt.dnd.DragSourceEvent;
-import java.awt.dnd.DragSourceListener;
-import java.awt.dnd.DragSourceMotionListener;
+import java.awt.dnd.*;
 import java.awt.event.MouseEvent;
 import java.awt.event.MouseListener;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
+import java.util.*;
 import java.util.List;
-import java.util.Map;
 import java.util.Map.Entry;
-import javax.swing.JComponent;
-import javax.swing.JScrollPane;
-import javax.swing.Scrollable;
-import javax.swing.SwingUtilities;
-import javax.swing.ToolTipManager;
+import javax.swing.*;
 import net.rptools.lib.CodeTimer;
 
 @SuppressWarnings("serial")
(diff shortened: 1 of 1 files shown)


JamzTheMan:
@pnichols04 This change breaks the ability to zoom the images. Using ctrl+mousewheel or the slider sets the images to smallest setting and doesn't zoom after that.

image

If you could create Issues for PR's in the future, it would be handy for tracking. Now we have to create a Bug issue to track this to make sure it's fixed or rolled back before 1.5.5.

JamzTheMan:
Nice job, this this is flying now! Really notice a difference in the large dirs and dirs with large files like maps.

Azhrei:
Does it still load the next few rows of images in a background thread or something?  Or maybe using invokeLater() to do a given number of them, and when that finishes, doing another set group?  I'm just wondering how the scroll speed is when the images aren't already in RAM so the ImageManager has to load them (IIRC, the IM talks to the AssetManager to load the image the first time, then caches a reference to it, but I could be misremembering).

Cool stuff!

pnichols04:
> Does it still load the next few rows of images in a background thread or something? Or maybe using invokeLater() to do a given number of them, and when that finishes, doing another set group? I'm just wondering how the scroll speed is...

I have an asset folder with ~1300 images, and I haven't seen an issue with scroll speed.  The worst case here *should* be no worse than the production code, where everything was pulled from the ImageManager during a scroll.  If people see a regression in the speed, though, I'm happy to look into it.

PR #723, opened 2019-09-19
Add scrolling to Select Map menu

### LINKED ISSUE #356: Select Map doesn't handle large numbers of maps well
I think this has been an issue from the begginning, just never bumped into it myself until I ran this sandbox campaign.

Currently, while you can make as many maps as your system can handle. You can have more maps than your screen can 'stack' vertically, but maptool does nothing to help navigate to those other maps. 

Maptool needs to either prevent you from making more maps than its menu can navigate, or better yet, allow you to scroll through the maps down the side. A possbly more complicated feature like a search could possibly also be great.

I know one can just export/import maps to manage large numbers of maps, and its probably the better way. But feels like a bug for the program to have a list that gets too big to properly display.

Merudo:
- Add new class "JScrollPopupMenu" to handle scrolling
- Class is fully functional but not complete.  Other methods relying on indexes of the components should be reimplemented before they are used
- Maximum of maps displayed at a time: 20 (scrolling will reveal them all)
- Close #356

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/723)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/JScrollPopupMenu.java
+++ b/src/main/java/net/rptools/maptool/client/ui/JScrollPopupMenu.java
@@ -1,0 +1,240 @@
+/*
+ * This software Copyright by the RPTools.net development team, and
+ * licensed under the Affero GPL Version 3 or, at your option, any later
+ * version.
+ *
+ * MapTool Source Code is distributed in the hope that it will be
+ * useful, but WITHOUT ANY WARRANTY; without even the implied warranty
+ * of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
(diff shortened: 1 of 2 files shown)


Phergus:
I suspect folks will want to make that maximumVisibleRows a user preference.

JamzTheMan:
Does it hide the scrollbars unless needed?

I wouldn't think a set number would be the way to go. Shouldn't it be based on the size of the component which should maybe be scaled in relation to the window height?

Merudo:
> Does it hide the scrollbars unless needed?
> 
> I wouldn't think a set number would be the way to go. Shouldn't it be based on the size of the component which should maybe be scaled in relation to the window height?

Agreed. But I don't feel confident having it depends on the window height, especially because the window can be resized.

PR #726, opened 2019-09-19
Add function json.toVars to turn jsonObject into variables

### LINKED ISSUE #591: Add json.toVars() function to turn json objects into vars like varsFromStrProp()
There is no built-in function to turn a json object into variables. Users have to make their own varsFromJson() function. 

The function varsFromStrProp() could take a "delim" parameter, that if set to json turns a json into variables.

Adding another function varsFromJson() or **json.toVars()**. Either of these would be fine really.

Merudo:
- keys become variable names
- values become the content of the variables
- Close #591

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/726)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/JSONMacroFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/JSONMacroFunctions.java
@@ -181,22 +181,27 @@ public class JSONMacroFunctions extends AbstractFunction {
     }
 
     if (functionName.equalsIgnoreCase("json.toVars")) {
-      FunctionUtil.checkNumberParam(functionName, parameters, 1, 2);
+      FunctionUtil.checkNumberParam(functionName, parameters, 1, 3);
       JSONObject jsonObject = FunctionUtil.paramAsJsonObject(functionName, parameters, 0);
-      boolean suffix =
-          parameters.size() > 1 && "SUFFIXED".equalsIgnoreCase(parameters.get(1).toString());
+      String prefix = parameters.size() > 1 ? parameters.get(1).toString() : "";
+      String suffix = parameters.size() > 2 ? parameters.get(2).toString() : "";
 
-      int count = 0;
+      JSONArray jsonNames = new JSONArray();
       for (Object keyStr : jsonObject.keySet()) {
-        String varName = keyStr.toString() + (suffix ? "_" : "");
-        Object value = jsonObject.get(keyStr);
-        // don't allow spaces in var name
-        if (!varName.contains(" ")) {
+        // add prefix and suffix
+        String varName = prefix + keyStr.toString().trim() + suffix;
+        // replace spaces by underscores
+        varName = varName.replaceAll("\\s", "_");
+        // delete special characters other than "." & "_" in var name
+        varName = varName.replaceAll("[^a-zA-Z0-9._]", "");
+
+        if (!varName.equals("")) {
+          Object value = jsonObject.get(keyStr);
           parser.setVariable(varName, value);
-          count += 1;
+          jsonNames.add(varName);
         }
       }
-      return count;
+      return jsonNames;
     }
 
     if (functionName.equals("json.set")) {
--- a/src/main/java/net/rptools/maptool/client/functions/JSONMacroFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/JSONMacroFunctions.java
@@ -56,6 +56,7 @@ public class JSONMacroFunctions extends AbstractFunction {
         "json.fromStrProp",
         "json.toStrProp",
         "json.toList",
+        "json.toVars",
         "json.append",
         "json.remove",
         "json.indent",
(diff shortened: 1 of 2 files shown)


Azhrei:
What does this do with multilevel objects?

JamzTheMan:
``` Object value = jsonObject.get(keyStr); ```
I would assume it would assign the resulting array or object to the variable? Unless the parser chokes on Object?

@Merudo was this test with keys that held JSON object or arrays?

**Comment by Merudo** [comment]
Example:

```
[h:monster1 = json.set("{}", "HP", 75, "Attacks", json.append("Claw", "Bite"))]
[h:monster2 = json.set("{}", "HP", 13, "Attacks", json.append("Sword", "Punch"))]
[h:monsters = json.set("{}", "Troll", monster1, "Orc", monster2)]

[h: json.toVars(monsters)]
[r: Troll]<br>
[r: Orc]
```

which returns

```
{"HP":75,"Attacks":["Claw","Bite"]}
{"HP":13,"Attacks":["Sword","Punch"]}
```

**Comment by Azhrei** [comment]
Cool.  That works for me. :)

**Comment by Phergus** [comment]
In the code you are skipping keys that have spaces in the name.  Maybe replace spaces in the name with underscores?  I'm iffy on it.

Also instead of returning a count of vars created how about returning an array of the keys made into vars?  Can still get a count and I can see it making it easier to do debugging.  I guess json.fields already does that but if you had both you could do a json.equals to compare them.  Thoughts?

**Comment by Azhrei** [comment]
If the car already exists, I assume it’s value is overwritten?  From a debugging standpoint, should there be an option to flag that as an error?  I know varsFromStr doesn’t do that but perhaps it should as well?  I don’t do a lot of MT scripting, so better if the geniuses at this stuff chime in...

**Comment by Merudo** [comment]
The behavior of ignoring fields with spaces, and returning the number of variables set, is straight from `varsFromStrProp`.

I think there is some value to keep `varsFromStrProp` and `json.toVars` similar, so maybe we could change `varsFromStrProp` to match it?

**Comment by Phergus** [comment]
I'm not really set on one way or the other.  If the folks doing the major scripting are happy with how varsFromStrProp works, then we can follow that model for this.

@aliasmask  Thoughts?

**Comment by aliasmask** [comment]
In my own version of json.toVars it removed spaces and other non var name characters, but still defines it locally. Also, json.toVars returns a json object of var names defined as the key and the actual name as the value. It also makes all the var names lowercase. I also have a parameter for a prefix.

The reason I did that is so I can use the original name in output while still having the varname with data defined locally.

```
<!-- json.toVars(obj,prefix): varList
   obj - variable is a json object
   varList - A json object of prop names to original key word.
   
   This function defines the keys of a json object as local variables with the format prefix ("var") + "." + lowercase keyword with no spaces.
 -->

[H: sub.obj = arg(0)]
[H, if(argCount() >= 2): sub.prefix = arg(1); sub.prefix = "var"]
[H, if(json.isEmpty(sub.prefix)): sub.prefix = "var"]
[H: sub.varList = "{}"]

[H, if(json.type(sub.obj) == "OBJECT"), code: {
   [H, foreach(varName,sub.obj), code: {
      [H: sub.varName = keyToVar(varName,sub.prefix)]
      [H: sub.varList = json.set(sub.varList,sub.varName,varName)]
      [H: set(sub.varName,json.get(sub.obj,varName))]
   }]
}]

[H: macro.return = sub.varList]```

In my example below, I only convert spaces, because that's all the current project needed, but I would update to include any non alpha-numeric character, underscore and period.
```


[H, if(argCount() >= 2): macro.return = arg(1) + "." + replace(lower(arg(0))," ",""); macro.return = replace(lower(arg(0))," ","")]
```

The one major thing that needs to be done is if you have json.toVars we have to have json.fromVars. I do this mostly with large form inputs where I can repackage the data changed and save it.

```

```
varList can be a list, array or object. If an object, then matching key values are saved as the new key in the new object.

Given all the above, I would be happy if it were exactly like the varsFromStrProp and I can do the rest on my own. I'm just demonstrating how it would be most useful to me.

Merudo:
Cool! Thanks for sharing this, @aliasmask.

Regarding variable names, what's your rationale for putting them all in lower case?

Merudo:
For handle the variables name, maybe something like this would be appropriate:
```java
varName = varName.replaceAll("[^a-zA-Z0-9._]", "");```

aliasmask:
Having it always lowercase is just to have one less thing to remember about the variable. Was that first letter capitalized or not? Is the first letter of additional words capitalized? Since I'm saving the original format of the key changing everything to lower case makes everything nice and uniform. I just do it for continuity in the naming convention. ie feat.twoweaponfighting

rkathey:
MTScript isn't case sensitive so you can mix and match how you like. I'm not sure if MTScript2 will be case sensitive or not.

aliasmask:
That's true for variable names, but not things like UDFs or macro functions. So, in the end it doesn't matter either way if its converted to lowercase other than for aesthetics.

Merudo:
> If an object, then matching key values are saved as the new key in the new object.

I'm not sure if I understand this sentence. What's the relationship between the object and the variables in this scenario?

Merudo:
New commit:

- Change: parameter "suffixed" is now replaced by "prefix" and "suffix", which can add a prefix and/or suffix to the names.
- Change: spaces in keys are now turned into underscore
- Change: characters not alpha-numerical other than "." and "_" are deleted from var names
- Change: function now return a JSONArray with the variabnle names

aliasmask:
What I mean is...

[H: myObj = '{"Item 1":"value 1","Item 2":"value 2"}']
[H: varList = json.toVars(myObj,"var")]





[H: var.item1 = "new value"]


[H: myObj = json.fromVars(varList)]

Phergus:
Tested and seems to be working as described.  Still needs documentation.

Issue #663, opened 2019-09-06
Add "isHostingServer" and "gmRevealsVision " to getInfo("server") results


Phergus:
**Is your feature request related to a problem? Please describe.**
Add the Hosting a Server status to the getInfo("server") results.
Same for the new server setting, GM Reveals Vision... (see #665)

**Describe the solution you'd like**
Add "hosting server" and "GM reveals vision for unowned tokens" to the JSON result from getInfo("Server")

JamzTheMan:
thanks :)

Phergus:
@JamzTheMan  Also needs getInfoFunction.java updated so that those settings show up in the returned JSON.

JamzTheMan:
Ah ok. I incorrectly assumed it was using the raw json that was returned here but couldn't test it because the parser died in dev. :(

Phergus:
I have no problem running from develop with gradlew run, using the uberJar or just launching from the project in Netbeans.  Macros work just fine for me.

JamzTheMan:
I still get this:

```java.lang.NoClassDefFoundError: net/rptools/parser/ExpressionLexer
	at net.rptools.parser.Parser.parseExpression(Parser.java:251)
	at net.rptools.common.expression.ExpressionParser.evaluate(ExpressionParser.java:214)
	at net.rptools.maptool.client.MapToolLineParser.parseExpression(MapToolLineParser.java:1490)
	at net.rptools.maptool.client.MapToolLineParser.parseLine(MapToolLineParser.java:1298)
	at net.rptools.maptool.client.MapToolLineParser.parseLine(MapToolLineParser.java:711)
	at net.rptools.maptool.client.macro.MacroManager.executeMacro(MacroManager.java:177)
	at net.rptools.maptool.client.ui.commandpanel.CommandPanel.commitCommand(CommandPanel.java:511)
	at net.rptools.maptool.client.ui.commandpanel.CommandPanel.commitCommand(CommandPanel.java:448)
	at net.rptools.maptool.client.AppActions$39.execute(AppActions.java:1468)
	at net.rptools.maptool.client.AppActions$ClientAction.actionPerformed(AppActions.java:3044)
	at java.desktop/javax.swing.SwingUtilities.notifyAction(SwingUtilities.java:1810)
	at java.desktop/javax.swing.JComponent.processKeyBinding(JComponent.java:2900)
	at java.desktop/javax.swing.JComponent.processKeyBindings(JComponent.java:2948)
[...]

PR #738, opened 2019-09-23
Changed getInfo("server") macro to use ServerPolicy.toJSON directly

Phergus:
Updated getInfo("server") macro function to pull JSON object from ServerPolicy instead of having duplicated code.
Updated ServerPolicy.toJSON() to include isAutoRevealMovement and isUseIndividualFOW.

Finish #663

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/738)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/getInfoFunction.java
+++ b/src/main/java/net/rptools/maptool/client/functions/getInfoFunction.java
@@ -26,9 +26,7 @@ import net.rptools.maptool.client.AppPreferences;
 import net.rptools.maptool.client.MapTool;
 import net.rptools.maptool.client.ui.token.BarTokenOverlay;
 import net.rptools.maptool.client.ui.token.BooleanTokenOverlay;
-import net.rptools.maptool.client.ui.tokenpanel.InitiativePanel;
 import net.rptools.maptool.client.ui.zone.ZoneRenderer;
-import net.rptools.maptool.client.walker.WalkerMetric;
 import net.rptools.maptool.language.I18N;
 import net.rptools.maptool.model.Campaign;
 import net.rptools.maptool.model.CampaignProperties;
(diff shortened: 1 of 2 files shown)


JamzTheMan:
___
*[src/main/java/net/rptools/maptool/client/functions/getInfoFunction.java, line 244 at r1](https://reviewable.io/reviews/rptools/maptool/738#-LpQ_6xy4JwbY2P6cbb9:-LpQ_6xy4JwbY2P6cbbA:b-gn4rf8) ([raw file](https://github.com/rptools/maptool/blob/b6714c11b716c12578699b47ac78cac288c394d8/src/main/java/net/rptools/maptool/client/functions/getInfoFunction.java#L244)):*
> ```Java
>       sinfo.put(
>           "initiative owner permissions",
>           ip.isOwnerPermissions() ? BigDecimal.ONE : BigDecimal.ZERO);
> ```

Are we going to leave this out now? Does it need it's own getInfo()?  I don't think it belongs in ServerInfo but it "may" break some existing macros?

Phergus:
That taking it out of the server info may break some existing macros is certainly a concern.   Seems kind of silly though to put it in its own info block by itself.   It appears to be the kind of setting that is set once and stays that way much like the other server settings.  Are there other initiative related settings that would justify having a new getInfo("initiative")?

JamzTheMan:
I can't think of any off the top but does not mean there isn't.

Semver tells us this shouldn't be a 1.5.x change. Well really, new features should bump our minor version number each time, only small changes and bug fixes should be 3rd dot increments...

ANYhoo, I think we should keep it for now and mark it deprecated in the wiki for removal in 1.6? In the meantime we can look to see where it can live, maybe there are some other random settings that would be useful to have?

If you can't easily modify that returned JSON from ServerPolicy, maybe overload the method in ServerPolicy and pass it a string and value to add?

Phergus:
FWIW, it's also in the info block for "campaign" as is "initiative movement locked".

It is trivial to remove it from the server policy JSON object as it isn't actually used anywhere, just reported as info in the Sentry context and to macro getInfo().  Nothing internal to MapTool is going to care if it goes away.

JamzTheMan:
Oh well then, if it's already in campaign info then it's a bug here and let's remove it. :)

Phergus:
Don't see the need for it in both places and since two initiative options are already in the `campaign` info, we can stick with that.

PR #741, opened 2019-09-23
Junit tests for Json function

bobifle:
far from being enough, but it's still better than nothing.

issue #670

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/741)

PR #744, opened 2019-09-26
Junit tests for Json function (#741)

JamzTheMan:
* jsonify "null", "true" and "false" into null true and false

* fix null handling, using the library Null object

* jsonify should be static and public

* json unitary tests

* use a resource fiel for json rather that constructed within the test code

* spotless on json tests

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/744)


JamzTheMan:
@Phergus is you want to review this request I can merge it. Finally got the CI's to report OK.

Issue #746, opened 2019-09-28
MapTool 1.5.5 beta builds have old parser


Phergus:
**Describe the bug**
MapTool 1.5.5 beta builds have old parser

**To Reproduce**
1. Start an MT 1.5.5 beta build
2. In chat window, type in `[r: abort(1,2)]`
3. Note output of `Invalid number of parameters 2, expected exactly 1 parameter(s)`

**Expected behavior**
Output should include the function that was called like this:
`Function 'abort' requires exactly 1 parameters; 2 were provided.`

**MapTool Info**
- Version: 1.5.5 beta build
- Install: New

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: 10

**Additional context**
Appears that new release builds are needed for Parser & DiceLib.

Phergus:
Pull request in DiceLib to get it using parser 1.5.5 opened. 
https://github.com/RPTools/dicelib/pull/8

Will need PR for MT to reference new dicelib in build.gradle file.

Phergus:
Once new dicelib release is available updating build.gradle for MapTool will complete this.

Phergus:
Fixed in 1.5.5 beta 5 build.  Closing.

PR #747, opened 2019-09-28
Update parser to 1.5.5

JamzTheMan:
Removed self dependency in Parser project.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/747)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -181,7 +181,7 @@ dependencies {
     implementation 'net.rptools.decktool:decktool:1.0.b1'
 
     implementation 'net.rptools.maptool.resource:maptool.resource:1.0.b18'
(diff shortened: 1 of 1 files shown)


JamzTheMan:
This should close #746 

Note, dicelib is still pulling in the older 1.1 parser, no idea if that project needs updating or not?

Phergus:
Once I update my local repo and do a uberJar build I'll be able to see if it is still pulling in the old parser.  I suspect DiceLib is going to need to have an update to the build file and then a new release.

Phergus:
The MapTool jar file is still pulling in both old & current parser classes.  Will need DiceLib updated.

PR #757, opened 2019-10-02
Fix prompt to "update" to old version when on pre-release

### LINKED ISSUE #627: 1.5.4 RC1 telling me 1.5.3 is more recent
**Describe the bug**
The maptool release candidate 1.5.4 RC1 tells us to "update" to 1.5.3

**To Reproduce**
1. Install Maptool 1.5.4 RC1
2. Launch and read the update message

**Expected behavior**
No message to update if you are on the most recent release candidate.

**MapTool Info**
- Version: 1.5.4 RC1
- Install: New

 - OS: Windows
 - Version 10

Merudo:
- Fix so that a release candidate / alpha / beta no longer gives prompt to "update" to an older release
- Change: users on a pre-release now get a prompt to update to the newest version, including pre-releases
- No change for those on stable releases
- Close #627

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/757)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppUpdate.java
+++ b/src/main/java/net/rptools/maptool/client/AppUpdate.java
@@ -32,12 +32,25 @@ public class AppUpdate {
   private static final Logger log = LogManager.getLogger(AppUpdate.class);
 
   static final String GIT_HUB_API_URL = "github.api.url";
+  static final String GIT_HUB_RELEASES = "github.api.releases";
   static final String GIT_HUB_OAUTH_TOKEN =
       "github.api.oauth.token"; // Grants read-only access to public information
 
+  /**
+   * Look for a newer version of MapTool. If a newer release is found and the AppPreferences tell us
+   * the update should not be ignored, give a prompt to update. If current version is a release,
+   * update to the most recent release. If the current version is a pre-release, update to the most
+   * recent version (pre-release or release).
+   *
+   * @return has an update been made
+   */
   public static boolean gitHubReleases() {
     // AppPreferences.setSkipAutoUpdate(false); // For testing only
     if (AppPreferences.getSkipAutoUpdate()) return false;
+
+    String version = MapTool.getVersion().toLowerCase();
+    boolean prerelease =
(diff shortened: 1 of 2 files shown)


JamzTheMan:
Is there a reason not to use GitHub API info? I'm pretty sure prerelease is a flag. Not sure about draft, although we need to change CI so it sticks to pre-release vs draft after it builds.

That way we don't have to hardcode looking for tag name (maybe there will be an alpha release? Who knows?)

Now that jayway path is available to us, it should be easier to use a JSON.path to get objects from the API response.

Merudo:
As far as I can tell, searching for "rc" or "alpha" in the version name seems to be the most straightforward way to tell if the current version is a release or a pre-release.

If I understand you correctly, you are suggesting that we instead loop through the list of all releases on github, find the one that matches the current one (if any), then check the `prerelease` flag. Is that correct?

JamzTheMan:
We shouldn't need to do any looping. I'd have to look at the API but we have the commitsha so should be able to query the API with that and get the object and check the flag.

There should be a direct way to get the info we need, I've done other work on getting blobs and can get any of that info using sha.

Your way does work but why not use the API?

Merudo:
That sounds good! But, I'm concerned for the pre-releases - it seems they all eventually get deleted from https://api.github.com/repos/RPTools/maptool/releases

JamzTheMan:
Well that's a good point! It's confusing (IMO) to leave em so we would need a permanent way to tell...

Could put it in the jar manifest but we would have to know this at build, may be easier when we move to GH actions but harder right now...

BUT, if we just assume if no info is found for that release (deleted) then we could assume it's a draft/pre-release/local build?

Merudo:
The second commit implements your suggestion.

It uses Jayway to find the latest commit that has a matching SHA, and look for the `prerelease` field.

If it can't find a match, the current version is assumed to be a pre-release.

JamzTheMan:
Awesome, looks great! We'll test this in the next round of beta/rc releases for 1.5.6

Issue #766, opened 2019-10-05
Overlay perlin/gausian noise on tiling textures


Syndaryl:
**Is your feature request related to a problem? Please describe.**
I have brain damage that has left me sensitive to repeating visual patterns, such as tiled textures. The smaller the texture and denser the tiling, the more "abrasive" it feels. I understand neurotypical people find this a little displeasing too, but I have a nails-on-blackboard problem with it.

**Describe the solution you'd like**
A ticky-box option to Overlay a perlin noise or gausian noise function over textures used with drawing tools, where they will be tiled to fill the drawing tool shape, and a separate ticky-box option to Overlay the same over a map background image, which is tiled to fill your entire screen omg.
This would generate a low contrast, relatively large scale noise function. It should not tile, it should be independent of the tile of the texture underneath.
Using Overlay slightly brightens part of the underlying texture and slightly darkens part of the underlying texture.

The result "breaks up" the repetition of the tiling pattern enough that it feels drastically better for me, and I am informed it is more aesthetically pleasing for neurotypical people as well. The Cartographers Guild recommend it.

I _think_ the best implementation would make it a client-side option in the accessibility settings, so that it can be turned off by people who don't like the effect or find the effect not worth the extra processing overhead. It could default to off as well.

**Describe alternatives you've considered**
The alternative puts the onus on the GM to create larger tiles and apply the noise function themselves. This results in a) more work and b) larger file sizes for the tiles, and c) if done incorrectly results in seams when the new tile is ... tiled, which nobody likes.

**Additional context**

![Tiling Grass](https://user-images.githubusercontent.com/8820565/66257023-309e4200-e762-11e9-88bd-6af25dd4a1a4.png)

![Tiling Grass with Noise](https://user-images.githubusercontent.com/8820565/66257029-385de680-e762-11e9-846e-87ab7fc659de.png)

JamzTheMan:
Sounds reasonable.

I know another thing that was considered, a lot of textures sometimes come with 4-8 "versions" like grass-a, grass-b, etc.  An option to select multiple textures when tiling and have it 'randomly" use those tiles may mix things up a bit.

TBH, I have not tried this (manually or otherwise) so not sure if the effect would be desirable or not.

Azhrei:
I love this idea!

Since the noise isn’t part of the tile, it will need to be generated after the tiles are blitted to the image.  If the noise isn’t saved somehow, every refresh of the image would generate different noise, so what happens when the map is being panned?

[...]

PR #794, opened 2019-10-14
Bg texture noise

cwisniew:
Background Texture noise off by default
Add on / off options to the /texturenoise command

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/794)


Phergus:
Applies to #766

PR #808, opened 2019-10-17
Fix incorrect tooltip for Show Chat Notification Background

### LINKED ISSUE #800: Incorrect tooltip for Show Chat Notification Background
**Describe the bug**
The tooltip for `Show Chat Notification Background` (in Preferences) is 

```
Color of the text for typing notifications
```
which is the same tooltip as for `Typing Notification Color`.

The tooltip should instead be something like `The background flashes when a message is received`.

Merudo:
- Change tooltip to "If enabled, the background flashes if a message is received when MapTool doesn't have the focus."
- Close #800

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/808)


Phergus:
That's not what that controls.  If enabled it puts a frame underneath the white text that is put over the map view in the top-left corner of the map when a client is typing in the chat window.

Screenshot 2019-10-17 14 23 59

Technically it's the Typing Notification Background.  Not sure why it is named differently.

I thought there was a chat message received notification toggle in preferences but I guess not.

Merudo:
Thank you. I did not expect this behavior at all!

I've changed the tooltip, and the name of the property to "Typing Notification Background".

Issue #814, opened 2019-10-18
Bug: exception when copyToken used with some updates


Merudo:
**Describe the bug**
Using `copyToken` with the following updates raises an exception: x, y, layer, tokenImage, portraitImage, handoutImage

**To Reproduce**
1. Run MapTool in developer mode (console)
2. Create a map with a `Dragon` token on Grasslands
3. Run the following code:
```
[h: updates = json.set("{}", "x", 1)]
[h: copyToken("Dragon", 1, "Grasslands", updates)]
```
4. Check the console. An exception is thrown.
 
**Expected behavior**
No exceptions.

**MapTool Info**
- Version: 1.5.6 Develop
- Install: New

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: 10

**Additional context**
Exception is thrown because some code in `copyToken` is trying to send a token update to the server before the token is even created on the server.

Merudo:
PR #815 solves this.

Phergus:
Good find.

Phergus:
PR #815 fixed most but not all.

Additional steps to above:
1. Copy the top-down "Dragon" _token_ from Default -> Tokens to the map.
2. Add a handout image to the new token. (It already has a portrait.)
3. Name it "Dragon 1" if it didn't default to that.

Run this code:
```
[h: updates = json.set("{}", "x", 1, "y", 2)]
[h,token("Dragon 1"): id = getTokenImage()]
[h: updates = json.set(updates, "tokenImage", id)]
[h,token("Dragon 1"): id2 = getTokenPortrait()]
[h: updates = json.set(updates, "tokenPortrait", id2)]
[h,token("Dragon 1"): id3 = getTokenHandout()]
[h: updates = json.set(updates, "tokenHandout", id3)]
<pre>[r: json.indent(updates,2)]</pre>
[h: copyToken("Dragon", 1, "Clone vat", updates)]
```
Only the token image is changed.  No change to portrait or handout.

Try this code:
```
[h: updates = json.set("{}", "x", 1, "y", 2)]
[h: updates = json.set(updates, "tokenImage", "Dragon 1")]
[h: updates = json.set(updates, "tokenPortrait", "Dragon 1")]
[h: updates = json.set(updates, "tokenHandout", "Dragon 1")]
<pre>[r: json.indent(updates,2)]</pre>
[h: copyToken("Dragon", 1, "Clone vat", updates)]
```
Macro fails with:
[...]

PR #816, opened 2019-10-18
Change copyToken options to tokenHandout, tokenPortrait

Merudo:
- Change handoutImage update field to tokenHandout
- Change portraitImage update field to tokenPortrait
- Legacy: old values still work
- Fix null pointer exception to give proper error message if the image token cannot be found
- Solve issues raised by @Phergus in #814

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/816)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/TokenCopyDeleteFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/TokenCopyDeleteFunctions.java
@@ -287,21 +287,29 @@ public class TokenCopyDeleteFunctions extends AbstractFunction {
       }
     }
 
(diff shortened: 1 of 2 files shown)


Phergus:
No need to make those `update` condition keyword changes if you want to go ahead and back them out.

Merudo:
It would make sense to change them though. Current functions are

`getTokenHandout / setTokenHandout`
`getTokenPortrait / setTokenPortrait`

so it would make a lot of sense to use `tokenHandout` and `tokenPortrait` for the update instead of handoutImage / portraitImage

Phergus:
Well my brain clearly thought so.  Since the current ones still work no real harm in leaving them.

PR #873, opened 2019-11-08
Add metadata to getMacroProps function

### LINKED ISSUE #872: Add metadata to getMacroProps function
The ability to manage and edit macros externally from MapTool could benefit by adding additional metadata to the getMacroProps(<index>, "json") macro frunction

Add a metadata object to the returned json that contains uuid, command checksum and properties checksum
```
"metadata":  { "uuid": String, "commandChecksum": String, "propsChecksum": String }
```

An external API is needed to generate a proper MD5 checksum which is slow and not ideal.

**Additional context**
commandChecksum is a MD5 checksum on the macro "command" only. With this, you can quickly tell if the macro has been edited or changed.
propsChecksum is a MD5 checksum for macro "properties" only, excluding the "command" and as such, can tell you if any of the properties have been changed.
The UUID is an easy way to identify macros even if the name has changed or if the name occurs multiple times.

JamzTheMan:
Fixes #872

Also changed json to build from LinkedHashMap instead of HashMap so json output now matches the insert order of the objects for a cleaner output that can be controlled.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/873)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/MacroFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/MacroFunctions.java
@@ -16,10 +16,11 @@ package net.rptools.maptool.client.functions;
 
 import java.math.BigDecimal;
 import java.util.ArrayList;
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import net.rptools.lib.MD5Key;
 import net.rptools.maptool.client.MapTool;
 import net.rptools.maptool.client.MapToolVariableResolver;
 import net.rptools.maptool.model.MacroButtonProperties;
@@ -30,8 +31,12 @@ import net.rptools.parser.function.AbstractFunction;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONException;
 import net.sf.json.JSONObject;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class MacroFunctions extends AbstractFunction {
+  private static final Logger log = LogManager.getLogger(MacroFunctions.class);
+
   private static final MacroFunctions instance = new MacroFunctions();
 
   private MacroFunctions() {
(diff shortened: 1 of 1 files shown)


cwisniew:
> Fixes #872
> 
> Also changed json to build from LinkedHashMap instead of HashMap so json output now matches the insert order of the objects for a cleaner output that can be controlled.
> 
> This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/873)

After the JSON update in progress there is no guarantee this will continue to be the case. GSON is less featured than the current json library and there is no way to control this (plus you shouldnt rely on ordering of fields in json objects)

JamzTheMan:
I swore I ran spotlessApply... ug...

Also, not sure why IntelliJ is saving in a different format when I imported the Java Google code specs...

JamzTheMan:
> > Fixes #872
> > Also changed json to build from LinkedHashMap instead of HashMap so json output now matches the insert order of the objects for a cleaner output that can be controlled.
> > This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/873)
> 
> After the JSON update in progress there is no guarantee this will continue to be the case. GSON is less featured than the current json library and there is no way to control this (plus you shouldnt rely on ordering of fields in json objects)

If it sticks it sticks, if not, no big deal. There's no "reliance" needed, it's simply makes for a nicer output in chat using json.ident. 😄 
It would be nice if GSON followed the Collections list order but I wouldn't worry about it in the least.

Merudo:
If GSON is less featured than our JSON, why not keep the current library?

We can use our existing functions to convert the output of `json.path` into JSON elements if need be.

By converting to GSON, we may be performing a needless downgrade. Unless there is some performance gain from using GSON?

cwisniew:
> If GSON is less featured than our JSON, why not keep the current library?
> 
> We can use our existing functions to convert the output of `json.path` into JSON elements if need be.
> 
> By converting to GSON, we may be performing a needless downgrade. Unless there is some performance gain from using GSON?

At the moment the json.path functions are converting JSON objects to strings, then parsing the strings to GSONs representation which is quite bad performance wise. Then making the change to convert that back to the JSON representation we are using isn't straight forward. So moving to GSON is the easiest way to fix the problems.

Merudo:
> At the moment the json.path functions are converting JSON objects to strings, then parsing the strings to GSONs representation which is quite bad performance wise. Then making the change to convert that back to the JSON representation we are using isn't straight forward. So moving to GSON is the easiest way to fix the problems.

I was planning on converting GSON to strings, then parsing the strings to JSON objects. Easy enough considering we already have these parsing functions.

Staying with GSON all the way through will improve the performance.

JamzTheMan:
I may be wrong but GSON has better class -> JSON serialization as well. So it may have less functionality in some places but more/better in others.

My "not totally researched but from memory" answer so take that as is.

PR #875, opened 2019-11-08
Changed REST.delete to follow post, patch, put format, fixes #874

### LINKED ISSUE #874: REST.delete function should allow request body
**Describe the bug**
REST specifications do allow for DELETE to send a request body (payload) and as such our macro function should allow for it as well. I've already encountered API's that require payloads on DELETE (like the GitHub API) and as such not able to perform DELETE's successfully with the current specifications.

**To Reproduce**
1. Try to send a REST.delete using headers and a payload

**Expected behavior**
REST.delete should follow the same macro function signature as REST.post (patch, & put)

**MapTool Info**
- Version: 1.5.0-1.5.6
- Install: Any

JamzTheMan:
It now accepts body/payload.

This changes the signature of the macro a bit so Wiki update is needed as it no long follows REST.get signature.
This fixes an issue where some API's (like GitHub API) require data to be posted in the body.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/875)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/RESTfulFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/RESTfulFunctions.java
@@ -56,6 +56,7 @@ import okhttp3.Response;
  * body of the response is returned in what ever form the call returns.
  */
 public class RESTfulFunctions extends AbstractFunction {
+
   private static final RESTfulFunctions instance = new RESTfulFunctions();
 
   private RESTfulFunctions() {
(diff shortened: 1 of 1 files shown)


cwisniew:
@JamzTheMan needs {} for the if/else

JamzTheMan:
> If we are going to adhere to the Google style guide then by 4.1.1 then the if/else/else if blocks need braces. I think the code is readable as is and generally if spotless isn't complaining then I'm good with it.
> 
> > _[Reviewable](https://reviewable.io/reviews/rptools/maptool/875#-:-LtA75G8-sBk4W9SQk1U:b-au4l53)_ status: 0 of 1 files reviewed, all discussions resolved (waiting on @cwisniew and @Phergus)

Code compliance completed. 😄

PR #877, opened 2019-11-09
Adds ISO Formatted time with UTC offset

JamzTheMan:
Current timeDate fields in Client Info does not include any timezone information which is needed for some API calls.

New field `isoTimeDate` was added which now includes UTC offset following ISO 8601 standards.
such as `2019-11-09T10:15:30.3444-06:00`

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/877)


Phergus:
Don't see an issue for this.

And what is AppVeyor whining about?

JamzTheMan:
I thought I had created it, I will as soon as I get a chance, got pulled away from the computer.

I think CI just hiccupped downloading stuff.

Phergus:
Completes #878

Issue #883, opened 2019-11-10
Keyboard shortcuts not working on macOS


mikehenry-se:
**Describe the bug**
A number of keyboard shortcuts are not working on macOS, for example:

**To Reproduce**
- Shift-Cmd-P does not switch between the Show As Player modes.
- Cmd-G does not show grid or switch grid off.
- etc.

**Expected behavior**
That all keyboard shortcuts should work as indicated in the menus.

**Screenshots**
n/a

**MapTool Info**
- Version: 1.5.6, 1.5.7-rc1
- Install: New

**Desktop (please complete the following information):**
 - OS: MacOS
 - Version 10.15.1 (but did not work with earlier versions either, i.e. 10.14

**Additional context**
n/a

Phergus:
Can you check with MT 1.5.0?

Also does it work if you use Ctrl instead of Command?

mikehenry-se:
Downloaded and ran MT 1.5.0. Same problem.
Tried it with Ctrl instead of Command. No difference.

Phergus:
Thanks for checking.  We'll try and track it down quickly.

mikehenry-se:
Menu-items tested for which keyboard shortcuts don't work on macOS:

- Fog-of-War (Cmd-W)
- FoW: Expose only at waypoints (Shift-Cmd-W)
- Show as Player (Shift-Cmd-P)
- Show Movement Distances (Cmd-D)
- Show Light Sources (Cmd-K)
- Show Grid (Cmd-G)
- Lock Player Movement (Shift-Cmd-L)

Not verified:

- Center Players on Current View (Cmd-F)
- Enforce Players to Current Map (Cmd-E)

Tested on MT 1.5.7-rc1

Phergus:
Sigh.

Thanks for the additional info.

mikehenry-se:
Thanks for your quick reply!

The most annoying of the verified ones missing are:

- Show Grid (Cmd-G)
- Show as Player (Shift-Cmd-P)

Others are nice to haves for me.

But I'll also say that the 1.5 versions of MapTool are awesome! I have been looking at the competition for a long time and think that they all pale in comparison, in particular when it comes to cross-platform support, scriptability and now Iso-maps/-tokens. You guys are doing great!

Phergus:
Yeah.  Those are two I use all the time and it would definitely be annoying not having them.  :)

Appreciate the kind words!

Phergus:
[...]

PR #886, opened 2019-11-11
Work around JDK bug https://bugs.openjdk.java.net/browse/JDK-8208712

cwisniew:
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/886)


cwisniew:
Fixes #883

Issue #762, opened 2019-10-04
Auto save feature for campaign and chat is not working.  Lost data.


adventuremagic123:
**Describe the bug**
In MapTool 1.5.4, it looks like the auto save feature isn't working?  I had all the options set as follows for last night's (10/2/2019) game:

1. Campaign autosave every "10" min.
2. Save reminder on close "v" (checked).
3. Time between chat log autosaves "10" min.

Only saves sometime shortly after I loaded the campaign were done.  About 2 hours into the game, my Windows 10 computer blue screened (first time that's ever happened because of a video driver update and I forgot to reboot).  I found out that there were no auto saves for either the chat or the campaign, except there was one for the campaign apparently from when I had first loaded the campaign.
.
So last night we played from 7:15pm CDT till 10:15pm on 10/2/2019, and with those settings, this is what my autosave directories look like the attached autosave images.

**To Reproduce**
1. See configuration attached.
2. Started server.
3. Players attached.
4. No autosaving done during about 2 hours of game play.

**Expected behavior**
Autosave files should exist at 10 minute intervals.

**Screenshots**

**MapTool Info**
- Version: 1.5.4

**Desktop (please complete the following information):**
 Windows 10
![pereferences for autosave](https://user-images.githubusercontent.com/50708405/66213648-ce700f00-e685-11e9-9632-e7987252d89a.png)
![autosave cmpgn](https://user-images.githubusercontent.com/50708405/66213674-de87ee80-e685-11e9-8f63-82f2755bd2fd.png)
![autosave chat](https://user-images.githubusercontent.com/50708405/66213680-e182df00-e685-11e9-83ad-1f088ec9591d.png)

**Additional context**
From MapTool, paste info from menu Help -> Gather Debug Information...
You can also attach files (drag and drop here) such as log files or paste a file sharing link.

JamzTheMan:
Yea, I got bit by that in 1.5.4 as well 2 weeks ago. Autosave has always been somewhat unreliable. It always seems to work when you check and fails at the worst times.

Maybe we need a watchdog process that watches the for the latest save and if it doesn't see a save for the last x (defined) minutes it changes the border to red or something to let you know autosave is behind schedule.

At least until we can be confident it works again. 

Also, maybe we change autosave to just save in directory mode and save on cpu/disk cycles for zip (a first step towards directory based saves)?

Azhrei:
This really concerns me.

[...]

PR #918, opened 2019-11-18
Fix auto save campaign saving once once before stopping

Merudo:
- Fix the auto save campaign timer not restarting after saving the campaign once.
- Partial fix to #762

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/918)


Merudo:
Closing because fix is included in #919

PR #919, opened 2019-11-18
Autosave fixes

Azhrei:
Issue #762 

Multiple little details about how the AutoSaveManager was screwed up.

Changing Preferences would call `restart()` when it shouldn't (necessarily).  Trying to `loadCampaign()` while autosaving had race conditions, as did `saveCampaign()` while autosaving.  Pausing and restarting the timer for ASM could have issues as well.  And of course, spotless added a bunch of `@Override` annotations... everywhere.

When merged, please delete the branch.  (No reason to leave it hanging around.)

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/919)


Azhrei:
Looks like the change to AppActions overlaps with Craig’s fix for the accelerator key.  I’ll look at it tomorrow and update the PR...

Issue #333, opened 2019-03-18
MapTool 1.5.0 needs tooltips in Preferences


Azhrei:
In particular, the **Startup** tab should have tooltips for the fields that currently don't (all Advanced Options) and tooltips for **Language** and **Data Directory** should be corrected (both are incorrect or incomplete).

PR #928, opened 2019-11-22
Add tooltips for Startup tab

Phergus:
Add/replace tooltips on Preferences Startup tab with i18n strings.   More changes are needed in the XML form for the prefs dialog to get rid of the hard-coded strings and to give all the labels names so they can also get localized tooltips.  Issue #333

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/928)


JamzTheMan:
@Phergus oh oh, spotless got ya! 😄

Phergus:
Dang it!  Should be happy now.

PR #937, opened 2019-11-25
Refactor updateTokenProperty

### LINKED ISSUE #936: Refactor updateTokenProperty and getTokenFromParam
The updateTokenProperty system for updating tokens on the server and clients involves passing a String `methodName` to indicate which token property needs to be updated. The proper way would be to replace the String by a proper `Enum`, which would be both faster and safer.

The function `getTokenFromParam` from `FunctionUtil` needs the `MapToolVariableResolver`. It would make more sense for it to require the `Parser` instead, since that's what we get directly from `childEvaluate`.

The code refactored to solve these two internal issues.

Merudo:
- Change updateTokenProperty to use an Enum instead of a String.
- Partial fix to #936

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/937)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ClientMethodHandler.java
+++ b/src/main/java/net/rptools/maptool/client/ClientMethodHandler.java
@@ -246,14 +246,14 @@ public class ClientMethodHandler extends AbstractMethodHandler {
                 MapTool.getFrame().refresh();
                 return;
 
-              case updateTokenProperty: // select token from sent zoneGUID & tokenGUID, then call
-                // Token.updateProperty()
+              case updateTokenProperty: // get token and update its property
                 zoneGUID = (GUID) parameters[0];
                 zone = MapTool.getCampaign().getZone(zoneGUID);
                 tokenGUID = (GUID) parameters[1];
                 token = zone.getToken(tokenGUID);
                 if (token != null) {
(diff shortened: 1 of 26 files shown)


Phergus:
Changing the Token Terrain Modifier from a BigDecimal to a double may not be a good idea.  BigDecimals have been used in a number of places because you can't rely on a double to give back what you set it to.   Macro code that does something like `if(TM >= 1.75)` may not work as expected if setting the TM as a double to 1.75 results in getTerrainModifier() giving back 1.74999999.  This has definitely been an issue before. @JamzTheMan ?

Merudo:
It was already a double, I only changed the code to make it clearer that it was the case.

JamzTheMan:
Double should be fine, if you want your mods to be 1.549 vs 1.5 or 2 you can, it all gets rounded up to the next cell cost anyway. It's never a calculated value from MT and only holds what you set so it should never return something "unexpected".

Merudo:
Maybe this is a little bit unwieldy? Should I separate the changes into multiple commits / PRs?

Phergus:
That would help.  I haven't had the time to go through such extensive changes and testing it will probably take a lot of time as well.

Merudo:
Ok I reduced the scope of the PR, now it's only about the update token property.

PR #950, opened 2019-11-28
Add function "getAssetProperties" to return properties of asset

### LINKED ISSUE #949: A function to return the dimension of an image from its assetID
It is to my knowledge impossible to get the dimensions of an image from it's asset ID. It makes it difficult to deal with images properly, for example to display an image in a frame of proper dimensions.

A function, maybe called `getAssetProperties`, to return a json holding the properties of an asset. Right now only images are assets, so the json could be something like:

`["type":"image", "width":300, "height", 200]`

Merudo:
- Function takes the asset id as parameter
- Function returns "type", "id", "name", "status", "width" "height".
- Close #949

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/950)


Phergus:
@Merudo  `Asset.getProperties()` is returning width for height:

```
properties.put("height", img.getWidth(null));
```

PR #963, opened 2019-12-03
Update change log for 1.5.8

Phergus:
Fixed some broken formatting

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/963)


JamzTheMan:
@Phergus There are more than just change log changes in here, you probably accidentally updated your branch.

Phergus:
@JamzTheMan  Oops.  Apparently so.   I thought PR 960 was already in 1.5.8-rc.1 but I guess not.  Looks like it will need to be RC.2 instead.  Without that fix getAssetProperties() on table images won't return the correct values.

JamzTheMan:
Ya, I did repoint it to release-1.5.8 branch so it probably wasn't obvious.

Phergus:
Well, I'd say that everything in there can be pushed into a 1.5.8-rc2 build then and in a couple days we can do a release.

JamzTheMan:
ok, wanted to make sure what's in here was ok. I'll merge and create a RC

PR #976, opened 2019-12-08
Fix fill selection box only works on Token layer

### LINKED ISSUE #796: Fill selection box option does not work on layers other than Token
**Describe the bug**
When deselecting “Fill selection box” under “Preferences > Application > Performance”, this works only on the Token layer, but not on the layers Hidden, Object, and Background.

**To Reproduce**
1. Go to 'Preferences > Application > Performance'
2. Deselect 'Fill selection box'
3. Draw a selection box on any layer other than Token (Hidden, Object, or Background). The selection box is still filled.

**Expected behavior**
The selection box should not be filled on any layer, if “Fill selection box" is deselected, just as it is on the Token layer.

**MapTool Info**
- Version: 1.5.6
- Install: Upgrade from 1.5.4

 - OS: MacOS
 - Version 10.15

xdy:
This should fix #796.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/976)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/tool/StampTool.java
+++ b/src/main/java/net/rptools/maptool/client/tool/StampTool.java
@@ -51,6 +51,7 @@ import javax.swing.SwingUtilities;
 import net.rptools.lib.image.ImageUtil;
 import net.rptools.lib.swing.SwingUtil;
 import net.rptools.maptool.client.AppActions;
+import net.rptools.maptool.client.AppPreferences;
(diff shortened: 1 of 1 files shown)


xdy:
Aargh. Sorry. Wrong branch.

Phergus:
Schedule for release 1.5.9.

Phergus:
Tested.  Preference setting now applies to all layers.

PR #977, opened 2019-12-08
#920 Don't download mac dock icon

### LINKED ISSUE #920: RPTools_Map_Logo.png is being downloaded instead of part of package
**Describe the bug**
In `OSXAdapter`, the image used for the macOS dock icon is being downloaded dynamically from RPTools.net and then cached for future use.  It's using an `http://` scheme that seems to fail in Eclipse (but works elsewhere), but we should be populating the JAR with this file instead and referencing it from there.

**To Reproduce**
1. Start MapTool within Eclipse on macOS
2. Note the exception from the PNG not being downloaded and the dock having the generic "Java" logo

**Expected behavior**
With the logo embedded, the image should always be available.

**MapTool Info**
- Version: latest `develop` branch (1.5.7)
- Install: None (Eclipse build)

 - OS: macOS only
 - Version macOS Mojave 10.14.6 (although I believe not OS version dependent)

**Additional context**
No idea why this wasn't embedded into the JAR years ago. :(

Once cached, the name is referred to as `maptool-dock-icon.png` so we should probably just name it that in the JAR and refer to it the same way...

xdy:
Might fix #920, I can't check as I don't have a mac...

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/977)


Phergus:
@Azhrei  - Need Mac user to review.

Issue #1037, opened 2019-12-28
MapTools uses a few dependencies with (according to snyk.io) high severity vulnerabilities.


xdy:
xstream: https://app.snyk.io/vuln/SNYK-JAVA-COMTHOUGHTWORKSXSTREAM-31394 and more medium severity
beanutils: https://app.snyk.io/vuln/SNYK-JAVA-COMMONSBEANUTILS-30077
bouncy castle:https://app.snyk.io/vuln/SNYK-JAVA-ORGBOUNCYCASTLE-32412
jetty: https://app.snyk.io/vuln/SNYK-JAVA-ORGECLIPSEJETTY-32383 https://app.snyk.io/vuln/SNYK-JAVA-ORGECLIPSEJETTY-30090 https://app.snyk.io/vuln/SNYK-JAVA-ORGECLIPSEJETTY-32385 https://app.snyk.io/vuln/SNYK-JAVA-ORGECLIPSEJETTY-460763 https://app.snyk.io/vuln/SNYK-JAVA-ORGECLIPSEJETTY-31117 https://app.snyk.io/vuln/SNYK-JAVA-ORGECLIPSEJETTY-32151 and more medium serverity
(commons collection was also noted by snyk, and kind of scary, but the PR for that is already in)

I'll try to fix them (or note in the build.gradle if we can't upgrade for whatever reason)
I may do the ones only affected by medium security vulnerabilities too (reflections and pdfbox).

xdy:
As far as I can tell this works, but given how wonky the windows installer is this should be tested on someone else's machine. (And probably on a linux and mac machine as well.)

Phergus:
Tested locally.  Was able to build a Windows installer and install.  Basic functionality working.

PR #1042, opened 2019-12-29
#1037 Updated all dependencies with vulnerabilities.

xdy:
Changed all uses of XStream to get an XStream with the security framework to allow classes from:
```
net.rptools.**, java.awt.**, sun.awt.**
```
might be too permissive, but, it works.
Ran gradle assembleDist deploy to make an installer.
Ran installer and started maptool from the installed location.
Loaded an old campaign, ran a server, connected to it, chatted a bit, moved a few tokens, changed fow, scribbled on the map, changed vision, saved and reloaded campign. Loaded images from pdf.
Seems to work.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1042)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -171,7 +171,7 @@ dependencies {
     implementation 'com.github.RPTools:clientserver:1.4.0.0'
     implementation 'org.hibernate:antlr:2.7.5H3'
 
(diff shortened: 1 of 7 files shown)


xdy:
Only tested on windows, and the installer is real wonky on my machine (the locally installed maptool works only occasionally, even on the base develop branch).

Issue #1040, opened 2019-12-29
Should the webapp feature be removed?


xdy:
This issue was created after a discussion on discord, was originally part of #1037, and should probably be done after that one is merged.
The webapp feature hasn't been touched since 2014, isn't being maintained and pulls in jetty as a dependency, which seems unreasonably heavyweight as well as risky. (It also includes ancient versions of bower and jquery. If the feature is kept that should be done some other way.)

I've made a PR that removes it, merge it if it seems like a good idea.

xdy:
Or, rather, *now* I've done the PR, https://github.com/RPTools/maptool/pull/1044

Phergus:
As something like this is only likely way that MapTool will ever support tablets and phones as clients, I'd hate to lose it.   However, as you note, it has languished since its initial exposure and is serious need of updating if it is to be supported versus a hidden/lost experimental feature.

xdy:
Yeah, it'd be better if someone adopted the feature.

xdy:
Closing issue after discussion on discord.

PR #1044, opened 2019-12-29
#1040 Removing the experimental webapp feature as it's been unused an…

xdy:
…d untouched since 2014 and is the only thing that uses jetty.
It also includes ancient versions of bower and jquery, if this feature is not removed that dependency should be done some other way.
Removed everything, including the webpage, js components, classes, etc.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1044)


xdy:
Closing the pull request as the code should be left in.

Issue #1045, opened 2019-12-29
Update all dependencies to the latest version


xdy:
Intended to be a sub-issue of #1041 
I'll start working on this.

xdy:
I've now created a PR, https://github.com/RPTools/maptool/pull/1048 that should update most every dependency that's not in another PR (such as xstream).

Phergus:
Awesome.  Working out if we want to pull this into 1.5.9 or put it in 1.5.10.

xdy:
If you decide leave this for 1.5.10, it'd probably be a good idea to at least do https://github.com/RPTools/maptool/pull/1042 in 1.5.9, unless it breaks on mac or linux of course. (But I bet the problem you had earlier with modules was the uberjar one I fixed in the 1048 PR.)

Phergus:
Pulled in for 1.5.9.   Builds and installs locally.  No issues found.

PR #1048, opened 2019-12-29
#1045 Update all dependencies

xdy:
I've updated everything that isn't in another PR, or is NoiseLib (as that is built using a later java version.)
I've done a fair amount of testing, and it seems to work fine.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1048)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ import org.gradle.plugins.ide.eclipse.model.AccessRule
 
 buildscript {
     dependencies {
-        classpath "com.diffplug.spotless:spotless-plugin-gradle:3.13.0"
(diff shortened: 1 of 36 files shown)


Phergus:
Was it Spotless that made all the changes to the various macro classes?

xdy:
Yep, that was spotless.

Phergus:
That wacky guy.

xdy:
Fixed merge problem (and removed unintended change to MapTool.iss.template)

xdy:
Obviously, this PR fixes all current dependabot PR:s, so any open such should be closed if/when this is merged.

JamzTheMan:
Noting here for future self: Updating log4j2 lib past 2.12.1 currently breaks logging after packaged as uberjar...

PR #1055, opened 2019-12-31
Updates commons-collections to 4.4 and patches the code to actually u…

xdy:
…se it rather than an older version.

When this is merged https://github.com/RPTools/maptool/pull/858 can be closed.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1055)


Phergus:
Dependabot should automatically close out the other PR.

PR #1068, opened 2020-01-02
More Terrain Modifier choices

### LINKED ISSUE #459: [Request] Flat movement cost increase option for Terrain Modifier (difficult terrain) tokens
**Feature enhancement request:**
The terrain modifier token feature applies a cost multiplier.  This is mostly fine, but doesn't work for the Pathfinder 2nd edition Playtest (PF2e) difficult terrain rules.  These are not necessarily the final Patfhinder 2nd edition rules, but it does not seem likely the final rules will change here.

The PF2e rules add a flat 5 or 10 additional feet of movement cost for difficult/greater difficult terrain, without any additional increase for diagonal movement.  Default diagonal movement uses the 5/10/5 alternating diagonal movement cost, and that combination is where the current implementation does not work.

**Current behavior**
_Example using terrain modifier of 2.0 (double cost) to model difficult terrain_
- The multiplier makes it so that the first 5 feet of horizontal or vertical movement costs 10, which is +5, matching what is needed for PF2e difficult terrain.
- Diagonal movement costs 15 for the first square of movement, which is 5 more than PF2e should cost for difficult terrain (default 5 + 5 for difficult terrain).
- The second square of diagonal movement costs 15, which is the same as it would cost for PF2e (10 for second diagonal + 5 for difficult terrain).

The modifier therefore does not model PF2e movement rules on diagonals.

_Example using terrain modifier of 3.0 (triple cost) to model greater difficult terrain_
- The multiplier makes it so that the first 5 feet of horizontal or vertical movement costs 15, which is +5, matching what is needed for PF2e's greater difficult terrain.
- Diagonal movement costs 20 for the first square of movement, which is 5 more than PF2e should cost (default 5 + 10 for greater difficult terrain).
- The second square of diagonal movement costs 25, which is 5 more than PF2e should cost (10 for second diagonal + 10 for greater difficult terrain).

The modifier therefore does not model PF2e movement rules on diagonals.

**Desired behavior**
Add an option (either in campaign properties, or preferences, or on the token itself - just somewhere) to have the terrain modifier option add a configurable flat cost to movement.  Each square of movement would take this additional cost without consideration for the original movement cost.

**Alternatives considered**
- Using the system as-is and mentally adjusting the costs of diagonal movement.
- Ignoring the diagonal movement discrepancy entirely
- Ignoring the terrain modifier feature
- Write macros for movement cost instead of the terrain modifier token feature (might have weird interaction with AI movement feature)

**Additional context**
n/a

### LINKED ISSUE #728: Add 'Ignore Terrain' to a Token
With the new terrain feature, there is no way to tell MapTool to ignore the movement penalty for a token when passing over a terrain modifier token.

Add a checkbox (or other control) to the edit token panel->config tab as well as macros to turn on and off terrain modifier considerations on token movement.

None.

**Additional context**
This could affect #634 and #459

### LINKED ISSUE #748: Extend getlastPath() function to return token centre
When you are using snap-to-size on square grids the getlastPath() function returns coordinates that are the top left boundary point of a cell.

For the purposes of collision detection, it would be useful if a parameter could be added to the function that caused it to return a path based upon the centre point of the token.

As well as being better for collision detection, it would also allow consistent behaviour for all grid types.

JamzTheMan:
* NEW TerrainModifierOperation added to TerrainModifier, the following options are now available:
   * NONE - Default, no terrain modifications to pathfinding cost
   * MULTIPLY - All tokens with this type are added together and multiplied against the Cell cost
   * ADD - All tokens with this type are added together and added to the cell cost
   * BLOCK - Movement through tokens with this type are blocked just as if they had VBL
   * FREE - Any cell with a token of this type in it has ALL movement costs removed
 * The new default value is 0 with an Operation of NONE
 * ADD does NOT use Cell units. e.g. if your Map's distances per cell is 5, and you wanted to add 10 feet per cell, enter 10 for the terrain modifier.
 * NEW Ignore Terrain option
   * Can specify one or more of the above TerrainModifierOperation's. Those selected operations will not be applied for this token.

*Note*: Terrain Modifier Tokens still require the token to not be free-size, for best results use snap-to-grid tokens as calculations are on a cell by cell case.

*2nd Note*: The changes look at least twice as big as it really is. There is a lot of code cleanup along with spotless putting back the {}'s all over the place per Craig's google style request...

Fixes #459, Potentially closes #728 

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1068)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/tool/ToolHelper.java
+++ b/src/main/java/net/rptools/maptool/client/tool/ToolHelper.java
@@ -23,7 +23,8 @@ import java.awt.geom.AffineTransform;
 import java.awt.geom.PathIterator;
 import java.text.NumberFormat;
 import java.util.Set;
-import javax.swing.*;
+import javax.swing.AbstractAction;
+import javax.swing.SwingUtilities;
 import net.rptools.maptool.client.AppUtil;
 import net.rptools.maptool.client.MapTool;
 import net.rptools.maptool.client.ScreenPoint;
@@ -63,7 +64,8 @@ public class ToolHelper {
                   .hideFrame(MapToolFrame.MTFrame.SELECTION.name());
             }
           }
-          Set<GUID> selectedTokenSet = renderer.getSelectedTokenSet();
+
+          Set<GUID> selectedTokenSet = Set.copyOf(renderer.getSelectedTokenSet());
 
           for (GUID tokenGUID : selectedTokenSet) {
             Token token = renderer.getZone().getToken(tokenGUID);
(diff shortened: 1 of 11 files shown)


JamzTheMan:
New UI for Terrain Modifier: 
image

Ignore Terrain option:
image

Phergus:
# HOLD UNTIL JSON UPDATE IS FINISHED

Issue #1079, opened 2020-01-05
Deleting multiple tokens causes ConcurrentModificationException


Phergus:
**Describe the bug**
When deleting multiple tokens an exception is thrown.

```Java
java.util.ConcurrentModificationException
	at java.base/java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:719)
	at java.base/java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:741)
	at net.rptools.maptool.client.tool.ToolHelper$1.actionPerformed(ToolHelper.java:68)
	at java.desktop/javax.swing.SwingUtilities.notifyAction(SwingUtilities.java:1810)
	at java.desktop/javax.swing.JComponent.processKeyBinding(JComponent.java:2900)
	at java.desktop/javax.swing.JComponent.processKeyBindings(JComponent.java:2962)
	at java.desktop/javax.swing.JComponent.processKeyEvent(JComponent.java:2862)
	at java.desktop/java.awt.Component.processEvent(Component.java:6366)
	at java.desktop/java.awt.Container.processEvent(Container.java:2261)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:4966)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2319)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4798)
	at java.desktop/java.awt.KeyboardFocusManager.redispatchEvent(KeyboardFocusManager.java:1950)
	at java.desktop/java.awt.DefaultKeyboardFocusManager.dispatchKeyEvent(DefaultKeyboardFocusManager.java:871)
	at java.desktop/java.awt.DefaultKeyboardFocusManager.preDispatchKeyEvent(DefaultKeyboardFocusManager.java:1140)
	at java.desktop/java.awt.DefaultKeyboardFocusManager.typeAheadAssertions(DefaultKeyboardFocusManager.java:1010)
	at java.desktop/java.awt.DefaultKeyboardFocusManager.dispatchEvent(DefaultKeyboardFocusManager.java:836)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:4847)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2319)
	at java.desktop/java.awt.Window.dispatchEventImpl(Window.java:2772)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4798)
[...]

PR #1081, opened 2020-01-05
Don't modify list while iterating through it...

Phergus:
Fix ConcurrentModificationException when deleting multiple tokens created by fix for NPE when restoring FoW.  Issue #1079.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1081)


Phergus:
Killing it to limit to 1.5.9 code.

PR #1085, opened 2020-01-05
Fix for ConcurrentModificationError when deleting tokens

Phergus:
Make a copy of the set so that deselecting a token as we go through doesn't produce the CME. Fix for #1079

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1085)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/tool/ToolHelper.java
+++ b/src/main/java/net/rptools/maptool/client/tool/ToolHelper.java
@@ -22,6 +22,7 @@ import java.awt.event.ActionEvent;
 import java.awt.geom.AffineTransform;
 import java.awt.geom.PathIterator;
 import java.text.NumberFormat;
+import java.util.LinkedHashSet;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


Azhrei:
Since the deselectToken was removed from removeToken, I take it there aren't any other references to removeToken?

Phergus:
The `deselect()` wasn't in `removeToken()` (as that is in Zone, not ZoneRenderer).  It was in the event listener for Zone.Event.TOKEN_CHANGED, .TOKEN_ADDED, and .TOKEN_REMOVED where I stupidly put it in the previous fix.  It was me adding the `deselect()` call that caused the current issue.  Moved the `deselect()` call up to `deleteTokenAction` in ToolUtil where the removal is kicked off.

There are, of course, a number of calls to `Zone.removeToken()` from different places in the code and it's possible there are pending issues with ghost selected tokens still there.  so they should probably be checked out as they may be the source for some of the NPEs that come in.

PR #1096, opened 2020-01-09
Fix tokenBounds rotation on 'Top Down' tokens.

### LINKED ISSUE #1060: Labels, Halos, and Selection boxes incorrectly render when changing facing on top-down tokens with off-center layouts
**Describe the bug**
I mentioned in #1049 that the selection box seems to rotate weirdly with facing on top-down tokens with off-center layouts. Halos thankfully track the selection box rather than the token image.  

**To Reproduce**
1. On a new map add a new token.
2. Edit the token. In the Config tab, change the Shape to 'Top Down'. In the Layout area, drag the token to the top right corner. Save the token.
3. Right click on the token and add a bright halo color such as Yellow. Optionally, select 'Show Token Names' in the 'View' menu.
4. Change the token's facing to observe the error. Notice you cannot click on the token image to select it, only on the box created by the halo.

**Expected behavior**
The selection box and halo should follow the token image.

**Screenshots**
Here is an example:
image
image

Here is an example showing the misplacement of the label and the VBL at the same time:
image

**MapTool Info**
- Version: 1.5.8
- Install: Upgraded from previous version

 - Windows 10

melek:
Change an incorrect 'x' to a 'y' in ZoneRenderer.java when transforming tokenBounds.

Fixes #1060

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1096)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
@@ -3014,7 +3014,7 @@ public class ZoneRenderer extends JComponent
       Area tokenBounds = new Area(origBounds);
       if (token.hasFacing() && token.getShape() == Token.TokenShape.TOP_DOWN) {
         double sx = scaledWidth / 2 + x - (token.getAnchor().x * scale);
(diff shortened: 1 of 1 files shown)


Phergus:
On hold pending JSON refactor.

PR #1109, opened 2020-01-13
Fixed up some JSON String conversions

cwisniew:
also added  added json.true, json.false, json.null

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1109)


Phergus:
> Added json.null, json.true, json.false variables so you can do
> json.set("", "isTrue", json.true, "isFalse", json.false, "isNull", json.null)
> mind you these are really only useful in this kinda context you cant do [if(x == json.true): ... ; ...] as that will blow up the parser
> plus nothing (At the moment) can return json.true, json.false, json.null

PR #1164, opened 2020-01-21
New command line argument option -load or -l

### LINKED ISSUE #1165: Command Line option to load campaign file on startup
When MapTool starts, I would like the ability to supply a path to the Campaign and have MapTool automatically load it.

Add a -load="<PATH>" to CLI options

Currently I have to use UI interface. This slows down development when MapTool is launched via IDE, especially when done repeatedly and campaigns are in different locations. 

**Additional context**
Having CLI option let's you set multiple IDE configs to different campaigns as well as script the launch of MapTool with campaigns using new test.functions to load multiple MapTool instances with specified campaigns.

JamzTheMan:
* Accepts a path to campaign file as a string and will attempt to load campaign on startup
 * Useful for testing where repeated restarts of MapTool are needed

Closes #1165

Sample: `-load="D:\Google Drive\Map Tool Resources\! Campaigns !\Dev Tests\AStartWalker.cmpgn"`

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1164)


Azhrei:
Is there not an issue filed for this?

Can we remove the option entirely?  If the file provided ends with “.cmpgn”, then open the campaign?  This allows most GUIs to just double click the file to open the associated application.  You could double click a campaign file on Linux and a Windows, at least, to open it.  I’m not sure what would be required on macOS.

It does require that the application be “registered” with the GUI but Windows installers can do that automatically, as can Linux installers.  It’s just a “file association”.  (So can macOS, I’m just not clear on what needs to be done in the installer.  I think the plist file needs the right tag added.)

Phergus:
Needs an issue opened.

JamzTheMan:
Ya I was looking for it, I thought there was but words fail. I'll create one later.

@Azhrei double click to launch is a different use case. That is for installers only, this is for command line option l, I'm mostly using for debugging in an IDE.

File association is a thing but I never could get it to work with the current packager for some reason.

The new packager for JDK 14 is suppose to support it and I had planned to add file association then, but that doesn't help launching MT from CLI.

JamzTheMan:
@phergus see #1165

Phergus:
> File association is a thing but I never could get it to work with the current packager for some reason.

We can definitely put the current command line options into the shortcut for the MapTool.exe which means the options are getting passed on.  Wonder what's up with the file association?

Azhrei:
The handling of options in `main()` needs rework.  There should be a single `DefaultParser` created and the `parse()` method should be called once.  Then the individual options can be queried from the `CommandLine` object that comes back.  That same object also allows leftover strings on the command line to be retrieved as well.

JamzTheMan:
New commit to change `-l` / `-load` to `-F` / `-file` flags.

PR #1169, opened 2020-01-22
Added support for positional parameter as campaign file name on cmd line

Azhrei:
Refactored use of Apache command line package.  #1165 

Added Sentry reporting of implementation version and user override value of version.

Replaced stack trace dumps with calls to `showWarning()` and `showError()` (which display a dialog to the user and add to the MapTool log).

Moved restart of autosavemanager to after `loadCampaign()` is called during `postinitialize()`.

( A lot of spotless cleanup for some reason...)

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1169)


JamzTheMan:
cool beans

Issue #1087, opened 2020-01-06
Frame is always in front of dialogs


mrkwnzl:
**Describe the bug**
When a frame is open, dialogs open behind the frame. The frame cannot be moved or closed until the dialog is closed, but the dialog cannot be closed if it’s behind the frame.

**To Reproduce**
1. Open a frame.
2. Open preferences. The frame is on front the preferences
3. Try to close the frame. It can only be closed after the preferences have been closed.

**Expected behavior**
Dialogs that absolutely need an action (such as preferences, campaign properties, or the dialog to delete a macro) should open on top of every other window. In some cases, the dialog cannot be reached if there’s a frame in front of it.

**MapTool Info**
- Version: 1.5.10
- Install: Upgrade from 1.5.8,

**Desktop (please complete the following information):**
 - OS: MacOS
 - Version 10.15.2

Phergus:
Appears to be specific to MacOS.  Under Windows the Preferences, Campaign Properties, and Edit Token dialogs all open on top of open frames.

@Azhrei - thoughts?

Azhrei:
(Tested with 1.5.9 snapshot install that I had available.  Inside a dev environment might be different, so I didn't use Eclipse.)

I tried what the OP tried, and the campaign properties and preferences opened on top of all frames, floating or docked.

However, when I right-clicked on a macro in the **Campaign** panel and chose **Delete...**, it opened on top of the main window but underneath the floating **Chat** panel.  So it is confirmed, at least for that type of dialog and on macOS.

I'm doing this as my normal user account and it sometimes has other issues with MT/Java, so I'll try a fresh user account as well, just to be certain...  Yes, it happened there instead.  The **Delete...** popped _under_ the frame, and so did the **Quit** confirmation when I terminated MapTool with a keystroke.

Perhaps related to the update of the JIDE libs?  Need to go back to an older build and test there.  I'll do that later tonight if no one beats me to it.  (Don't have time to download and test right now.)

mrkwnzl:
Since I’m working a lot with frames as character sheets, this is really annoying. I’ve had it happen more than once, where I got stuck as I forgot to close the frame before editing or deleting a macro, and I had to force quit MapTool (and lost what I did), because I couldn’t reach the dialog behind it. 

Is there anything I could try to help with this issue?

Phergus:
Only thing I know if is that if the frames are docked then the dialogs can't open behind them.

Azhrei:
[...]

PR #1170, opened 2020-01-22
macOS mouse wheel issue

### LINKED ISSUE #413: Facing is set when using a BT mouse on macOS
**Describe the bug**

**To Reproduce**
1. Start MapTool.

**Expected behavior**

**MapTool Info**
- Version: 1.5.1
- Install: New (from .DMG)

 - OS: macOS

**Additional context**

Azhrei:
Fixes #413 

Filters all `MouseWheelEvent`s with the `Shift` modifier turned on when the platform is OSX.  This eliminates horizontal scrolling so that token facing isn't randomly changed by accidental left/right motion on the mouse.

This is a short-term fix.  A longer term fix is being worked on.  The issue has more, including a link to a full description of the problem and proposed solution.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1170)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppActions.java
+++ b/src/main/java/net/rptools/maptool/client/AppActions.java
@@ -14,6 +14,7 @@
  */
 package net.rptools.maptool.client;
 
+import com.jidesoft.docking.DockableFrame;
 import java.awt.Dimension;
 import java.awt.Event;
 import java.awt.Graphics2D;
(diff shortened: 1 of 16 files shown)


cwisniew:
This PR changes more than just the mouse wheel it is also changing the behaviour of several dialogs.

This new behaviour may be correct but it should be in a separate PR for that issue otherwise someone testing wouldn't even know they are supposed to be testing for these dialogs and if there is a problem caused then the mouse fix would have to also be reverted.

Also the XXX comments are confusing for other people reading the code especially those that are questions, if there is a problem an issue should be created, please dont create a personal TODO/Follow up list in the code base

cwisniew:
hit wrong damn button

Azhrei:
> This PR changes more than just the mouse wheel it is also changing the behaviour of several dialogs.

Crud, I branched off of my fix for #1087 instead of `develop`.

Sigh.  I don't do this often enough.  I'll cancel the whole thing and figure out how to cherry-pick just the last few commits into a new branch.

Issue #1178, opened 2020-01-23
AI distance calculations incorrect when using terrain modifiers


JamzTheMan:
**Describe the bug**
In some paths taken when including diagonals across terrain modifiers, costs could be incorrect for some cells due to rounding errors.

**To Reproduce**
Use a Multiplier 4 terrain modifier and move token diagonal several times and note the costs.

**Expected behavior**
Correct math!

**Screenshots**
See discord channel....

**MapTool Info**
- Version: 1.5.10 (probably back to 1.5.0)
- Install: Any

**Desktop (please complete the following information):**
 - OS: Any

**Additional context**
A sample campaign I use to run tokens through...
[AStartWalker.zip](https://github.com/RPTools/maptool/files/4100856/AStartWalker.zip)

Phergus:
**Changes**
 * **New Map Option** for AI rounding of cell costs, NONE, CELL_UNIT, INTEGER. More can easily be added if needed.

**Debugging Options**
 * ZoneRenderer now show blocked movements if class log debug is on (previously required manual boolean change while developing)
 * ZoneRenderer now shows raw cost without cell units while debugging when showing token Path

Phergus:
New rounding modes and terrain modifier types tested with low to medium complexity maps.

**Note for Documentation:**  Terrain Modifiers on tokens/objects only apply when the token/object has a _set size_.  Free Size will be treated as a single cell with TM at the center.

Phergus:
[Introduction to Tokens](http://www.lmwcs.com/rptools/wiki/Introduction_to_Tokens#Config) page updated with screenshot and some verbage.

Phergus:
Add a basic [Map Properties](http://www.lmwcs.com/rptools/wiki/Map_Properties) page to wiki.

Phergus:
Added new info to the [AI Pathfinding](www.lmwcs.com/rptools/wiki/AI_Pathfinding) wiki page.

PR #1179, opened 2020-01-23
Fixes AI Terrain Modifier distance calculations

JamzTheMan:
* Previously the Distance calculations relied on rounding where all diagonals internally cost 1.5 vs 1-2-1.
   Adding terrain modifier costs introduced additional decimal rounding and terrain could not always be multiplied by 1.5 but rather 1 or 2 depending.
   Additional variables and methods added to CellPoint and related classes to track costs to correct the issue.
 * New Map Option for AI rounding of cell costs, NONE, CELL_UNIT, INTEGER. More can easily be added if needed.
 * Some refactoring of variable names in AStar classes, longer but more clear IMO
 * ZoneRenderer now show blocked movements if class log debug is on (previously required manual boolean change while developing)
 * ZoneRenderer now shows raw cost without cell units while debugging when showing token Path
 * Lots of {}'s added to match style guide

New map option will need documentation and word smithing...

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1179)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/MapPropertiesDialog.java
+++ b/src/main/java/net/rptools/maptool/client/ui/MapPropertiesDialog.java
@@ -32,7 +32,21 @@ import java.awt.image.ImageObserver;
 import java.io.File;
 import java.io.IOException;
 import java.util.List;
-import javax.swing.*;
+import javax.swing.AbstractAction;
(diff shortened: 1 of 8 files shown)


JamzTheMan:
Closes #1178

PR #1182, opened 2020-01-23
macOS mouse wheel fix

### LINKED ISSUE #413: Facing is set when using a BT mouse on macOS

Azhrei:
Fixes #413 

There's one strange aspect of this.  The _inertial scrolling_ on the Mac means that I can "spin the mouse wheel" (so to speak; I'm actually using a touch sensitive surface) while changing the facing, and if I release the Shift key, the inertial scrolling doesn't stop so the action switches to changing the zoom level instead.  Those folks with a Mac should try it; it's difficult to describe if you're not familiar with inertial scrolling.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1182)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppUpdate.java
+++ b/src/main/java/net/rptools/maptool/client/AppUpdate.java
@@ -233,6 +233,7 @@ public class AppUpdate {
 
     Runnable updatethread =
         new Runnable() {
+          @Override
           public void run() {
             try (InputStream stream = assetDownloadURL.openStream()) {
               ProgressMonitorInputStream pmis =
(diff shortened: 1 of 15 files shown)


Azhrei:
Hm.  Looks like some tool in the chain (spotless? Eclipse?) added `@Override` to a bunch of methods.  I must've touched those files at some point and it caused the tool to make the changes.

@cwisniew There are two **XXX** comments in here, but they both relate to the mouse wheel.  Hopefully, they're not deemed "off-topic". :)

cwisniew:
> @cwisniew There are two **XXX** comments in here, but they both relate to the mouse wheel. Hopefully, they're not deemed "off-topic". :)
Comments that add no value and serve only to confuse are always off topic :\

Issue #1056, opened 2019-12-31
Remove unused dependencies


xdy:
Intended to be a sub-issue of #1041
Initially I'll run https://github.com/nebula-plugins/gradle-lint-plugin/wiki on the project.
Seems I'll also need to make PR:s on the other rptools projects maptool depends on. I won't make issues for those, just reference this issue in those PR:s. (I know I'm in a minority, but, well, I *like* monorepos. I'm lazy. :)

xdy:
Done first round of PR:s on the sub-projects (upgrade gradle to 5.3.1)
https://github.com/RPTools/clientserver/pull/11
https://github.com/RPTools/parser/pull/17
https://github.com/RPTools/maptool-resources/pull/1
https://github.com/RPTools/dicelib/pull/20

I'll unassign myself from this one for now, if anyone feels like grabbing it, go ahead.

Merudo:
I believe the 11 following dependencies are unused:

- Antlr: added before 1.4, used to compile other RPTools project but not this one
- decktool: added before 1.4, appears to be at the prototype stage
- 4 jide dependencies: not needed by the jide elements we use
- jetty-servlet, websocket-client: added before 1.4, not used by webapp
- jetty-continuation: deprecated, added before 1.4, not used by webapp
- ca.odell.renderpack: fetched from Craig's website. Contains class HTMLTableCellRenderer, HTMLRenderer, HTMLListCellRenderer, but these are not used. Added before 1.4.
mockito-core: for mocking features during unit tests, but is not used by any test. Added in 1.5.0.

Phergus:
Builds and runs after changes.

PR #1195, opened 2020-01-25
Remove unused dependencies found by gradlelint

Merudo:
- Remove 11 unused dependencies found by gradlelint
- Change to runtimeOnly: jide-shortcut, maptool.resource, commons-logging, commons-beanutils, slf4j-simple, jxpath, commons-jxpath, ezmorph, jai-imageio-core, imageio-core, imageio-jpeg, imageio-psd, xpp3_min, xmlpull, and bcmail-jdk15on
- #1056 for a general discussion of dependencies removal

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1195)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -156,8 +156,8 @@ dependencies {
     implementation group: 'org.apache.logging.log4j', name: 'log4j-api', version: '2.12.1'
     implementation group: 'org.apache.logging.log4j', name: 'log4j-1.2-api', version: '2.12.1'	// Bridges v1 to v2 for other code in other libs
 
-    implementation group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.30'
-    implementation group: 'commons-logging', name: 'commons-logging', version: '1.2'
+    runtimeOnly group: 'commons-logging', name: 'commons-logging', version: '1.2'   // Dependency of jxpath
+    runtimeOnly group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.30'         // Required to load class org.slf4j.impl.StaticLoggerBinder
 
     // For Sentry bug reporting
     implementation group: 'io.sentry', name: 'sentry', version: '1.7.29'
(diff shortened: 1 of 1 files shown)


JamzTheMan:
Umm, expect things to break after this merge, not sure I'm trusting that plugin, I see more than one lib that should be needed...

Merudo:
If the code refers to a dependency only via reflection, gradlelint will mistakenly see it as unused.

I suggest we move such dependencies to `runtimeOnly` and document the reflection used.

Phergus:
Did you try a `gradlew deploy` or `gradlew uberJar` after this change?

Merudo:
> Did you try a `gradlew deploy` or `gradlew uberJar` after this change?

I did and there is no issue, but missing dependencies only needed at runtime via reflection could still cause `NoClassDefFoundError` errors.

Phergus:
Okay, so I really meant did you then go on to test that the installer build and uberJar both worked?

Merudo:
As far as I can tell, they work. But I probably didn't try every feature of MapTool.

For example, `org.bouncycastle:bcmail-jdk15on` is said to be used to "decrypt passworded/secured pdfs". Can MapTool even do that?

Phergus:
Well MT can extract images from PDFs that have password protection but all the PDFs I've checked here had Copy Content allowed even with the password protection.  So I'm not sure when it might be needed.  @JamzTheMan ?

Merudo:
Closed because of the likelihood of breaking an obscure feature outweighs the benefits of removing the dependency.

JamzTheMan:
FWIW MT can not opened password locked PDF's at this time but it was required by PDFBox to open some "secured" PDF's (ones just signed)

Also the without the various imagei libs MT wouldn't always extract every image, so it may seem to work but you need to compare images extracted.

Merudo:
Yep, the image libs are loaded at runtime by ImageIO, which is why the PR had it changed from `implementation` to `runtimeOnly`.

JamzTheMan:
Ya, I have no problem changing them from implementation to runtimeonly. We can take that up in a separate PR.

PR #1207, opened 2020-01-28
Minor json code clean up and trapping when function isnt defined

### LINKED ISSUE #1206: JSON macro functions will behave incorrectly if function is not found
**Describe the bug**
The json function code returns null if the function is not found. This should throw an exception with a valid error message. Currently neither of these will occur but the exception should be put in place to catch any problems with future edits.

cwisniew:
fixes #1206

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1207)


Merudo:
Could the error message be localized via I18N (`macro.function.general.unknownFunction`)?

src/main/java/net/rptools/maptool/client/functions/json/JSONMacroFunctions.java
```diff
@@ -518,8 +518,7 @@ public Object childEvaluate(Parser parser, String functionName, List<Object> arg
           }
         }
     }
-
-    return null; // TODO CDW
+    throw new ParserException(functionName + "(): Unknown function");
```
Merudo:
```suggestion
    throw new ParserException(I18N.getText("macro.function.general.unknownFunction", functionName));
```

PR #1214, opened 2020-02-01
Feature token vbl simplification

JamzTheMan:
Early draft of token polygon simplification using JTS libs

TODO:
- [ ] ALL the UI, but must first determine what UI is needed
- [x] Create Issue
- [ ] Going to need some wiki docs

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1214)


JamzTheMan:
Used existing UI elements for early draft:

1. Generate Tokenl VBL as usual
2. Select level of detail using VBL Sensitivity
3. Use Visibility Tolerance to drive JTS Lib "tolerance". Setting to 1 turns it off, 2-9 sets it to 1-8

image

JamzTheMan:
Note: Currently method is hard-coded to Douglas Peucker method, see the TokenVBL.createVblArea class/method to change enum to VW Simplifier ofr Topology Preserving Simplifier if you want.

Phergus:
See #1218 for further discussion.

Issue #1231, opened 2020-02-03
Choosing a non-writable directory fails silently while downloading a new version


emmebi:
**Describe the bug**
When MapTool tries to download a new version, it fails with no error message if it can't write on chosen directory.

**To Reproduce**
1. Download a version which is not the latest (e.g. 1.5.10)
2. Run it
3. Accept to download a new version
4. When asked where to download choose c:\ (or similar, non writable by a normal user directory)

**Expected behavior**
1. At least, you should receive an error message stating the nature of the error (e.g. you can't write in that directory)
2. Ideally, the tool should ask you to choose a new directory

**MapTool Info**
- Version: 1.5.10
- Install: New

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: 10

Phergus:
@emmebi  Was there an issue with having it prompt for a new directory or where you just going to pass on that for now?

Phergus:
Currently user is warned if chosen directory is not writable by user.

emmebi:
@Phergus I didn't think of that; let me see.

Phergus:
Excellent.  I'll wait on closing this till you report back.

emmebi:
I actually have spotted a problem with previous version; I am fixing this, while asking for a new file location; I probably will push an update this evening (UTC time).

Phergus:
Very good.  I'll keep an eye out for it.

PR #1235, opened 2020-02-03
Adds an error message when a new version of MapTool can't be saved

emmebi:
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1235)


Phergus:
Addresses #1231

PR #1266, opened 2020-02-08
Decouples AssetPanel from Resource window so that each one can browse independently

### LINKED ISSUE #256: Drawing Foreground/Background Texture Select Using Resource Library Directory
When selecting a drawing texture on the draw widget, the images shown in the initial window are from the directory selected on the Resource Library, not the directory selected on the Popup select window.

emmebi:
Tests done:
- new map/map => success
- new map/background => success
- drawing/choose paint => success

Fixes #256

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1266)


JamzTheMan:
Nice, that's been an annoyance for a while now!

Issue #1267, opened 2020-02-09
Null Pointer Exceptions for Zone fields occuring during Campaign Load and Auto Save


cwisniew:
**Describe the bug**
During both Campaign Load and Auto Save a null pointer exception occurs some times when the Zone is accessing its fields. This is due to the inherent lack of thread safety in Campaign/Zone/(you could keep going most of the way down) which results in other threads "seeing these fields before seeing the completion of initialisation of these objects". 

While this is down to the the lack of thread safety in these classes the code for auto saving has comments saying that because the copy happens on EDT it is safe. This comment is completely incorrect, this does nothing to address the threading issues at all. For clarity sake multithreading issues/bugs are not just limited to concurrent access issues but also encompass state visibility.

While there is a larger issue of how exactly state persistence should work in a threaded environment where state can change during the saving process to be had (I have created issues #1263 and #1264 for this). A shorter term solution is also required to address the null pointer/incomplete object initialization problem.

Phergus:
Tested.  Previous test cases no longer produce NPEs.  Closing.

PR #1273, opened 2020-02-10
Fixes for various reasons NPE can sometimes occur during campaign load / autosave

cwisniew:
Fixes for a few issues causing null pointer exceptions during campaign load and save.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1273)


Phergus:
Fixes for #1267

Issue #1270, opened 2020-02-10
Can't delete a map image


Merudo:
**Is your feature request related to a problem? Please describe.**
Once a map image has been set (Map -> Edit Map -> Map), it is impossible to remove the image.

The image can be replaced by another image, but not removed.

**Describe the solution you'd like**
There should be an extra button to remove the image.

Alternatively, pressing the cancel button should remove the image.

amccarthy1:
How would something like this look:

![image](https://user-images.githubusercontent.com/5650580/75419096-ecb30180-58e9-11ea-8d7a-e4a1e9af6210.png)

The "Clear" button just resets the image. If this is acceptable / desired functionality I can open a PR.

Phergus:
I think that I would have put it on the Map Properties.  Probably like can be see on the edit token dialog for Portraits and Handouts.  Something like this:

![RemoveMapConcept](https://user-images.githubusercontent.com/34379254/75450150-82bb4c00-592b-11ea-85ce-37484d560fab.png)

There is a little `cancel_sm.png` image under resources used on the token config panel.

Or the current Map button could be changed to Add Map with another button under it that said Remove Map.

Or the current Map button could change text depending on if the zone already has a map.  It would say Remove Map if there is one and Add Map (or just Map) if there isn't.

JamzTheMan:
I like that design Phergus. Assume we would only sure the X when there is a map image added.

Phergus:
Yeah.  Just like the token config panel.  Simple and pretty obvious what it does I think.

Phergus:
Tested

- [ ] Map menu -> New Map -> Add Map
- [ ] Map menu -> New Map -> Add Map -> Remove Map
- [ ] Map menu -> Edit Map -> Add Map
- [ ] Map menu -> Edit Map -> Remove existing map
- [ ] Map menu -> Edit Map -> Replace existing map
- [ ] Resource Library -> New Map ->  Remove
- [ ] Resource Library -> New Map ->  Remove -> Add different
 
No issues found.

PR #1328, opened 2020-02-27
Add a clear button to reset the map

amccarthy1:
This adds a button that resets the map to `null`, since this behavior is not possible right now.

Screenshot of the modified UI can be found in the linked issue.

Addresses issue #1270

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1328)


src/main/java/net/rptools/maptool/client/ui/MapPropertiesDialog.java
```diff
@@ -594,37 +594,42 @@ public void actionPerformed(ActionEvent e) {
...
+    /**
+     * Open the dialog and allow the user to select an asset for the map.
+     *
+     * @param asset - The currently selected asset.
+     * @return The selected asset. If "Clear" was clicked, returns null.
+     */
+    public Asset chooseAsset(Asset asset) {
+      originalAsset = asset;
       setVisible(true);
+      originalAsset = null;
```
emmebi: Why the double assignment?

amccarthy1: I did this to allow the previous asset to be garbage collected once it's been swapped or cleared. Might not actually make a difference depending on when the whole dialog gets GC'd though.

emmebi: I understand that, just not getting why putting to a value and then to null later. My understanding is that you are going to use originalAsset to restore the status of this in case the user hits the cancel button. However, at the end of this method originalAsset is always null.

So this will work if you selected a map and then hit the clearSelection, or if you had a map with an image selected and you want to remove it, but not in the case that you have a map with an image selected, select a new image, and then hit the cancel button: in that case I expect that the selectedAsset will become null.

I can't test it now, will do it later.

Phergus: Also see my comments on the issue.

amccarthy1:
Going to change the UI based on the comments on the issue, should be able to get to this tonight.

Phergus:
Sounds good!

Phergus:
@emmebi   Need you to sign off on the discussion.

PR #1395, opened 2020-03-20
WIP - Optimized TokenVBL

JamzTheMan:
* Performance increase upwards of 50x
 * doubleclick on preview window resets preview scaling
 * No long requires 'valid geometry' which allows for vbl with 'holes'
 * Fix tokenPropertiesDialog.xml to use generic label vs RSyntax beans
   so they are not needed when using forms designer which just removes
   the component if you didn't have forms designer setup properly.
 * Added threading in a few places for performance and for proper UI updates
 * Added Color Picker; you are no longer restricted to just transparency but can now pick a color. Although a popular color like Black may be overzealous.
 * Lots of bug fixes and improvements
 * Updated UI to final preview
 * Tested with very large maps (10,000 x 10,000 pixels) takes ~30-40 seconds but UI no longer locks up and dialog can be closed. Most objects take <1 second.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1395)


Azhrei:
This is a pretty big patch; I'll get to it tonight or tomorrow afternoon.  It sounds pretty sweet, though!

JamzTheMan:
@Azhrei Yea and it's only partial.  FWIW you don't have to spend TOO much time on it. The final/full review should happen when it goes back to develop branch.

Or rather, you'll have to review it again when merging to develop. :)

Mostly here for feedback...

Merudo:
Very nice! I'm excited for this to be merged to develop!

I really like the new token VBL menu. I just wish the optimization methods would fit into the dropmenu.

JamzTheMan:
> Very nice! I'm excited for this to be merged to develop!
> 
> I really like the new token VBL menu. I just wish the optimization methods would fit into the dropmenu.

Thanks, the heavy lifting goes to the JTS lib :)
For the drop down menu, I suppose you want it to optimize all tokens in one go then?  I plan on adding the optiimzation options to the macro function next so hopefully that lets everyone "fix" existing tokens in mass...

I also plan on (in another issue) taking a crack/pass on adding this to the map level VBL. 
Another issue is also planned to optimize the VBL by replacing all the references to Awt Area class with the JTS geometry class, as adding those polys together is much faster.

PR #1490, opened 2020-04-08
Issue413 mac mouse wheel

### LINKED ISSUE #413: Facing is set when using a BT mouse on macOS

Azhrei:
Removes horizontal scrolling via Apple's mouse/trackpad but still allows user to horizontal scroll by holding down the Shift key and using "scroll up/down"

Closes #413

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1490)


Azhrei:
Tested with Apple MagicMouse 2 and Apple Wireless Trackpad

Phergus:
@Azhrei   Somebody forget to run spotless.

Azhrei:
(Sigh.)  Done and pushed...

PR #1492, opened 2020-04-09
Add "blockopaque" value for --pointermap

Merudo:
- Add "blockopaque" value for --pointermap. An element with the style --pointermap:blockopaque; will block clicks if the pixel at the click location is opaque (alpha == 0), and will let the click go to the map if the pixel is transparent
- Discussed in #1485

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1492)


JamzTheMan:
This sounds backwards, shouldn't blockopaque block clicks on pixels that are opaque?

Merudo:
Yep! My bad, I had the PR description wrong.

PR #1502, opened 2020-04-10
Fix user data directory message when running MapTool twice

### LINKED ISSUE #1501: "User data directory is already in use" with frame5/dialog5 on two instances of MapTool
**Describe the bug**
When two instances of MapTool on the same machine call frame5 or dialog5, an ugly error message is printed to the chat: `User data directory [null (C:\Users\user\AppData\Roaming\java\webview)]`.

**To Reproduce**
1. Start two instances of MapTool on the same machine
2. Run `[frame5("test"):{test}]` in both
3. Look at the error message

**Expected behavior**
No error message.

**MapTool Info**
- Version: 1.6.0 RC1
- Install: New

 - OS: Windows
 - Version: 10

**Additional context**
I do not believe it is possible to turn off the user data directory. The user directory is not used by frame5 and dialog5 anyway and so I suggest we just hide the message.

Merudo:
- Hide "user data directory is already in use" error message, as the user data directory is not used anyway
- Close #1501

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1502)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLJFXPanel.java
+++ b/src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLJFXPanel.java
@@ -235,7 +235,10 @@ public class HTMLJFXPanel extends JFXPanel implements HTMLPanelInterface {
    * @param event the error event
    */
   private static void showError(WebErrorEvent event) {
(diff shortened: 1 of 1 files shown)


Merudo:
This merged got reverted somehow?

Phergus:
I think it never made it back to develop from the branch.

PR #1514, opened 2020-04-11
HTML Overlay

Phergus:
Merge HTML overlay branch to develop.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1514)


JamzTheMan:
Looks good. I'm going to give @cwisniew a chance to review as well as I know he had questions on how the mouse clicks were being handled and passed through just in case he has any last minute questions or concerns.

Phergus:
Sounds like a good idea.

Merudo:
In case this is helpful:

The mouse events are first captured from the JFXPanel containing the overlays. If the mouse event is a drag/movement/etc, the event gets dispatched to the current zone renderer via `passMouseEvent`. So, it's as if the overlay wasn't there.

In case of a mouse click or press, the click is first ran through `validateAndPassEvent` to see if it should be forwarded to the current zone renderer or not. The validation is done by the webView, which check the value of `--pointermap` for the html element clicked. If the value is `--pointermap:block`, the mouse event is not forwarded; if it is `--pointermap:pass`, it is forwarded, and if it is `--pointermap:blockopaque`, the transparency of the point clicked determines if the click goes through or not.

PR #1530, opened 2020-04-13
1.6 maintenance

JamzTheMan:
Release 1.6.1 to Master

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1530)


JamzTheMan:
@Phergus Not sure why but there is a merge conflict on the Japanese i18n file. Can you take a look? I would assume the one from 1.6 is the correct version?

Phergus:
That's weird.  Checking it.

Issue #1218, opened 2020-02-01
Optimize Token VBL Generation


Phergus:
**Is your feature request related to a problem? Please describe.**
Token VBL generation produces very complex VBL structures leading to poor performance.  Especially when used on map images.

**Describe the solution you'd like**
Use the polygon simplification of the JTS libs to reduce the number of points and edges produced by the VBL generation.

**TODO:**
- [x] ALL the UI, but must first determine what UI is needed
- [x] Create Issue
- [x] Take work off the UI thread?
- [x] Going to need some wiki docs

**Additional context**
[JTS Lib Simplify](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/simplify/package-summary.html)

Campaign file with two maps and closed/open object images.
[VBL Simplified Test-cmpgn.zip](https://github.com/RPTools/maptool/files/4142897/VBL.Simplified.Test-cmpgn.zip)

Phergus:
D-P simplify cannot handle images with holes but does a great job on images that are otherwise solid.

**MT 1.5.12**
![Token VBL 1 5 12](https://user-images.githubusercontent.com/34379254/73596240-d75ed900-44dd-11ea-8ab7-1c749daffbaf.png)

**MT 1.6 develop**
![Token VBL 1 6 dev](https://user-images.githubusercontent.com/34379254/73596245-e47bc800-44dd-11ea-907b-8e85f7bd47aa.png)

See campaign attached to top for closed form images.

Phergus:
Need to test with the other simplify methods.

JamzTheMan:
I think I one of the methods had a second option that may need to be explored.

I assume we want holes for things like caverns and such.

I'm liking the reduction in points though!

Phergus:
Need to look at the options and see if they can help.

JamzTheMan:
Lots of options added to the VBL panel 😨 
Most are temporary until we find the best JTS method (I don't think we need to supply an option here, and will always optimize I think is best)

Color picking isn't enabled yet but I see that replacing "Sensitivity" (the old Alpha Sensitivity" and let you pick a RGBA instead. So you could pick "black" vs just alpha.

Added an Inverse button so you can generate VBL around an object vs over now, good for maps!

[...]

PR #1544, opened 2020-04-14
Fix token editor pick VBL color cursor working once

Merudo:
- Fix VBL color cursor not longing changing to the crosshair cursor after being used once
- Issue discussed in #1218

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1544)


JamzTheMan:
Really, that's all it took lol? Thanks for fixing this!

Merudo:
Yep it's all it took :) I have no idea why Java would behave like that.

PR #1577, opened 2020-04-16
Adds option to extract all pages as images from pdfs in the resource …

### LINKED ISSUE #1518: MapTool should be able to extract full pages (as a pdf viewer would render them using default layers) from a pdf
In many non-paizo pdf:s (such as the Mongoose ones) extracting images gets one only the various parts and layers of the full image as separate files. I.e. nothing that can be used as a map or token.

If I could extract the whole page from the pdf this would often help. It's easy enough to mask out any 'non-player' parts with VBL or drawings.

Keep using printscreen, save to disk, load in maptool.

**Additional context**
I have this implemented on a branch (https://github.com/xdy-forks/maptool/tree/extractPdfPage), and will make a PR when the issue is assigned to me.
The way I've implemented it does what I wanted it to do, but has the effect that all pages in the pdf are extracted as images (as it hooks into the 'extract all images from a page' code), with a DPI of 96 (as 72 was just too grainy for me). The added time to process a pdf isn't enough to bother me (haven't really measured the time though.)
While doing this - or, rather, while fixing the bugs I caused by doing this :) - I also fixed a few warnings, added some javadocs, improved logging around this a bit, and even tested pdfs from various origins.
Maybe change to save 'whole page' files to separate directory, or just clear name standard.
Maybe change so this feature is turned off by default (use something like -Dmaptool.pdf.extract_rendered_pages=true to turn it on)

**Possible future improvements**
Make it possible to turn this on and off using the gui.
Make it possible to change the DPI.
Make it possible to only get the pages that are 'mostly image'.
Make it possible to 'split' a page into several images based on some criteria/get only a certain part of a page (useful for pdf:s with geomorphs).

xdy:
…library.

Should fix #1518

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1577)

### DIFF
--- a/src/main/java/net/rptools/lib/image/ImageUtil.java
+++ b/src/main/java/net/rptools/lib/image/ImageUtil.java
@@ -34,6 +34,7 @@ import java.io.File;
 import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Arrays;
 import java.util.Map;
 import javax.imageio.ImageIO;
 import javax.swing.ImageIcon;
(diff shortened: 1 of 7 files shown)


JamzTheMan:
@cwisniew We can no longer "update branch" or at least unable to update branch for this PR.

Maybe something with latest GitHub administrative changes?

JamzTheMan:
@xdy Unless we can figure out why, can you do a `pull -r upstream develop` to bring your branch up to date and push?

xdy:
That's weird. Anyway, I did that now, I guess we'll see whether that works.

Phergus:
That got it.  I think you may have unchecked "_Allow edits from maintainers_." when you did the PR.

unknown

All better now. :)

xdy:
Could be, not knowingly though. Have to make sure it's checked next time in case it defaults to the last PR:s value.

PR #1584, opened 2020-04-17
Give an option to SwingUtil.constrainTo to limit whether an image can…

### LINKED ISSUE #1583: Restrict portrait in overlay to maximum image size when resizing with SwingUtil.constrainTo

Currently the statsheet overlay with portrait turned on is forced to a static value from AppPreferences.getPortraitSize. I've bumped up that setting to a large value so I can get a decent sized portrait to show up (larger than 175). This works for those pics that are large enough:

image

Smaller images get stretched to my new limit of 600 though, which leads to ugly stretching.

image

I think it would be better if the portrait doesn't get stretched beyond what its native resolution is. In my 2nd case above

image

There could be another option in the preferences but that dialog seems overloaded already.

**Additional context**

I have a fix ready and figuring out how to make that a pull request.

nmeier:
… grow beyond its natural size.

Make portrait in overlay/statsheet not grow beyond image size.

Overload existing constrainTo since it's used in many places - all of those will continue to stretch images up for now.

Fixes #1583

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1584)

### DIFF
--- a/src/main/java/net/rptools/lib/swing/SwingUtil.java
+++ b/src/main/java/net/rptools/lib/swing/SwingUtil.java
@@ -143,12 +143,18 @@ public class SwingUtil {
   }
 
   public static void constrainTo(Dimension dim, int size) {
+    constrainTo(dim, size, true);
+  }
+
+  public static void constrainTo(Dimension dim, int size, boolean grow) {
     boolean widthBigger = dim.width > dim.height;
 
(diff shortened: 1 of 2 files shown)


nmeier:
spotlessApply and spotlessCheck passed locally
tested with images that are either width or height max'd

PR #1622, opened 2020-04-19
Add more token sizes, up 10x10.

### LINKED ISSUE #1621: More sizes for square grids would be useful
My DM just asked if it was possible to have a 5x5 size token in MapTool (square grid).

I'd like to be able to select among a wider variety of token sizes on a square grid. 10x10 sounds like a reasonable maximum size (the current implementation is a hard-coded list read from an xml file.)

Just Say No. But, he's my DM, there would probably be Consequences...
Use free sizing and turn off snap to grid. Ok for a single token, but it gets tedious fast.
Move token footprint handling from "comment out the irrelevant footprint-types in TokenFootprintCreator, run the class to print out what should be in squareGridFootPrints.xml, cut-and-paste the relevant bits to said file" to "having a gui kind of like Lights, Sight, etc."

**Additional context**
I've added token sizes up to 10x10 on the branch I'll shortly mention in this issue, which covers what I need.
Refactoring TokenFootPrints (perhaps along the lines of the above) sounds like it would be a good thing to do, but is way beyond the time I have available.
Several of the sizes in the current squareGridFootprints.xml didn't exist in the TokenFootprintCreator (the ones smaller than Medium), so I added them while I was at it.

xdy:
Fixes #1621

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1622)

### DIFF
--- a/src/main/java/net/rptools/maptool/tool/TokenFootprintCreator.java
+++ b/src/main/java/net/rptools/maptool/tool/TokenFootprintCreator.java
@@ -24,8 +24,8 @@ import net.rptools.maptool.model.TokenFootprint;
 public class TokenFootprintCreator {
   public static void main(String[] args) {
     // List<TokenFootprint> footprintList = makeHorizHex();
-    List<TokenFootprint> footprintList = makeVertHex();
-    // List<TokenFootprint> footprintList = makeSquare();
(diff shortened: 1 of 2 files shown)


Phergus:
Merged to feature branch for further work.  See #1621

PR #1624, opened 2020-04-19
Fix token editor crash when using translations

### LINKED ISSUE #1613: Theme is not applied at menu bar and menu items
**Describe the bug**

Theme in config directory is not applied at menu bar and menu items.

When I copy  a theme file in `.maptool-rptools/theme/ui/Default.theme` to my user directory (%userprofile%), such items are showed well.

**Screenshots**

Select "Default" theme.  
select-default

Copy "Default" to user directory.  
default-in-user-dir

**MapTool Info**
- Version: 1.6.1
- Install: New

 - OS: Windows
 - Version 10
 - Language: Japanese

Merudo:
- Fix #1613

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1624)


Phergus:
Going to need the same thing for the Hero Lab tab.

Merudo:
I did not enable translation for the tab, because "Hero Lab" is a copyrighted name. Should I have?

Phergus:
I think you mean trademarked but in any case if it isn't exposed to translation then no worries.

I'm iffy on whether it should be translated.  If Hero Lab is localized for Japanese (or other languages) did they keep "Hero Lab" designation or translate it?  As a trademark it would normally be left alone so I guess that it should stay as is.

Issue #1422, opened 2020-03-28
Add Campaign Note Books


cwisniew:
**Is your feature request related to a problem? Please describe.**
Campaign Note Books could be used to store things in the Campaign that currently have no good place to be stored.

Note Books could contain
* Images
* Sounds
* Text based data such as
	* HTML
	* CSS
    * JavaScript
    * Text Notes
    * Macros

This would also provide alternate storage for lib tokens and tables.

Campaigns could have more than one Note Book which can be imported and exported and a Note Book could contain ALL the information required for a frameworks.

bubblobill:
Rather than do this from scratch, can't we just incorporate something like 
- https://joplinapp.org
- https://xournalpp.github.io/
- etc.

Pmofmalasia:
From the description it doesn't seem like this is a literal notebook - it sounds more like what are now called addons, considering the proposed list of contents. Seems like all that's missing from the list currently is tables?

bubblobill:
Hmm.. you're probably correct. Although it bears a lot of similarity to people asking for a distributed note-taking facility.

PR #1665, opened 2020-04-22
Feature note book

cwisniew:
Start of Note Book Functionality, still a long way to go

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1665)


melek:
Dropping a note that this relates to #1422 so I can find it later. So excited for this feature!

PR #1681, opened 2020-04-23
Feature macro editor

JamzTheMan:
Add the ace editor src which is sort of important. 🤕

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1681)


Azhrei:
Why include the **demo** directory?  Are you planning to support Prolog and Erlang as scripting languages for MApTool? ;)

JamzTheMan:
Because I didn't have time to weed out what would be useful. Eventually we would delete it all but I find it useful to have examples when it's time to implement this we want.

PR #1716, opened 2020-04-27
Use rendering hints for zone and portrait rendering 

### LINKED ISSUE #1709: Portrait image rendering is not scaled with sufficient quality
**Describe the bug**

Various places in MapTool set or don't set the rendering hint for interpolation. The PointerTool painting the portrait for example sets no hint, resulting in a badly scaled image. The Handout (AssetViewer) does set Bilinear (only applies when picture is larger than screen size). 

Also considered : the Token rendering itself (ZoneRenderer) is done without appropriate hint (right side, left side with patched in rendering hint)

image

The ImagePanel has a switch - shrink with bilinear, grow with bicubic as of [#691](https://github.com/RPTools/maptool/issues/691)

```
          if (dim.width < image.getWidth(null) || dim.height < image.getHeight(null)) {
            g.setRenderingHint(
                RenderingHints.KEY_INTERPOLATION, RenderingHints.VALUE_INTERPOLATION_BILINEAR);
          } else if (dim.width > image.getWidth(null) || dim.height > image.getHeight(null)) {
            g.setRenderingHint(
                RenderingHints.KEY_INTERPOLATION, RenderingHints.VALUE_INTERPOLATION_BICUBIC);
          }

```

**To Reproduce**

Look at rendered images. 

**Expected behavior**

Use bilinear everywhere consistently, not sure bicubic is worth it anywhere.

**MapTool Info**
- Version: 1.6.1/develop

 - OS: Windows

**Additional context**

SwingUtil has a common antialiasing method/pattern, that should probably be used everywhere. Maybe it should just set up antialiasing and interpolation hint to 'pretty'. There are some spots in code that explicitly turn off antialiasing (sharp lines needed).

nmeier:
Images are more pleasant when zoomed: Token and Portrait image 

Quick fix only, doesn't address any of the spread out rendering hint logic.

I'm surprised noone noticed this. I don't have display scaling on as far as I can see - this seems to be completely due to the graphics2d rendering hints unless I'm missing something.

Obvious concern: is there a machine that get images scaled like this through their imaging pipeline?

Fixes #1709

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1716)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/tool/PointerTool.java
+++ b/src/main/java/net/rptools/maptool/client/tool/PointerTool.java
@@ -1828,6 +1828,8 @@ public class PointerTool extends DefaultTool implements ZoneOverlay {
                         AppStyle.panelTexture.getWidth(),
                         AppStyle.panelTexture.getHeight())));
             statsG.fill(bounds);
+            statsG.setRenderingHint(
+                RenderingHints.KEY_INTERPOLATION, RenderingHints.VALUE_INTERPOLATION_BILINEAR);
             statsG.drawImage(image, bounds.x, bounds.y, imgSize.width, imgSize.height, this);
             AppStyle.miniMapBorder.paintAround(statsG, bounds);
             AppStyle.shadowBorder.paintWithin(statsG, bounds);


Phergus:
This will have a big impact on map performance.  Drop a largish map image (4k x 4k or so) on a background layer and then open up the Performance Data window (Tool -> Collect Performance Data).  Zoom in and out on map.  Note rendering times.   Do the same in a release version.

Merudo:
If there is a performance / quality tradeoff, then we should provide the option to the user.

JamzTheMan:
Wouldn't you use the more performant option to render as you zoom and pan, and then render a final pass after x time using the higher quality?

Of course this would be more terrible to code but better than yet another option and having to choose one or the other.

nmeier:
I'll run some tests on that 4k-x-4k and, say, a hundred tokens

nmeier:
4500x3136 background image token adds 10-129ms to rendering, avg 37, median 34 .

While that's not much on my machine this could be noticeable on a slower machine, certainly if there are two or three of those images. 

Looking at the result of the map I'm trying out there's actually not much noticeable difference, a minor diff in jagged edge.

image

nmeier:
324 tokens, medium size, aligned to grid 
```
with rendering hint   , 10-111ms, 35ms avg, 26ms mean
without rendering hint, 10- 34ms, 13ms avg, 12ms median

```

So overall not much impact for this amount of tokens imo. But then again the small tokens quality is hardly noticeable.

nmeier:
PointerTool's Portrait  painting with/without cubic rendering hint for image scaling 

```
without: 24-98ms, 42ms avg, 40ms median
with   : 29-87ms, 47ms avg, 42ms median

```

So overall I'd venture to say that painting the portrait in higher quality is negligible in terms of performance and renders a better picture. 
Since the AssetVieweDialog (Handout) also renders in higher quality (VALUE_INTERPOLATION_BILINEAR) the same intent of a good looking image would be in the same spirit here.

nmeier:
Suggestion

1) use bilinear for portrait overlay in PointerTool - that picture will look better (especially since it's forced scaled as per user specified 'constrainTo' size). Performance is fine. Same intent as AssetViewer/Handout, same quality.

2) don't apply scaling for zone renderer (background/tokens) unless there's an option to chose high quality rendering from. Performance can be an issue depending on users' machine and irritate fluidity of moving/zooming across map.
Benefit is minimal unless you're pixel peeping at small zoomed tokens ... are small.

Phergus:
Agreed.  Keep the improved scaling for portraits.  For now, leave it out of zone renderer as the minimal improvement in quality isn't worth the performance hit.

Phergus:
@nmeier Revert your change to the ZoneRenderer and push it up.  Thanks.

nmeier:
pushed

PR #1717, opened 2020-04-27
More informative error message

Phergus:
Inform user if a macro set is selected for import instead of a macro.  Issue #1570 .

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1717)


Merudo:
I'm getting double periods in the error message.

Phergus:
Fixed.

Issue #1720, opened 2020-04-28
Campaign token props with default values are not trimmed


derficus:
**Description**
When configuring campaign token properties, property names are trimmed of trailing spaces if they have no default value, or if they have a short name, or a short name and a default value.  In the case where they have no short name, but do have a default value, if there is a space at the end of the property name it will not be trimmed.

**To Reproduce**
(MapTool 1.6.1)
1. Edit -> Campaign Properties, and leave the Token Properties tab up
2. Create a new Token Type by entering the name TestType in the Name field
3. Enter the following lines to configure the properties:
```
testval  
testval2 (t2):5
testval3 :5
endprop
```
4. Hit Update, then OK
5. Run the following macro:  [r: getAllPropertyNames()]
The output should clearly show a space only at the end of testval3, like so:
testval,testval2,testval3 ,endprop

**Expected behavior**
Considering that the property names are trimmed in all other cases, it would be consistent to expect that they should be trimmed in this one as well.

**Screenshots**

**MapTool Info**
- Version: 1.6.1
- Install: Built from source

**Desktop (please complete the following information):**
 - OS: Ubuntu derivative
 - Version: 18.04

**Additional context**
Will submit a pull request with fix.

Fixing this would break existing campaigns if they have properties with trailing spaces and have worked around it by using a trailing space in getProperty() calls, but I can't imagine that's a common enough case (if there are any at all) to warrant leaving it unfixed.

Phergus:
@derficus  PR has an issue but it is an easy fix.

Phergus:
Tested.  No longer allows properties with leading or trailing spaces when a default value and/or short name is provided.

PR #1721, opened 2020-04-28
Fix for #1720 - trim() campaign property names with default values

derficus:
Moved the trim() of the property name outside of the if() that was parsing out the short name.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1721)


derficus:
Woop.. looks like this should probably have gone against the develop branch instead of 1.6-maintenance

PR #1802, opened 2020-05-10
Improve token image handling

### LINKED ISSUE #1801: Token image handling is slow
**Describe the bug**

Switch back and forth between two maps with tokens - notice that switching is slow.

Copy a set of tokens, switch to another map, paste - notice that tokens are placed but background loading images is progressive and slow.

Select a set of tokens, start dragging - notice the initial delay before the drag previews smoothly follows the mouse.

**To Reproduce**

Do any zone switching, copy/pasting, or dragging with 100 or so tokens to really notice the effect.

**Expected behavior**

All these activities should be pretty instant as image data is already loaded. Instead there is image data that is force-uncached without need and temporary images generated in getRGB/setRGB loop.

**Screenshots**

dragging a large number of tokens

image

**MapTool Info**
- Version: 1.6 + 1.7 beta

 - OS: Windows

nmeier:
Keep soft references to loaded images so they don't have to be re-regenerated unless the VM is looking for memory to garbage collect. This makes switching between zones faster.

Don't generate a gray image when dragging tokens - use an alpha composite instead. This improves select/drag performance and uses less memory.
This is a bit more tricky as playing with the Graphics2D state can affect rendering of the next token. There's more opportunity to use a simple transformation for flip/flip-iso which are still being generated.

Fixes #1801

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1802)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
@@ -76,7 +76,6 @@ import javax.swing.JComponent;
 import javax.swing.SwingUtilities;
 import net.rptools.lib.CodeTimer;
 import net.rptools.lib.MD5Key;
-import net.rptools.lib.image.ImageUtil;
 import net.rptools.lib.swing.ImageBorder;
 import net.rptools.lib.swing.ImageLabel;
 import net.rptools.lib.swing.SwingUtil;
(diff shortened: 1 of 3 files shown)


Merudo:
From my tests, it appears that this is causing #2069.

src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
```diff
@@ -3394,20 +3366,19 @@ protected void renderTokens(
...
             cellArea.intersect(cb);
-            cellOnly.setClip(cellArea);
-            cellOnly.drawImage(workImage, at, this);
+            tokenG.setClip(cellArea);
+            tokenG.drawImage(workImage, at, this);
           }
         }
       } else {
-        clippedG.drawImage(workImage, at, this);
+        tokenG.drawImage(workImage, at, this);
```
Merudo, line 3381: This changed line causes #2069.

Issue #1804, opened 2020-05-11
Rework zone (map) selection to make it simpler and behave like a normal popup 


nmeier:
**Describe the bug**

The zone (map) selection component behaves unusual for a swing component and doesn't work well with LnF

* the popup hides the button
* the entries are rendered complicately with a button signalling the current zone

|  ![image](https://user-images.githubusercontent.com/312857/81515218-f8b33a80-9300-11ea-9df1-5775efede214.png)  |  ![image](https://user-images.githubusercontent.com/312857/81515206-eafdb500-9300-11ea-9ae4-973122c3585e.png) |

**To Reproduce**

Click on the zone selection button in top right corner - observe the popup handling and how the items are rendered

**Expected behavior**

Make it a simple menu - checkbox on current zone

![image](https://user-images.githubusercontent.com/312857/81515275-3c0da900-9301-11ea-9d23-a8f7bc7e3d9d.png)

**MapTool Info**
- Version: 1.6/1.7 beta

**Desktop (please complete the following information):**
 - OS: Windows

nmeier:
PR coming

JamzTheMan:
Be sure to test with a lot of maps. It was recently changed to give it a scrollbar.

Phergus:
@nmeier This change doesn't handle more than 20 maps.  Needs the scroll bar.

Here's a campaign file with 50+  maps.

[50PlusMaps156.cmpgn.zip](https://github.com/RPTools/maptool/files/4607515/50PlusMaps156.cmpgn.zip)

Phergus:
Also it does this:
```
      BufferedImage thumb =
          MapTool.takeMapScreenShot(new PlayerView(MapTool.getPlayer().getRole()));
```
But doesn't use it.

bubblobill:
Sounds like an image that should appear on hover that didn't eventuate

nmeier:
yeah, sorry, that was not a good PR. I was playing with a thumbnail but didn't get it to work yet. And scrolling is broken. Looking into it.

Phergus:
Scrolling is fixed and now noticed that the Hidden maps no longer have "highlighting" to indicate they are hidden.

I know.  Never happy.  ;)

nmeier:
That's fine - I have to catch up what's possible in MT - hidden maps, can indicate that. checking how that looked in 1.6

nmeier:
[...]

PR #1810, opened 2020-05-11
Bug Fixes for 1801 and 1804

nmeier:
Token image handling is slow, doesn't break rendering of combination halo/state/facing now #1801
Rework zone (map) selection that doesn't break scrolling #1804

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1810)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/JScrollPopupMenu.java
+++ b/src/main/java/net/rptools/maptool/client/ui/JScrollPopupMenu.java
@@ -141,6 +141,8 @@ public class JScrollPopupMenu extends JPopupMenu {
       int height = heightMargin + extent;
 
       setPopupSize(new Dimension(width, height));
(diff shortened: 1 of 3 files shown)


nmeier:
Added on tiny change to revert back popup menu's preferred size - so there's no scrollbar space reserved after it goes way when number of map goes back to <=20

JamzTheMan:
Shouldn't it be based on screen size and not number of maps? What screen scaling and resolution only allow 15 maps?

nmeier:
@JamzTheMan 20 was the old limit, you're right though, one could argue that something like half the screen should be used for that popup,

JamzTheMan:
Gotcha.

Phergus:
Wouldn't want more than that.  Too much scrolling when the mouse-wheel is faster.

PR #1815, opened 2020-05-12
Fix transferVBL not being permitted to run if called by GM in chat

### LINKED ISSUE #1814: Some functions using incorrect test for trusted status causes failure when called by GM in chat
**Describe the bug**
If the GM calls a macro from the chat, the macro path should be considered trusted.

However, the `macroPathTrusted` parser property property is set to false regardless of who typed the macro.

**To Reproduce**
1. Identify a macro checking permission through `isMacroPathTrusted`, such as `transferVBL()`.
2. Run the macro through chat as a GM.
3. You get the error: `You do not have permission to call the "transferVBL" function.` 

**Expected behavior**
A GM should have permission to run all macros from chat.

**MapTool Info**
- Version: 1.6.1
- Install: New

 - OS: Windows
 - Version: 10

**Additional context**
It's not clear to me why some macros verify permission through `isMacroPathTrusted()`, which other macros verify permission through `isMacroTrusted()`.

Maybe we should stick to one permission method?

Merudo:
- Change macroPathTrusted to depends on GM status if there is no context (accessed through chat)
- Close #1814

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1815)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
+++ b/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
@@ -2058,11 +2058,12 @@ public class MapToolLineParser {
    */
   public void enterContext(MapToolMacroContext context) {
     // First time through set our trusted path to same as first context.
-    // Any subsequent trips through we only change trusted path if conext
+    // Any subsequent trips through we only change trusted path if context
     // is not trusted (if context == null on subsequent calls we dont change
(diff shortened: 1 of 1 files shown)


cwisniew:
@Merudo you were too quick :) I was unable to comment on the issue before this fix was submitted, please see issue.

Merudo:
Thank you for your comments.

Issue #1669, opened 2020-04-22
Add Overlay Helper Functions [1.7a]


melek:
**Is your feature request related to a problem? Please describe.**
The new overlay roll option is very exciting, but lacks important helper functions that frames and dialogs enjoy.

**Describe the solution you'd like**
I would like a few functions:
* `isOverlayRegistered(overlayTitle)` - Returns 1 or 0. (Checked status in Window -> Overlays menu.)
* `closeOverlay(overlayTitle)` - Closes named overlay if open & returns 1. Returns 0 if the overlay wasn't found but doesn't error
* `get / setOverlayProperties` - Returns / sets any properties, especially z-order.
  `{"name":"Overlay","zorder":0,"visible":1}`

**Describe alternatives you've considered**
Just leaving overlays open all the time, because I can't close them.

**Additional context**
`isOverlayVisible` would be the consistent name for the first proposed function, but since users can toggle 'Visibility' of overlays it might be confusing.

Phergus:
Just FYI, you can close overlays, just pass an empty content block:  `[overlay("Overlay"):{}]`

But having the convenience methods wouldn't be a bad thing.

melek:
In the interrim knowing that I cam close overlaps is helpful, thank you :)

Merudo:
PR #1826 adds isOverlayRegistered, closeOverlay, getOverlayProperties

Phergus:
This seems to have broken something.

With 1.7.0b2 I can open two overlays at the same time with different titles and then close and reopen them with no issue.

With current develop, if I open two overlays they see to get bound together.  Only the 2nd one is "registered"  and closing it closes both.  Calling either macro opens the combination of them once the both have been opened.

Merudo:
Do both overlay have the same Z order?

Phergus:
I assume so since I didn't specify.

Merudo:
Might be a java bug. `ConcurrentSkipListSet` should be able to two elements such that `o.compareTo(e) == 0` as long as `o.equals(e) == false`.

Yet, it doesn't seem to be the case.

Merudo:
Well the information from the javadocs is seriously misleading. Apparently `o.compareTo(e) == 0` must imply `o.equals(e)` for the contract of the SortedSet interface to be fulfilled. 

You won't find this in any of the documention of concurrentskiplistset, though.

https://stackoverflow.com/questions/48101005/concurrentskiplistset-using-comparator-will-not-add-new-unique-values

[...]

PR #1826, opened 2020-05-14
Add functions isOverlayRegistered, closeOverlay, getOverlayProperties

Merudo:
- Add overlay helper functions isOverlayRegistered, closeOverlay, getOverlayProperties
- Functions described in #1669

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1826)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/MacroDialogFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/MacroDialogFunctions.java
@@ -18,8 +18,7 @@ import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
 import java.math.BigDecimal;
 import java.util.List;
-import java.util.Map;
-import javafx.application.Platform;
+import java.util.concurrent.ConcurrentSkipListSet;
 import net.rptools.maptool.client.MapTool;
 import net.rptools.maptool.client.ui.htmlframe.HTMLDialog;
 import net.rptools.maptool.client.ui.htmlframe.HTMLFrame;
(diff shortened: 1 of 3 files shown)


cwisniew:
In HTMLOverlayPanel

- Its probably better to use ConcurrentHashMap rather than Collections.synchronizedMap(), in fact we should probably replace synchronizedMap() where we can when we come across it (where we can being anywhere that is not serialized as that would break save compatibility).
- if the return of getOverlays() is not intended to be modified it should return a copy of / read only version of overlays. If it is its probably even more important its a ConcurrentHashMap.
- Can you also make the MapUtil inner class private too, it should probably be replaced using streams at some point but until then its best to make sure it doesn't find its way to being used elsewhere (it will be weird to see HTMLOverlayPanel.MapUtil.sortByValue() popping up in code base in areas that have nothing to do with overlays)

Phergus:
@Merudo holding pending action on Craig's comments.

Merudo:
Thank you for your comments, @cwisniew 

1) The reason I used a `Collections.synchronizedMap` instead of a `ConcurrentHashMap` is that overlays are kept in an ordering corresponding to the z order of the overlays. `ConcurrentHashMap` does not support such ordering, but `Collections.synchronizedMap` together with `LinkedHashMap` do.

2) Agreed.

3) Sure.

Merudo:
Instead of a ConcurrentHashMap I could use a ConcurrentSkipListMap.

cwisniew:
> Instead of a ConcurrentHashMap I could use a ConcurrentSkipListMap.
Yep if you want sorted then this is probably way to go.

Merudo:
The added commit add these changes:

- Replace overlay collection LinkedHashMap with ConcurrentSkipListSet
- Change getOverlays to return a clone instead of the collection
- Remove MapUtil

melek:
All three functions added to the wiki and can be seen here:
https://lmwcs.com/rptools/wiki/Category:Overlay_Function

I left all with the 'stub' template until they are reviewed/improved.

Issue #1838, opened 2020-05-15
Fix Show as Player so GM can reliably see what players see


Merudo:
**Describe the bug**
"Show as Player" for GMs is notoriously unreliable, and what is displayed does not match what players see.

For example, the Initiative Panel and Map Explorer do not currently take into account "Show as Player".

**Expected behavior**
The "Show as Player" show the GM a view corresponding to what players see. More precisely, it should behave as if all tokens with an owner also had the GM as an owner.

Alternatively, the "Show as Player" menu could be replaced with a submenu listing all players. The GM could then select which player(s) to view the map as.

**MapTool Info**
- Version: 1.6.1, 1.7.0 beta, develop
- Install: New

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: 10

Merudo:
PR #1839 adds the functions `MapTool.getPlayer().isEffectiveGM()` and `MapTool.getPlayer().getEffectiveRole()`, which will make it easier to manage code so that it takes `Show as Player` into account.

Merudo:
PR #1840 fixes this for the Map Explorer.

Merudo:
PR #1841 fixes it for the Initiative Panel.

Phergus:
Show as Player was only ever about the Map view and not an actual role swap.  What is the purpose of having the Initiative and Map Explorer panels conceal information from the GM?

Merudo:
By checking "Show as Player", the GM could quickly see what tokens are visible in the Initiative and Map Explorer panels. This information would require several properties to be checked for each token otherwise (token could be invisible, on the token layer, not owned, etc).

Map Explorer would quickly lists which PC tokens are visible to players on the map.

Initiative Panel would let the GM know if some tokens are not visible in the initiative for some reason. Maybe they are on the wrong layer or something.

My understand is that the "Show as Player" option is used for the DM to quickly check if something is not displayed correctly.

Why not have that be the behavior for the Map Explorer and Initiative Panel? To me, that's more useful than showing information the GM can already see in his usual GM view.

Phergus:
Found the conversation on Discord.

Map Explorer already shows what is and isn't flagged as Visible to Players.  As does the Initiative panel.  The Map Explorer sorts first by Visible and Not Visible and then by name within those categories.  Initiative panel changes the shading.

> My understand is that the "Show as Player" option is used for the GM to quickly check if something is not displayed correctly.

[...]

PR #1841, opened 2020-05-15
Fix Show As Player not working for Initiative Panel

Merudo:
- Part of #1838

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1841)


Merudo:
Show as Player should not apply to initiative panel.

Issue #1847, opened 2020-05-15
On macOS, restore horizontal scrolling while still fixing the bugs found in #1317


mrkwnzl:
This is a follow-up to #1317. Also pinging @Azhrei 

I experimented around and found a way to restore horizontal scrolling on macOS, while still avoiding the conflicting behavior while zooming or editing Token Layout (Edit Token > Config). My solution is to check if on macOS and if shift is pressed, then ignore only on problematic cases, instead of globally. 

It’s still not possible to use shift + mouse wheel to set the facing.

It certainly needs a bit more testing, especially on Windows and Linux. I *think* my changes only affect macOS, but I’m not entirely sure.

Phergus:
No impact seen on Windows.  Facing, zooming, config zoom all working as before.

mrkwnzl:
Sorry for the back and forth, but after some more rigorous testing, I found some more problematic cases (frame5/dialog5 and scrolling in windows with _only_ horizontal scrollbars), which I cannot easily solve with my method above.

With the final release of 1.7.0 being close, I’d prefer our previous solution of ignoring all horizontal scroll events, as that makes MapTool less glitchy, at least until I worked out those kinks, too. 

I have for now reverted the changes and issued a PR in the hopes that you agree that a less glitchy release would be preferred.

Azhrei:
There is code deep in the peer UI classes for Swing that applies scroll events to whichever scroll bar is visible if there’s only one.  That’s likely what you’re seeing in regards to what happens when only the h.scroll is visible.  Just FYI.

mrkwnzl:
Thank, @Azhrei, I’ll take a look at that, but I’ll need more time and testing for that. I didn’t realize that the final 1.7.0 release was so close when I issued #1848, otherwise I would have waited for a later release with that.

Phergus:
Just checking but you didn't expect reverting the change would fix the horizontal scrolling in frame5 for Windows (#1862 )?

mrkwnzl:
No, #1842 must have a different root cause. I suspect is has something to do with scrolling in frame5/dialog5 being handled by JavaFX, while the rest of scrolling in MapTool is handled by Swing.

Phergus:
That's what I thought.

PR #1848, opened 2020-05-15
Restore horizontal scrolling on macOS

mrkwnzl:
Discussion and description in #1847

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1848)


Azhrei:
So why do some pieces of code use `SwingUtil.isShiftDown(e)` and others use `e.isShiftDown()`?  If you're in there changing things, we should probably consolidate.  And perhaps we should use the `SwingUtil` approach since that delegates event specifics to a separate class which means global updates could be done in one place in the future.

Phergus:
That sounds like a good idea to me and its a good time to do it.

mrkwnzl:
According to the search in IntelliJ, my instances were the only ones. And as I just learned, GitHub updated this already.

PR #1854, opened 2020-05-16
Fixes incorrect ordering of CSS when the CSS is loaded from a macro button (eg CSS@lib:css)

### LINKED ISSUE #1853: Frame5/Dialog5/Overlay load CSS in incorrect order when link points to  macro
**Describe the bug**
When CSS is loaded from a macro via the link tag it is inserted at the top of the head element before anything else.

**To Reproduce**
Campaign Attached which demonstrates the error.

**Expected behavior**
CSS should be loaded in the order it is defined in the link tag.

**MapTool Info**
1.7 Alpha
[CSSPrecedenceError.zip](https://github.com/RPTools/maptool/files/4637990/CSSPrecedenceError.zip)

cwisniew:
Does what it says on the tin
closes #1853

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1854)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLWebViewManager.java
+++ b/src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLWebViewManager.java
@@ -358,7 +358,7 @@ public class HTMLWebViewManager {
     // Deal with CSS and events of <link>.
     nodeList = doc.getElementsByTagName("link");
     for (int i = 0; i < nodeList.getLength(); i++) {
-      fixLink(nodeList.item(i).getAttributes(), nodeCSS, doc);
+      fixLink(nodeList.item(i), nodeCSS, doc);
     }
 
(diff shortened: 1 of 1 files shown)


cwisniew:
#1853 contains attached campaign demonstrating the error

Merudo:
Won't this cause issues if more than one link is added to the page?

Merudo:
Nice! Cleaner logic too than what was there before.

Two things:

1) Parameter `refNode` is now useless and should be removed
2) `System.out.println(html);` prints the whole document to the console. I assume this was for testing, but it should be removed

cwisniew:
yeah I need to remove that output oops!
But just noticed there is a bug still when there are multiple CSS from macro they are in reverse order.

cwisniew:
pushed fix for reverse ordering of ccs@lib:token
also addressed above feedback.

PR #1875, opened 2020-05-18
Fix incorrect macroprop "toolapplyToSelected"

### LINKED ISSUE #1874: getMacroProps returns misnamed property "toolapplyToSelected"
**Describe the bug**
The json returned by `getMacroProps` returns the property `toolapplyToSelected`.

The property name should be `applyToSelected` instead.

**To Reproduce**
1. Use `getMacroProps` on a macro.
2. Notice the included property `toolapplyToSelected` and the missing `appyToSelected`

**Expected behavior**

The property is named  `appyToSelected`

**MapTool Info**
- Version: 1.7.0
- Install: New

 - OS: Windows
 - Version: 10

Merudo:
- Fix getMacroProps returning properties with "toolapplyToSelected" and "tooltooltip" instead of "applyToSelected" and "tooltip"
- Fix #1874

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1875)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/MacroFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/MacroFunctions.java
@@ -170,12 +170,8 @@ public class MacroFunctions extends AbstractFunction {
       props.addProperty("playerEditable", mbp.getAllowPlayerEdits());
       props.addProperty("command", mbp.getCommand());
       props.addProperty("maxWith", mbp.getMaxWidth());
-      if (mbp.getToolTip() != null) {
-        props.addProperty("tooltip", mbp.getToolTip());
-      } else {
-        props.addProperty("tooltooltip", "");
-      }
-      props.addProperty("toolapplyToSelected", mbp.getApplyToTokens());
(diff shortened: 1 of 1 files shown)


src/main/java/net/rptools/maptool/client/functions/MacroFunctions.java
```diff
@@ -175,7 +175,7 @@ public Object getMacroButtonProps(Token token, int index, String delim) throws P
       } else {
         props.addProperty("tooltooltip", "");
```
selquest, line 176: Looks like this will also create a bad property name "tooltooltip" in cases of missing tooltip.

Merudo: Good catch!

Phergus:
@selquest   You have a blocking comment that needs resolution.

Merudo:
I'm sorry@selquest - I didn't mean to dismiss you from the discussion, I was just trying to mark your comment as resolved since I did implement your very helpful suggestion :).

selquest:
Ahh, sorry - didn't realize I was holding things up with that.  No worries!

PR #1877, opened 2020-05-18
Fix quote symbols created by [if(0), code:{}]

### LINKED ISSUE #1876: Unwanted quote symbols created by [if(0), code:{}]
**Describe the bug**

The code

```
[if (condition), CODE:{ some code }]
```
Will print a unwanted `''` as output if the condition is equal to zero (false).

The current workaround is to add an empty `{}` for the other condition,

```
[if (condition), CODE:{ some code };{}]
```

**To Reproduce**
1. Type `[if(0), code:{}]`
2. Notice the unwanted symbol `''`.

**Expected behavior**
No `''` if no condition is evaluated.

**MapTool Info**
- Version: 1.5.0 to 1.7, probably earlier as well
- Install: New

 - OS: Windows
 - Version: 10

Merudo:
- Fix unwanted quote symbols been produced as output when using the [if] condition together with the CODE argument when the condition is false and no second branch is defined
- Fix #1876

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1877)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
+++ b/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
@@ -1257,8 +1257,10 @@ public class MapToolLineParser {
                   Matcher testMatcher = Pattern.compile(testRegex).matcher(roll);
                   if (testMatcher.find()) { // verifies that roll body is well-formed
                     rollBranch = testMatcher.group(1 + whichBranch);
-                    if (rollBranch == null)
-                      rollBranch = "''"; // quick-and-dirty way to get no output
+                    if (rollBranch == null) {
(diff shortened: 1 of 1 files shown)


melek:
The end of an era.

Phergus:
That was my that as well @melek.  lol  My fingers are already programmed to just put that second set of braces in.

Merudo:
I'm wondering if any of our veterans will benefit from this. They'll probably just add the `;{}` by force of habit anyway. :wink:

JamzTheMan:
Ha!  I've always refused to add them out of spite! I'll throw all output away and use broadcast before input them in!

So thanks! Im sure I missed a few places...

Issue #1898, opened 2020-05-21
Improve Macro performance


nmeier:
**Describe the bug**

Macros have a couple of areas where they could be improved

1. low level usage of lesser than optimal APIs (e.g. trim() where isBlank can be used) creating more temporary Strings - while that's not a big deal in modern Java, repeated ops like that can slow things down

2. Parser/Expression re-creation on every macro line, even in a loop. One could re-use ExpressionParser/Expressions

3. slower than optimal String functions as every string operation causes the full list to be parsed and converted into ArrayList<String> first, Cost: more String instantiations (minor) but more importantly doesn't allow for optimization (e.g. listGet(0) which should be O(1) but is O(n))

4. Even if a script uses GUIDs (getTokens() rather than getTokenNames() the lookup in Zone.resolveToken() does an expensive case insensitive check for all token names first, before trying to interpret as GUID. That can be made faster.

5. json Arrays are flattened/reparsed unnecessarily between parser lines (in both maptool and parser)

**To Reproduce**

Hard to measure. I've run performance tests at scale (over 500 tokens, with 500 elements) to work around variances in function execution. My test macros (will attach) for sufficiently large sets of tokens to iterate over improved by 3-5x.

I've also run specific function tests at scale which are 10% faster for worst-case and 5-10x faster for delete(), find(), get(), append() for optimal indexes/simple appending.

**Expected behavior**

Faster macros 

**Screenshots**

n/a

**MapTool Info**
- Version: 1.7
- Install: dev

**Desktop (please complete the following information):**
 - OS: Windows

**Additional context**

Adding macro examples with PR. I'll prepare PRs for different areas 1-4

Merudo:
I'm not sure if I fully understand what you mean by 3a and 3b, but I think 1, 2, and 4 should definitely be implemented.

cwisniew:
I think 3a/3b boil down to don't use String lists :)
The only internal representation for them is Strings. 
Even though json has a parsing cost for each fetch from a property its at least a one off, rather than the pathological case for looping through a string list. 

I have plans for notebooks to reduce that parsing for json to once per campaign, actually we could probably do that for token properties too. But I have no such plans for String lists because well people expect them to be Strings so how else would you store them?

nmeier:
Tried to clarify 3a a bit - if you call stringGet(i) for all elements in a list, which is a string with delimiter in a macro, the string is parsed n times into a List<String> to then do a O(1) access for that element.

[...]

PR #1920, opened 2020-05-23
Macro Performance - improve strfunctions (3) and fix jsonarrays being flattened unnecessarily (5)

nmeier:
The string parsing doesn't generate a fully parsed temporary string-list. Functions like append do the appending in one step. Functions that don't have to loop the full list (e.g. get(0)/remove(3)) don't walk the full list.

With sufficient large lists I've seen 3-5x perf improvement.

Added unit test class.

Important to consider as discussed: any pretty formatted string (VALUE+SEPARATOR+SPACE+VALUE+SEPARATOR+SPACE+...) stays pretty formatted. Any non-pretty string (e.g. "a,b, c,d") is not guaranteed to be reformatted (i.e. "a, b, c, d").

Suggestion for #1898

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1920)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
+++ b/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
@@ -973,8 +973,7 @@ public class MapToolLineParser {
 
                     // If we still dont have a list treat it list a string list
                     if (foreachList == null) {
-                      foreachList = new ArrayList<String>();
-                      StrListFunctions.parse(listString, foreachList, listDelim);
+                      foreachList = StrListFunctions.toList(listString, listDelim);
                     }
                     loopCount = foreachList.size();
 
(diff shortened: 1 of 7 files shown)


cwisniew:
@nmeier please see comment I left in the issue. I much prefer a solution that addresses string lists / prop / json faster. Other wise when we fix json then string list / props will get left out of those performance benefits (I.e. I won't be rolling back changes here to reimplement)

nmeier:
Yeah, this makes string stlist functions faster. It doesn't address json or any possible json improvements. Matter of priorities - fix obvious performance problems with what is there or go for something more holistic. 
I'd still venture to say that this change gets you immediate performance improvement for no real con. Would be great if someone who has slow macros could try it out.

cwisniew:
@nmeier can you resubmit this seems to be out of sync with develop

nmeier:
updated

nmeier:
started work on the flattening of json arrays - this requires a change in https://github.com/RPTools/dicelib/issues/43

WIth a nice dicelib version that allows to optionally skip determinism, I can tune MapTool json array to trim 60% of runtime for my macro append test.

PR #31, opened 2015-11-08
Add setup documentation for option UI tools

joshsziegler:
This documentation is by Jagged from this forum thread:
http://forums.rptools.net/viewtopic.php?f=86&t=25838

I'm not sure if this documentation is up to date.

PR #1942, opened 2020-05-27
Update parser to 1.7.1 to enable ISO-8859-1 characters

Merudo:
- Enable ISO-8859-1 characters, both in strings and in variable names
- Discussed in RPTools/parser#30

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1942)


Merudo:
Characters now supported:

1920px-Latin-1-infobox svg


The ones after the `......................................` are added by PR #31.

The last two rows can be used in variable names, except for the × and ÷ symbols.

Issue #1938, opened 2020-05-26
Macros are not set to be Player Editable when they should be


Merudo:
**Describe the bug**
Some macros are not set to be Player Editable when they should be:

- Macros created through the `createMacro()` command
- Macros created by dropping a token from outside MapTool

**To Reproduce**
1. As a player, create a macro through one of these two methods
2. Notice the macro has the Player Editable field set to `false`.

**Expected behavior**
The Player Editable field is set to `true`.

**MapTool Info**
- Version: 1.7, develop
- Install: New

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: 10

Merudo:
PR #1939 fixes this.

Phergus:
Testing current develop.  Executing this:
`[createMacro("Test", "this is a test", "autoExecute=true;playerEditable=0;color=green", ";") ]`
on a logged on player client produces this error:
```
java.lang.NullPointerException: Cannot invoke "net.rptools.maptool.model.Token.getName()" because the return value of "net.rptools.maptool.model.MacroButtonProperties.getToken()" is null error executing expression createMacro("Test", "this is a test", "autoExecute=true;playerEditable=0;color=green", ";") .
```
Without the `playerEditable` option it works as expected.  Error should probably be trapped and a message to the effect of "Setting state of **playerEditable** is not allowed from player clients."

Other wise, player created macros (using `createMacro` or dropped in tokens with macros) are set as Editable by Players and access to the checkbox is disable on player clients.

Merudo:
The ParserException error is itself throwing an exception :(. Not my code but I will fix.

Phergus:
Tested.  Attempting to create a macro with either `playerEditable` or `command` in the `props` argument fails with message that user lacks permission.

PR #1957, opened 2020-05-29
Fix createMacro() throwing NPE if playerEditable is specified

Merudo:
- Fix createMacro() throwing a null pointer exception if playerEditable is among the properties defined and the player is not a GM
- Issue discussed in #1938

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1957)


Merudo:
I'm not sure what's wrong with the macOS build.

Phergus:
Dunno.  I restarted once but it got hung up again on the MacOS build.

PR #1968, opened 2020-06-01
Use new option in dicelib to only evaluate with a deterministic outco…

nmeier:
…me IF the detailed expression is actually used (i.e., TOOLTIP, expandRoll). That turns off the whole deterministic AST building in dicelib when not needed. It also avoids the associated flattening of variables (e.g., json array) when all that counts is the result, e.g. in a macro loop.

This still works

```
/roll 1d6 + 4
* nmeier rolls: 1d6 + 4 => 5 + 4 = 9

```
With this change the example macro in #1898 that appends 10k elements to a json.array does that in a loop as true list appends now.

This reduces the runtime of that example macro from ~11-12s to 4-5s on my machine.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1968)


nmeier:
This is going to need a bump in dicelib and version dependency. I'm building locally against dicelib develop.

PR #1990, opened 2020-06-07
[lib/i18n] bundle.getString can't return null

grimreaper:
Problem

bundle.getString throws an NPE instead of returning null if string can't
be found. However, we have a null check.

Solution

Remove null check. This then trivially tranforms into a more
functional-like approach.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1990)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,6 +1,9 @@
 ---
 name: Bug report
 about: Create a report to help us improve
+title: ''
+labels: ''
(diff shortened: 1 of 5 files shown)


Phergus:
Couple problems:
- There is no issue open on this.
- It was done against `master` instead of `develop`

https://github.com/RPTools/maptool/wiki/Submitting-Pull-Requests

grimreaper:
Thanks for the info. I'll read through the guidelines now. Apologies for not doing so in advance. Same for my other one.

PR #1993, opened 2020-06-08
[lib/i18n] bundle.getString can't return null

### LINKED ISSUE #1992: bundle.getString is checked for null
In `net.rptools.lib.i18n.I18NManager#getText` bundle.getString is checked for null. However, it is not legal for this function to return null and in fact throws a `MissingResourceException` if the resource does not exist.

In a search across the rest of the code bast this is the only instance of this issue since `net.rptools.maptool.language.I18N#getString(java.lang.String)` translates MRE to `null`.

grimreaper:
Problem

bundle.getString throws an NPE instead of returning null if string can't
be found. However, we have a null check.

Further, this class is a utility class but is not idiomatically marked as so

Solution

Remove null check. This then trivially tranforms into a more
functional-like approach.

Mark the class as final with a private ctor (and private fields).

fixes #1992

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1993)

### DIFF
--- a/src/main/java/net/rptools/lib/i18n/I18NManager.java
+++ b/src/main/java/net/rptools/lib/i18n/I18NManager.java
@@ -20,21 +20,23 @@ import java.util.Locale;
 import java.util.ResourceBundle;
 import java.util.concurrent.CopyOnWriteArrayList;
 
-public class I18NManager {
+public final class I18NManager {
 
-  private static List<String> bundleNameList = new CopyOnWriteArrayList<String>();
-  private static List<ResourceBundle> bundleList = new ArrayList<ResourceBundle>();
+  private static final List<String> BUNDLE_NAME_LIST = new CopyOnWriteArrayList<String>();
+  private static final List<ResourceBundle> BUNDLE_LIST = new ArrayList<ResourceBundle>();
   private static Locale locale = Locale.US;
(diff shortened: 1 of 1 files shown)


Merudo:
What's the reason for making the class final?

cwisniew:
> What's the reason for making the class final?
> 
> Instead, I would make `bundleNameList`, `bundleList` and `localeListenerList` final.

The class is not intended to be instantiated nor subclassed, so I am more than happy for there to be a private constructor and for the class to be marked final. 
But I would also prefer that the above class fields also be marked final too, at least that way anyone adding to the class knows they should not be modifying them

Phergus:
@grimreaper   See Craig's comments above.

grimreaper:
Sounds good. I'll update the diff "soon" and respond to all comments. I've been busy and trying to learn the norms of this community so please have some patience

Phergus:
@grimreaper No problem.

grimreaper:
As mentioned above, the reason to mark the class final is this is how one usually marks a "utility class": made final with a private constructor. Its idiomatic but certainly not required.

I will mark the fields final too.

grimreaper:
I renamed the fields to call capitals because they are now `static final`

Issue #1981, opened 2020-06-04
Update MapTool to Build / Run on JDK 14


cwisniew:
This is a [ZenHub](https://www.zenhub.com/) Epic

**Summary of the Epic.**
Current MapTool build runs on Java 10 and requires an older Oracle JDK to build.

**Describe the solution**
See attached issues (via [ZenHub](https://www.zenhub.com/) or plug in)for further information

**Additional context**
See attached issues (via [ZenHub](https://www.zenhub.com/) or plug in)for further information

PR #2035, opened 2020-06-20
Fix JVM cfg file loading

JamzTheMan:
Update JVM user preferences so it can load cfg file with dynamic filename based on release type.
Add special logic handling for -XX: type jvm parameters.
Updated error logging

Continued work against #1981

FYI: the cfg file reader/writer uses an apache config for INI files which is closest match to the jpackager cfg file. It can't handle parameters that do not match the normal `key=value` notations so special logic has to be added for any JVM parameters like `XX:+ShowCodeDetailsInExceptionMessages`

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2035)

### DIFF
--- a/.github/workflows/verify-build.yml
+++ b/.github/workflows/verify-build.yml
@@ -17,6 +17,16 @@ on:
       - feature*
       - release*
 jobs:
+  formatting:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-java@v1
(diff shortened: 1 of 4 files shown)


Merudo:
You forgot about spotless.

JamzTheMan:
I was um testing the be GitHub actions. Yea. That's the ticket...

JamzTheMan:
fixed. i really need to automate that somehow on my end...

PR #2050, opened 2020-06-22
Remove unused classes

### LINKED ISSUE #2047: remove java version checks
**Describe the bug**
There is a function verifyJavaVersion which checks to see if we're running on Java 6+ but
1) We provide our JRE/JDK
2) We actually require Java 14+
3) The version check is actually ended in the class itself (java won't run invalid versions)

Similarly we use reflection in AssetGroup to permit multiple java versions when it is now unnecessary.

**Expected behavior**

Remove said code

grimreaper:
Problem

There is a function verifyJavaVersion which checks to see if we're running on Java 6+ but

    We provide our JRE/JDK
    We actually require Java 14+
    The version check is actually ended in the class itself (java won't run invalid versions)

Solution:

Remove said code

Fixes #2047

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2050)


Azhrei:
I don't recognize any of those classes.  If others are fine with removing them, go for it.

grimreaper:
something went weird with this commit/message. will recreate.

grimreaper:
Replaced with #2054

PR #2062, opened 2020-07-01
Optimize for performance by only using deterministic parser where needed

### LINKED ISSUE #1898: Improve Macro performance

nmeier:
(TOOLTIP, expandRoll). This makes headless macros (those that work with h and r roll options) faster as no deterministic AST tree is generated within parser.

Added unit tests for macro branch roll options that previously failed.

Unit testing MapToolLineParser is difficult as it's not truly headless. There are calls to static methods that fail, so I hardened a bit where needed.

fixes #1898

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2062)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/MapTool.java
+++ b/src/main/java/net/rptools/maptool/client/MapTool.java
@@ -1382,7 +1382,7 @@ public class MapTool {
   }
 
   public static boolean useToolTipsForUnformatedRolls() {
-    if (isPersonalServer()) {
+    if (isPersonalServer() || getServerPolicy() == null) {
       return AppPreferences.getUseToolTipForInlineRoll();
     } else {
       return getServerPolicy().getUseToolTipsForDefaultRollFormat();
--- a/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
+++ b/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
@@ -631,7 +631,7 @@ public class MapToolLineParser {
       Object retval = params[index];
       // No parsing is done if the param isn't a String (e.g. it's already a BigDecimal)
       if (params[index] instanceof String) {
-        Result result = parseExpression(res, tokenInContext, (String) params[index], true);
+        Result result = parseExpression(res, tokenInContext, (String) params[index], false);
         retval = result.getValue();
       }
       return retval;
@@ -737,7 +737,6 @@ public class MapToolLineParser {
   public String parseLine(
       MapToolVariableResolver res, Token tokenInContext, String line, MapToolMacroContext context)
       throws ParserException {
-
     // copy previous rolls and clear out for new rolls.
     if (parserRecurseDepth == 0 && macroRecurseDepth == 0) {
       lastRolled.clear();
@@ -854,7 +853,7 @@ public class MapToolLineParser {
                   outputOpts.add("w");
                   for (int i = 0; i < option.getParamCount(); i++) {
                     String arg =
-                        parseExpression(resolver, tokenInContext, option.getStringParam(i), true)
+                        parseExpression(resolver, tokenInContext, option.getStringParam(i), false)
                             .getValue()
                             .toString();
                     if (arg.trim().startsWith("[")) {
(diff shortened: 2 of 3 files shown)


Merudo:
This gives my framework errors:

`Illegal argument type com.google.gson.JsonObject, expecting java.math.BigDecimal or java.lang.String`

nmeier:
@Merudo checking your test on illegal argument

nmeier:
confirmed - since json arrays are not flattened all the time the addition doesn't handle the + well, I'll add a test-case. You're workaround to convert to string is right but I think we can make this backwards compatible

`[h: "test" + json.set("{}", "property", "value")]
`

Merudo:
It's because the parser was automatically converting json arrays and objects in variable to strings, and it isn't anymore.

Example:

```json
[h: jobj = json.set("{}", "property", "value")] 
[r: "Json Object: " + jobj] 
```
This returns `Json Object: {"property":"value"}` in 1.7 but gives an error in develop.

Which may be fine, really. The workaround is just to add `string()` strategically.

Merudo:
> confirmed - since json arrays are not flattened all the time the addition doesn't handle the + well, I'll add a test-case. You're workaround to convert to string is right but I think we can make this backwards compatible
> 
> `[h: "test" + json.set("{}", "property", "value")] `

Yep, jsons should be converted to strings when used with the `+` operator, otherwise the operation is not meaningful.

nmeier:
Yeah, it does makes sense to require conversion, otherwise most of functions don't make sense. 

Not good for those macro scripts using any of those functions below at this point

        net.rptools.maptool.client.functions 
            InputFunction  
            ReturnFunction  
            StrListFunctions  
            StrPropFunctions 
        net.rptools.parser.function 
            AbstractComparisonFunction 
            AbstractLogicalOperatorFunction 
            AbstractNumberFunction 
        net.rptools.parser.function.impl 
            Addition 
            Equals  (equals could do an equals on a json array, requires another parser update though)
            Eval 
            Mean 
            Median 
            NotEquals  (same, is possible to do on json array)
            StrEquals 
            StrNotEquals

Phergus:
@nmeier should we expect a fix for this or should it be reverted?

selquest:
My dev environment is failing the unit test starting at this merge commit - `MapToolLineParserTest.testMacros()`:
```
expected line #1:`expanded roll sh[...]` doesn't match ==> expected: <
expanded roll shows ...if\(1 == 1, 1, 0\) = 1.
> but was: <
expanded roll shows if(1 == 1, 1, 0) = 11
>
```

nmeier:
I think we should leave as is - making these operations backwards comp we can rule out for operations between string and the (non flattened) json objects. I think one can argue about (not)equals but the code is just full of assumptions about string/bigdecimal so I'm not sure it's worth it.

nmeier:
@selquest looking at that test-case, it works here on Windows. The special characters that maptool adds for expanded Rolls might be different in your setup.

Merudo:
I think it would make sense for the Addition functions to automatically convert json elements into strings. This should be a straightforward modification to the parser that would preserve backward compatibility.

nmeier:
added a PR that makes those operations (+, ==, etc.) work backwards so this works fine again

`[r: "test" + json.set("{}", "property", "value")]
`

https://github.com/RPTools/parser/pull/42

Phergus:
Parser bumped to 1.8.1.

Issue #1726, opened 2020-04-29
add lumens support to Sight (Light already has it)


Azhrei:
**Describe the bug**
Lights are currently capable of expressing "negative lumens", causing darkness.  There's a note in the Help text of the Campaign Property's **Light** tab that says the dark vision **Sight** is equivalent to 100 lumens.  The problem is that I have a creature that can cast darkness at lumens=-100 (per the **Light** tab) but needs to be able to see through it and can't (because the **Sight** tab doesn't have a "lumens" keyword).

**Expected behavior**
Since the **Light** tab allows negative lumens, the **Sight** tab should allow the definition of vision types that can see through it.

**MapTool Info**
- Version: 1.6.1

**Desktop (please complete the following information):**
 - OS: macOS
 - Version 10.15

**Additional context**
I haven't looked at the code (I have no development environment installed, currently) but this may or may not be a lot of work.  Since the **Light** tab says darkvision (which should be any vision type with the **r** modifier) effectively has "lumens=100", I'm hoping there's already a field in the structure and it just needs the proper parsing in the **Sight** tab to be active. 🤞

selquest:
I'll update the various Sight/Light wiki pages to reflect these changes coming in 1.8

selquest:
Updated the "Introduction to Lights and Sights" wiki page to include the existing option of Lumens for Light Sources, and advise of the upcoming option to add them to Sight.

Phergus:
I don't think this is working as desired.

Map Vision set to Night.  No FoW.  Info token has Darkness Light with lumens=-120

GM Client - Hero token selected
![image](https://user-images.githubusercontent.com/34379254/87465669-4898d100-c5d2-11ea-86d4-87970b14d4f8.png)

GM Client - Elf token selected
![image](https://user-images.githubusercontent.com/34379254/87465717-5b130a80-c5d2-11ea-8bca-caa3f2efe221.png)

Hero Client - Nothing is visible in the area covered by the darkness as his light source is only a Lumens 100 and he has Normal vision.  The problem is that the darkness area itself is also not visible.  No indication that there is an area within his view that is covered by darkness.
![image](https://user-images.githubusercontent.com/34379254/87467270-ec837c00-c5d4-11ea-9ec4-21f3e8447221.png)

Elf Client - Has personal light with lumens=150.  Same issue as Hero.  Can't see that there is an area of darkness.
[...]

PR #2070, opened 2020-07-05
Add lumens support to personal light sources created with Sight types.

selquest:
- Allow Sight types to specify a lumens value.
- Update sightHelp text to include the new option
- Update lightHelp text to reflect that 100 is now only the _default_ lumens for personal light sources

Fixes RPTools/maptool#1726

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2070)


Azhrei:
I'm looking forward to trying this change.  My current character in a game can cast _deeper darkness_ which other characters cannot see in, but she can.  So right now, her light source is set to -110 lumens, but then even she can't see!  With this change, her personal sight could be set to 120 and she'd be able to see within her own _darkness_ field, and creatures with the _see in darkness_ ability could also have that vision setting.  (Of course, she'd be able to see within any other darkness area as well, even if it wasn't her spell effect.  That's not the way her ability works.  The ideal solution would be to somehow apply ownership/visibility on a per light source basis, but that might be a huge change.)

JamzTheMan:
@Azhrei 

> (Of course, she'd be able to see within any other darkness area as well, even if it wasn't her spell effect. That's not the way her ability works. The ideal solution would be to somehow apply ownership/visibility on a per light source basis, but that might be a huge change.)

Do you want to open another issue for this?

Maybe "personal lights" should be ripped out of Vision definitions going and "lights" granted the ability to apply ownership instead. And perhaps a `!ownership` as well, aka "a darkness effect that affects all but me"?

At least it's worth discussing in a new thread...

PR #2072, opened 2020-07-06
full token visible as soon as 1 pixel of token's square is within view

### LINKED ISSUE #2069: 1.7.0: full token visible as soon as 1 pixel of token's square is within view
**Describe the bug**
When a PC moves so that the NPC's token is not entirely blocked by VBL, the entire NPC becomes visible. In 1.6.1, only the part of the NPC token not blocked by VBL was visible.

**To Reproduce**
1. Create a new map with vision set to "Day"
2. Put a line of VBL
3. Place a PC token on one side
4. Place an NPC token where VBL prevents the PC from seeing it. Make sure "Is Visible over FOW" is not checked.
5. Move the PC little by little.
6. Once a single pixel of the NPC's *square* is not blocked by VBL, the entire token is visible.

**Expected behavior**
Only the portion of the NPC token that is not blocked by VBL should be visible. The PC could, for example, only see half of the NPC's token.

**Screenshots**
image
image

**MapTool Info**
- Version: 1.7.0
- Install: New
- Issue appears both on .deb package (Gentoo Linux) and .exe (Windows 7)

 - OS: Windows 7, Gentoo Linux-5.4.28

nmeier:
fixes #2069

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2072)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
@@ -3204,7 +3204,7 @@ public class ZoneRenderer extends JComponent
       timer.stop("renderTokens:OnscreenCheck");
 
       // create a per token Graphics object
-      Graphics2D tokenG = (Graphics2D) g.create();
+      Graphics2D tokenG = (Graphics2D) clippedG.create();
 
       // Previous path
       timer.start("renderTokens:ShowPath");
@@ -3376,6 +3376,7 @@ public class ZoneRenderer extends JComponent
           }
         }
       } else {
+        // fallthrough normal token rendered against visible area
         tokenG.drawImage(workImage, at, this);
       }
       timer.stop("tokenlist-7");
--- a/src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
@@ -14,25 +14,7 @@
  */
 package net.rptools.maptool.client.ui.zone;
 
-import java.awt.AlphaComposite;
-import java.awt.BasicStroke;
-import java.awt.Color;
-import java.awt.Cursor;
-import java.awt.Dimension;
-import java.awt.Font;
-import java.awt.FontMetrics;
-import java.awt.Graphics;
-import java.awt.Graphics2D;
-import java.awt.Image;
-import java.awt.Paint;
-import java.awt.Point;
-import java.awt.Rectangle;
-import java.awt.RenderingHints;
-import java.awt.Shape;
-import java.awt.Stroke;
-import java.awt.TexturePaint;
-import java.awt.Toolkit;
-import java.awt.Transparency;
+import java.awt.*;
 import java.awt.dnd.DropTargetDragEvent;
 import java.awt.dnd.DropTargetDropEvent;
 import java.awt.dnd.DropTargetEvent;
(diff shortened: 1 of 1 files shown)


Merudo:
Seems like this PR breaks "Is Visible over FoW".

Phergus:
Got to make sure that it also isn't clipping the GM view.

nmeier:
hmm, another option, I see. Will check "Visible over FoW"

nmeier:
Question: why is the token not clipped against the field of view, when "Visible over FoW"?

This is 1.6.1 - the top right wolf is shown completely as soon as one pixel comes into field of view

image

Merudo:
> Question: why is the token not supposed to be clipped against the field of view, when "Visible over FoW"?
> This is 1.6.1 - the top right wolf is shown completely as soon as one pixel comes into field of view. Shouldn't it still be clipped?

I'm not sure if I understand your question. Are you asking why enabling "Visible over FoW" makes the token visible over the FoW?

nmeier:
> > Question: why is the token not supposed to be clipped against the field of view, when "Visible over FoW"?
> > This is 1.6.1 - the top right wolf is shown completely as soon as one pixel comes into field of view. Shouldn't it still be clipped?
> 
> I'm not sure if I understand your question. Are you asking why enabling "Visible over FoW" makes the token visible over the FoW?

Rephrasing - a token with "Visible over Fow" is rendered over fog of war, makes sense. It's rendered completely though even where it can't be seen from a Vision Blocking point of view (1.6.1)

Here the top right wolf is shown over fog of war, good. It's also shown though where the VBL would block it. 

image

Logically "over fow" doesn't mean "ignore VBL" - wouldn't this be the correct result?

image

Shouldn't the VBL block off vision of tokens that are not visible

Azhrei:
I believe Jamz said that the final implementation (in 1.7) is that the token is broken into nine cells and if more than half are visible, the entire token is visible.  IIRC, it had something to do with doors being only partially visible meant that players didn’t recognize them as doors (and that was the original use case for VoFOW).  VoFOW was not meant for NPCs.

JamzTheMan:
Correct, the original intention is that this is usually used for objects like roofs, doors, etc where you want the see then even though the block vision.

You want objects that block los but you need to see what that object is, like a bus or car or tank. But you don't want to see typically they have vbl over them (think doors) so clipping them isn't possible.

There is a 1/9th tolerance built in and settable on the token (1-9) to determine how much of the token you would need to see before revealing.

It isn't the best way but until we refactor the renderer it is what it is.

nmeier:
ok, understand. Suggest to still tweak this a bit 1.6.1 vs what I think it should look like


version | GM | Player  |  Note
--|--|--|--
1.6.1 |  image | image | Token transparency not applied to top fox, halo and state not shown for the 1/9th tolerance case
suggestion |  image |  image | transparency correctly applied in all cases, all info about token pops up in 1/9th tolerance case

Phergus:
That looks good to me.

Merudo:
Figure tokens are also supposed to be fully displayed.

nmeier:
latest test case with normal token, over-fow token, figure tokens, a VBL and player token

image

[visiontest-2069.zip](https://github.com/RPTools/maptool/files/4885123/visiontest-2069.zip)

nmeier:
I've reworked the logic a bit as I found the existing if-else statements very confusing - there's now one place that checks isTokenInNeedOfClipping and based on that either the vision clipped G is used or not.

renderTokens has ~700 lines of code, that's a mouthful to refactor. I understand there's a possible future renderer so I'm keeping  the refactoring to the minimum.

FYI the tokenGraphics and managing it was part of using AffineTransform for alpha. There's opportunity to also handle flip by the rendering pipeline instead of genering a workimage.

JamzTheMan:
Yea, the current code has gotten a little out if hand over the years.

It needs an overhaul and it's on the short list now after all the scripting changes.

PR #2086, opened 2020-07-11
Make API changes from Parser refactoring

### LINKED ISSUE #1898: Improve Macro performance

nmeier:
see  
  https://github.com/RPTools/parser/pull/45
  https://github.com/RPTools/dicelib/pull/57

Changes:
 * refactor all functions that now get the variable resolver as argument
 * update/add unittest
 * New MapToolExpressionParser that adds caching of expression to the Parser. If an expression has been parsed, its soft cached Expression instance is reused.

fixes #1898

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2086)


Phergus:
@nmeier  Go ahead and commit a new build.gradle that bumps Dicelib to 1.6.5 and Parser to 1.8.2.

Issue #1263, opened 2020-02-08
Campaign Loading does not happen in a thread safe way


cwisniew:
**Describe the bug**
Details to follow (need the issue number)

nmeier:
I believe this pr fixes the existing load- /adds the missing save- SwingWorker 
https://github.com/RPTools/maptool/pull/1668

Phergus:
No issues seen loading campaigns with these changes.  Appears to be working as expected.

PR #2115, opened 2020-07-24
Minor error message key fix introduced in 1263 

nmeier:
Maptool is showing save campaign error-text if a load error occurs

I can create an issue if you want to to track this as such.

#1263

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2115)


Phergus:
Go ahead and open an issue since 1263 was part of the last release.

nmeier:
see #2116

PR #2132, opened 2020-08-01
Obvious static analysis fixes

ebudai:
Work on #2135 

This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2132)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppActions.java
+++ b/src/main/java/net/rptools/maptool/client/AppActions.java
@@ -474,7 +474,7 @@ public class AppActions {
 
             for (Map.Entry<MD5Key, Asset> entry : missing.entrySet()) {
               String remote = entry.getKey().toString();
-              repoEntries.put(remote, dir == null ? remote : new File(dir, remote).getPath());
+              repoEntries.put(remote, new File(dir, remote).getPath());
               ftp.addToQueue(
                   new FTPTransferObject(
                       Direction.FTP_PUT, entry.getValue().getImage(), dir, remote));
--- a/src/main/java/net/rptools/maptool/client/MapTool.java
+++ b/src/main/java/net/rptools/maptool/client/MapTool.java
@@ -1404,7 +1404,7 @@ public class MapTool {
             }
           };
 
-      webAppThread.run();
+      webAppThread.start();
     } catch (Exception e) { // TODO: This needs to be logged
       System.out.println("Unable to start web server");
       e.printStackTrace();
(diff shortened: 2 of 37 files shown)


Phergus:
@ebudai   Hi!  Thanks for diving in to help the community.

Two things.  There must be an issue opened for the PR(s) and you'll need to run `gradlew spotlessApply` to fix the formatting of the code.

https://github.com/RPTools/maptool/blob/main/doc/Code_Style_and_Guidelines.md

ebudai:
Apologies!  Would you like one issue per change, or all of them in one?

Phergus:
No worries.  For general code clean up you can just just use one issue and just update with the actions taken.

If you've identified a specific bug or problem that is impacting users through the UI or macro functions, it's better to keep things to one issue and hopefully one PR.

cwisniew:
> No worries. For general code clean up you can just just use one issue and just update with the actions taken.
> 
> If you've identified a specific bug or problem that is impacting users through the UI or macro functions, it's better to keep things to one issue and hopefully one PR.

As above... Looking through these changes I think they are fine for the one general clean up issue/PR...

Ugh I cant believe there are so many double checks for null in the code base... :(

ebudai:
There were several more useless null checks (esp after newing).  Would you like me to remove those as well?

Phergus:
If you're already doing other fixes in a file might as well clean those up as well.

ebudai:
One more quick question: there are several places where redundant null checks occur, but it might be to aid in debugging.  An example:

  public LargeImage(BufferedImage image, File cacheDir) throws IOException {
    info = loadInfo();
    if (info == null) {
      info = new ImageInfo(image.getWidth(), image.getHeight());
    }
    this.cacheDir = new File(cacheDir.getAbsolutePath(), getId(image));
    this.cacheDir.mkdirs();
    chunkize(image);
  }

  private ImageInfo loadInfo() {
    return null;
  }

In this example, should "if (info == null)" be removed?

Phergus:
I don't think the LargeImage class is used at all.

ebudai:
It's just an example, that kind of thing is all over.

Phergus:
So no point in making changes in it then.

ebudai:
Ok, but what about all the others?  Ignore those too?  They are not all in LargeImage, btw.

Phergus:
If the redundant checks are in active code, go ahead and make the changes.  Just no point in fixing code that isn't used and will likely be excised at some point.

ebudai:
There are other types of things to be cleaned up, such as static method called from instance, deprecated method calls (ie: new Boolean(true) -> Boolean.TRUE), redundant throws Exception, etc.  Should they all be here, or in separate PRs?

Phergus:
They can all be lumped together.

PR #2134, opened 2020-08-01
Fixed bad usage of Set<String>.contains(MD5Key)

ebudai:
contains(Object) was being called on Set<String>, with an instance of MD5Key as the parameter, leading to the check always returning false.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2134)


Phergus:
Note this code is for updating remote repositories via FTP which is not enabled in the current code.

Phergus:
Lumping in with #2135

Issue #2135, opened 2020-08-01
Code cleanup


ebudai:
There are several redundant null checks, and many places where if conditions are always true or false.

Phergus:
@ebudai let me know when you have no more changes to commit and we'll close this one.

ebudai:
I'll do another couple PRs against this so they don't get too large

Phergus:
@ebudai   Seeing a unit test failure when doing a `gradlew build` after merging in #2143 .

`MapToolLineParserTest.java`
```
org.opentest4j.AssertionFailedError: expected line #1:`expanded roll sh[...]` doesn't match ==> expected: <
expanded roll shows ...if\(1 == 1, 1, 0\) = 1.
> but was: <
expanded roll shows ???if(1 == 1, 1, 0) = 1?1?
>
```

selquest:
@Phergus that's the test failure I had in my original dev environment since #2062.  I upgraded my dev box (was Mint 19, now Mint 20) and that got the test working for me.  Still working after merging #2143 too.

That test is looking for specific characters that are apparently somewhat environment-specific - not really ideal in a unit test, but it's working in all the auto-build environments, so..

Phergus:
Interesting.  IntelliJ has had several updates lately so maybe that's what has broken it here on my Win10 system.

It's failing from both IntelliJ and the command line for me so probably not IntelliJ.  Maybe a Windows update somehow.

Phergus:
No issues seen with code changes.

PR #2137, opened 2020-08-01
a lot of small clean-ups/fixes

ebudai:
against #2135

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2137)


ebudai:
The unthrown exception signatures went pretty bad, ill remove all that stuff and make a new PR

PR #2141, opened 2020-08-02
more cleanups

ebudai:
For the exception throws removals, it looks at first blush that removing them would do harm, but they are either unchecked or the called methods deal with them in try/catch blocks.

Work on #2135 

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2141)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/tool/PointerTool.java
+++ b/src/main/java/net/rptools/maptool/client/tool/PointerTool.java
@@ -296,11 +296,11 @@ public class PointerTool extends DefaultTool {
       }
 
       // Lee: fog exposure according to reveal type
-      if (zone.getWaypointExposureToggle()) FogUtil.exposeVisibleArea(renderer, exposeSet, false);
-      else {
+      if (zone.getWaypointExposureToggle()) {
+      } else {
         FogUtil.exposeLastPath(renderer, exposeSet);
-        FogUtil.exposeVisibleArea(renderer, exposeSet, false);
       }
+      FogUtil.exposeVisibleArea(renderer, exposeSet, false);
     }
   }
 
--- a/src/main/java/net/rptools/maptool/client/tool/PointerTool.java
+++ b/src/main/java/net/rptools/maptool/client/tool/PointerTool.java
@@ -296,8 +296,7 @@ public class PointerTool extends DefaultTool {
       }
 
       // Lee: fog exposure according to reveal type
-      if (zone.getWaypointExposureToggle()) {
-      } else {
+      if (!zone.getWaypointExposureToggle()) {
         FogUtil.exposeLastPath(renderer, exposeSet);
       }
       FogUtil.exposeVisibleArea(renderer, exposeSet, false);
--- a/src/main/java/net/rptools/maptool/client/ui/campaignproperties/CampaignPropertiesDialog.java
+++ b/src/main/java/net/rptools/maptool/client/ui/campaignproperties/CampaignPropertiesDialog.java
@@ -738,7 +738,7 @@ public class CampaignPropertiesDialog extends JDialog {
             gmOnly = false;
             owner = false;
           }
-          owner = gmOnly == true ? false : owner;
+          owner = gmOnly ? false : owner;
           try {
             Light t =
                 new Light(
--- a/src/main/java/net/rptools/maptool/client/ui/campaignproperties/CampaignPropertiesDialog.java
+++ b/src/main/java/net/rptools/maptool/client/ui/campaignproperties/CampaignPropertiesDialog.java
@@ -738,7 +738,7 @@ public class CampaignPropertiesDialog extends JDialog {
             gmOnly = false;
             owner = false;
           }
-          owner = gmOnly ? false : owner;
+          owner &= !gmOnly;
           try {
             Light t =
                 new Light(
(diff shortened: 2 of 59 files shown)


Phergus:
I am somewhat concerned about removing the try/catch blocks specifically in AddResourceDialog.  Someone went to the effort to properly setup `showMessage()` with i18n strings and that usually means users were getting cryptic exception messages.  However it doesn't seem like those exceptions could be seen there.

src/main/java/net/rptools/maptool/client/tool/PointerTool.java
```diff
@@ -296,11 +296,11 @@ public void exposeFoW(ZonePoint p) {
       }
 
       // Lee: fog exposure according to reveal type
-      if (zone.getWaypointExposureToggle()) FogUtil.exposeVisibleArea(renderer, exposeSet, false);
-      else {
+      if (zone.getWaypointExposureToggle()) {
+      } else {
```
cwisniew, line 300: For the change in PointerTool I prefer changing the if into the negative rather than an empty if block

src/main/java/net/rptools/maptool/client/ui/campaignproperties/CampaignPropertiesDialog.java
```diff
@@ -738,7 +738,7 @@ private void commitSightMap(final String text) {
             gmOnly = false;
             owner = false;
           }
-          owner = gmOnly == true ? false : owner;
+          owner &= !gmOnly;
```
cwisniew, line 741: Maybe 
owner = gmOnly ? false : owner;

owner &= !gmOnly;

makes me have to stop to think about it each time :)

src/main/java/net/rptools/maptool/model/Path.java
```diff
@@ -139,18 +139,16 @@ public T getLastJunctionPoint() {
         indexCheck++;
       }
 
-      if (waypointCheck.isEmpty())
-        processPath.addAllPathCells(nw.calculatePath(prevPoint, terminalPoint));
-      else {
+      if (waypointCheck.isEmpty()) {
+      } else {
```
cwisniew, line 143: Prefer not to have an empty if block

src/main/java/net/rptools/maptool/client/ui/campaignproperties/CampaignPropertiesDialog.java
ebudai: Intellij suggested the same thing; I went ahead and k-mapped it in my head.  I'll use the popular suggestion.

ebudai:
> I am somewhat concerned about removing the try/catch blocks specifically in AddResourceDialog. Someone went to the effort to properly setup `showMessage()` with i18n strings and that usually means users were getting cryptic exception messages. However it doesn't seem like those exceptions could be seen there.

I'll fix this by rethrowing the exceptions caught in AppSetup.installLibrary(), and restoring the catch blocks in AddResourceDialog

Phergus:
Eric - if you feel confident that those exceptions can't actually be seen there in AddResourceDialog then I'm willing to roll the dice on it.

ebudai:
> I am somewhat concerned about removing the try/catch blocks specifically in AddResourceDialog. Someone went to the effort to properly setup `showMessage()` with i18n strings and that usually means users were getting cryptic exception messages. However it doesn't seem like those exceptions could be seen there.

The exceptions are supposed to come from AppSetup.installLibrary(), but the lines which throw those are inside an EventQueue.invokeLater().  I think the solution is to put the I18n handlers inside the invokeLater() parameter.

ebudai:
> Eric - if you feel confident that those exceptions can't actually be seen there in AddResourceDialog then I'm willing to roll the dice on it.

They definitely cannot.  They are checked exceptions, so omitting them in the throws clause is a compile error.  It comes down to EventQueue.invokeLater() not throwing any checked exceptions.

ebudai:
> Also as a general comment feel free to change to lambdas when refactoring since the next release will be java 14 and no longer target older JVMs.

This is in another PR due to its immense size.

Phergus:
> > I am somewhat concerned about removing the try/catch blocks specifically in AddResourceDialog. Someone went to the effort to properly setup `showMessage()` with i18n strings and that usually means users were getting cryptic exception messages. However it doesn't seem like those exceptions could be seen there.
> 
> The exceptions are supposed to come from AppSetup.installLibrary(), but the lines which throw those are inside an EventQueue.invokeLater(). I think the solution is to put the I18n handlers inside the invokeLater() parameter.

That would better.

src/main/java/net/rptools/maptool/client/ui/campaignproperties/CampaignPropertiesDialog.java
Phergus: Yeah. I had to stop and work through it in my head when I saw that.  This is not performance sensitive code so readability is probably more important.

ebudai: It wouldn't even be faster.  It's just a bad habit.

Phergus:
@cwisniew Need your approval on the three changes.

cwisniew:
> > Also as a general comment feel free to change to lambdas when refactoring since the next release will be java 14 and no longer target older JVMs.
> 
> This is in another PR due to its immense size.

Opps I want meaning ALL the lambdas, just what you were cleaning up at the time. Feel free to do them all if you want but just didn't want my comment to be viewed as pointing out what to do :)

Phergus:
@ebudai, @cwisniew   you guys need to resolve the open discussions.

PR #2144, opened 2020-08-02
minor perf, readability improvements

ebudai:
Also some bugfixes

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2144)

### DIFF
--- a/src/main/java/net/rptools/lib/BackupManager.java
+++ b/src/main/java/net/rptools/lib/BackupManager.java
@@ -82,7 +82,7 @@ public class BackupManager {
         new Comparator<File>() {
           public int compare(File o1, File o2) {
 
-            return o1.lastModified() < o2.lastModified() ? -1 : 1;
+            return Long.compare(o1.lastModified(), o2.lastModified());
           }
         });
 
--- a/src/main/java/net/rptools/lib/ModelVersionManager.java
+++ b/src/main/java/net/rptools/lib/ModelVersionManager.java
@@ -98,9 +98,7 @@ public class ModelVersionManager {
 
     List<ModelVersionTransformation> transformList =
         transformMap.computeIfAbsent(version, k -> new LinkedList<ModelVersionTransformation>());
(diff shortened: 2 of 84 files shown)


Azhrei:
One thing that would help performance is for calls to the Logger to use lambdas whenever the string being passed is not a simple string.  In other words, the latter is better as it defers string building to only if the lambda is executed:

    log.info("message" + var1 + "message");
    log.info(() -> "message" + var1 + "message");

So if you're in the code anyway and see examples of the former, they may as well be changed.  Just something else to add to the "cleanup" list...

cwisniew:
> One thing that would help performance is for calls to the Logger to use lambdas whenever the string being passed is not a simple string. In other words, the latter is better as it defers string building to only if the lambda is executed:
> 
> ```
> log.info("message" + var1 + "message");
> log.info(() -> "message" + var1 + "message");
> ```
> 
> So if you're in the code anyway and see examples of the former, they may as well be changed. Just something else to add to the "cleanup" list...

I would say this detracts from readability and performance wise unless its in a tight loop is negligible. So I am all for leaving it as is

Azhrei:
> I would say this detracts from readability and performance wise unless its in a tight loop is negligible. So I am all for leaving it as is

I'm not surprised.  I find that an extra four characters makes it much more difficult for me, too.  YMMV.  (And if you're not concerned about performance with logging turned on, you should talk to Wolph42.  He has reported this problem more than once.)

cwisniew:
> > I would say this detracts from readability and performance wise unless its in a tight loop is negligible. So I am all for leaving it as is
> 
> I'm not surprised. I find that an extra four characters makes it much more difficult for me, too. YMMV.

Stop with the snark or don't comment please.

> (And if you're not concerned about performance with logging turned on, you should talk to Wolph42. He has reported this problem more than once.)

Have you done any investigation into Wolph42's slow downs or did you just assume it was this? Is he turning logging on at debug or info level? If so this will make it slower if anything else. Making changes like this with no information wont help anyone.

Issue #2046, opened 2020-06-22
NPE on cancel out of "Rename Macro Group..."


bubblobill:
**Describe the bug**
Cancel out of "Rename Macro Group..." dialog produces NPE

**To Reproduce**
1. Go to selected panel
2. Right Click on macro group, select "Rename Macro Group..."
3. Click Cancel

**Expected behavior**
No change

java.lang.NullPointerException
	at net.rptools.maptool.client.ui.macrobuttons.buttongroups.ButtonGroupPopupMenu$RenameGroupAction.actionPerformed(ButtonGroupPopupMenu.java:647)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2308)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:405)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
	at java.desktop/javax.swing.AbstractButton.doClick(AbstractButton.java:369)
	at de.muntjak.tinylookandfeel.TinyMenuItemUI.doClick(Unknown Source)
	at de.muntjak.tinylookandfeel.TinyMenuItemUI$MouseInputHandler.mouseReleased(Unknown Source)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6589)
	at java.desktop/javax.swing.JComponent.processMouseEvent(JComponent.java:3342)
	at java.desktop/java.awt.Component.processEvent(Component.java:6354)
	at java.desktop/java.awt.Container.processEvent(Container.java:2261)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:4966)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2319)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4798)
	at java.desktop/java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4914)
	at java.desktop/java.awt.LightweightDispatcher.processMouseEvent(Container.java:4543)
	at java.desktop/java.awt.LightweightDispatcher.dispatchEvent(Container.java:4484)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2305)
	at java.desktop/java.awt.Window.dispatchEventImpl(Window.java:2772)
[...]

PR #2145, opened 2020-08-02
Fix for Rename Macro Group cancellation causing NPE #2046

ebudai:
fikes #2046 
null check inserted

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2145)


Phergus:
@ebudai  when working on issues opened by others, add a comment to it that you are going to work on it. Thanks.

PR #2147, opened 2020-08-03
minor perf improvement - change keySet() to entrySet() or values() #2135

ebudai:
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2147)

### DIFF
--- a/src/main/java/net/rptools/lib/swing/ImagePanel.java
+++ b/src/main/java/net/rptools/lib/swing/ImagePanel.java
@@ -406,9 +406,9 @@ public class ImagePanel extends JComponent
   }
 
   protected int getImageIndexAt(int x, int y) {
-    for (Rectangle rect : imageBoundsMap.keySet()) {
-      if (rect.contains(x, y)) {
-        return imageBoundsMap.get(rect);
+    for (var imageBounds : imageBoundsMap.entrySet()) {
+      if (imageBounds.getKey().contains(x, y)) {
+        return imageBounds.getValue();
       }
     }
     return -1;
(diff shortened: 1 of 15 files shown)


Phergus:
___
*[src/main/java/net/rptools/maptool/client/functions/TokenLightFunctions.java, line 186 at r1](https://reviewable.io/reviews/rptools/maptool/2147#-MDqptWSBqiBLbzYq3NH:-MDrDVCGAwQe0badheIQ:bx3ga3h) ([raw file](https://github.com/rptools/maptool/blob/939d6595ef1ac12ad937e17270775ec2247cd403/src/main/java/net/rptools/maptool/client/functions/TokenLightFunctions.java#L186)):*
<details><summary><i>Previously, ebudai (Eric Budai) wrote…</i></summary><blockquote>

its always false
</blockquote></details>

How do you figure that?  Run `[getLightSource("Generic", "*")]` on a token with generic light source and it will return true.  Myself, I would have used `"*".equals(name)` out of an abundance of caution but that's neither here nor there.

Phergus:
___
*[src/main/java/net/rptools/maptool/client/functions/TokenLightFunctions.java, line 186 at r1](https://reviewable.io/reviews/rptools/maptool/2147#-MDqptWSBqiBLbzYq3NH:-MDtckUX5qoKxlJ1giAg:byvbo0f) ([raw file](https://github.com/rptools/maptool/blob/939d6595ef1ac12ad937e17270775ec2247cd403/src/main/java/net/rptools/maptool/client/functions/TokenLightFunctions.java#L186)):*
<details><summary><i>Previously, ebudai (Eric Budai) wrote…</i></summary><blockquote>

we are in a block in which category.equals("*") is true.  
On line 176:
if (category.equals("*") && name.equals("*")) {
   return token.hasLightSources();
 }
This means that if name.equals(*) is true, we have returned, so name.equals(*) must always be false (here).

I agree about *.equals(category).  I'll change that now.
</blockquote></details>

You're right.  I was confusing what was being changed with the first if() check.

PR #2150, opened 2020-08-04
FogUtil.calculateVisibility() speed improvements

ebudai:
… improved

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2150)


Phergus:
What was the performance improvement from your testing?

ebudai:
https://docs.google.com/spreadsheets/d/1rs7saqvNUSViIN1Zk_s71Cg72SZT-hnA-C8W0oGtLiE/edit?usp=sharing

cwisniew:
> What was the performance improvement from your testing?

It would be interesting to know this as well... 

I have approved it in any case though as its a fairly standard optimisation (comparing squares of distances/magnitudes rather than tanking square root and comparing) when it comes to computer graphics with no down sides and it makes even more sense in this case where the original value had to be multiplied by 1000 so that the sort was stable.

Phergus:
Even 10% improvement will be nice.

Issue #2151, opened 2020-08-05
Visible Over Fog of War objects are always visible


Phergus:
**Describe the bug**
With current develop branch, items marked as Visible over Fog of War are always visible.

**To Reproduce**
1. Load attached campaign.
2. Go to Hero token and expose current area.
3. Switch to Player View
4. See all of the VOFoW  doors on map:
![image](https://user-images.githubusercontent.com/34379254/89364781-84363080-d690-11ea-96e3-aa528bc291ad.png)

Same thing is seen on a connected player client.

**Expected behavior**
Expect only doors within visual range of token.

**MapTool Info**
- Version: Current develop branch at least as far back as July 20th.

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: 10

**Additional context**
[DonjonMapTest180alpha.cmpgn.zip](https://github.com/RPTools/maptool/files/5025795/DonjonMapTest180alpha.cmpgn.zip)
Remove `.zip` suffix.

ebudai:
I'll take a stab at this one.

Phergus:
Go for it.   Be aware that there are a lot of complex interactions between sight, light and VBL then compounded by GM vs Player view plus Individual Views and so on.

JamzTheMan:
I know this didn't do this originally, at least for IViews/IFoW. May be worth it to test in 1.5.0 and if it behaves differently, may follow through and see when the bug was introduced.

Phergus:
MT 1.7.0 works correctly so it is something that popped up in develop.  It's present in code back to 7/20 so some time before that.

JamzTheMan:
Ah well that makes it easier then :)
To the Git Blame!

nmeier:
Could be related to #2069

nmeier:
yeah, broken in https://github.com/RPTools/maptool/commit/eeb57c9ea3108631abaa3e3c6930d1bf745e934a

I was tweaking that over-fow tokens would become visible (again) when a certain amount of surface is visible.

ebudai:
I found the same, and have restored the offending bit of code (in timer 7).  I'm trying to reduce it now as there are a lot of DRY violations.

nmeier:
it's a brutal method - the changes in #1801 and #2069 were done to speed up rendering and also refactor the method a bit. There's a test case in #2609 that doesn't cover tokens in the object layer exposed here. 
I think I know how to add support for the object layer case by removing more code that changes the flow for clipping FYI.

[...]

PR #2153, opened 2020-08-06
fix for FoW issue

ebudai:
Restores previous handling of VoFoW tokens to address #2151

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2153)


nmeier:
this looks like a simple revert rather than a fix that carries the intended change forward, is that so?

nmeier:
The whole code block you're intending to revert was removed on purpose to improve readability. I suggest to extend the test case and handle object layer objects correctly rather than going half a step back as you're bringing back the multi page if-else statement.

ebudai:
You are correct, it's a reversion instead of a "fix".  I attempted to reduce it, but the condition on the else block looked strange, so I just put it back to what it was when it was working.  I'll reattempt.

ebudai:
I have reduced it as much as I can.  I cannot seem to find any tests regarding zone rendering to extend.

ebudai:
I am having trouble reproducing the issue on the develop branch with the test campaign included in your (now deleted) comment, but that is almost certainly my fault.  I'll continue to investigate after work.

ebudai:
Apologies, the comment wasn't deleted; I confused the discussion in the PR with the discussion here.

I cannot get the block to reduce further.

Phergus:
@nmeier  did you have other thoughts on this?

ebudai:
@nmeier stated he would attempt to rectify #2151 via the original refactoring.  I have nothing more to contribute.

PR #2168, opened 2020-08-16
Fix for variables beginning with tru or fal... 

cwisniew:
New Crowdin updates


cwisniew:
Requires a new build of the parser project after applying RPTools/parser#50

This adds true and false to MapToolVariableResolver as constants (as this is closest there is to a symbol table for the parser). Now the lexer will not try and search for true and false so there will be no ambiguity, and they will be able to be used anywhere variables are except as targets of assignment.

PR #1872, opened 2020-05-18
1.7 maintenance

JamzTheMan:
So we don't forget to merge back to master.

@Phergus confiming this is the branch you created the release from?

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/1872)

PR #2181, opened 2020-08-26
Include in develop PRs merged into main

Azhrei:
Somehow #1872 was missed as well as the commits that updated the GitHub templates.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2181)


Phergus:
GitHub no likee.

Azhrei:
I guess that’s because these commits were based on repo state and that isn’t in main?  I don’t see how that could be, since this PR was generated by GitHub by telling it to bring develop up to date with main.  But I’m no Git expert...

So...  does that mean develop has to first be merged into main?  What a PITN. Do you want me to just create a PR for these changes instead of trying to merge main into develop?

Phergus:
I don't really know why it is complaining but I wasn't actually surprised that it did.

I suspect that it will require using cherry pick to get the changes to the templates from main to the develop branch.  Which may or may not cause things to blow up again when we merge develop back to main.

This pretty much sums up my philosophy on git.
git

PR #2189, opened 2020-08-31
Add 'removeDrawing' macro function

### LINKED ISSUE #2188: Feature Request: the ability to remove a drawing via a macro
Currently, macros are able to get information about drawings, change their properties, move them around, etc.

They're not able to create or delete them however. Creating drawings is definitely more convenient by hand, so it's understandable for a drawing to not be create-able via a macro, but they should be delete-able.

For a concrete example of how this might help:
In our D&D campaign, we use the drawing tool to draw the AOE for fireballs and the like.
It would be convenient if the macro that prints fireball damage could also then delete the AOE since it's been fully resolved.

**Describe alternatives**

Setting opacity. It clutters up the draw explorer list to use this workaround.

**Additional context**

This request is proxied over from this forum thread someone else posted: https://forums.rptools.net/viewtopic.php?f=26&t=28823&p=276549

euank:
This adds a 'removeDrawing' macro function. It's a trusted macro (consistent with the other similar drawing macros).

I've manually tested that it seems to work for both drawing-tool drawings and template drawings.

Fixes #2188

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2189)


euank:
I don't think I need to. The `undoDraw` function on both the server and client calls `removeDrawable` for us if I'm reading this right.

I'm basing this off of:

https://github.com/RPTools/maptool/blob/6d193e3f5cc635aed30730ec09fde2a210870054/src/main/java/net/rptools/maptool/client/ClientMethodHandler.java#L338-L350

and

https://github.com/RPTools/maptool/blob/6d193e3f5cc635aed30730ec09fde2a210870054/src/main/java/net/rptools/maptool/server/ServerMethodHandler.java#L758-L777

Phergus:
Interesting.  I just happened to look at the `bringToFront()` method and noticed it calls `removeDrawable()` on the zone just before `MapTool.serverCommand().undoDraw()` so that is probably redundant.

Issue #1827, opened 2020-05-14
Replace EventDispatcher with an off the shelf solution


cwisniew:
**Is your feature request related to a problem? Please describe.**
The EventDispatcher class should be replaced with an off the shelf event dispatcher. 

**Describe the solution you'd like**
Replace EventDispatcher with Google EventBus. 

**Additional context**
This required replacing current 

- MapTool.ZoneEvent
- MapTool.PreferencesEvent

Used in 
- ClientMethodHandler
- MapTool
- MapToolFrame
- PreferenceDialog
- CommandPanel
- HTMLFrameFactory
- ImpersonatePanel
- SelectionPanel
- WebAppInitiative

MapTool.setCampaign() should be changed to dispatch a new campaign event, replacing the below code
```java
    AssetManager.updateRepositoryList();
    MapTool.getFrame().getCampaignPanel().reset();
    MapTool.getFrame().getGmPanel().reset();
    // overlay vanishes after campaign change
    MapTool.getFrame().getOverlayPanel().removeAllOverlays();
    UserDefinedMacroFunctions.getInstance().loadCampaignLibFunctions();
```

Eventually this can be expanded to include other events things like advancing initiative etc.

cwisniew:
Should also replace 
ZoneRenderer.updateAfterSelection which currently does
```java
  public void updateAfterSelection() {
    MapTool.getFrame().getSelectionPanel().reset();
    MapTool.getFrame().getImpersonatePanel().resetIfNotImpersonating();
    HTMLFrameFactory.selectedListChanged();
    repaintDebouncer.dispatch();
  }
```

ModelChangeEvent which is fired off by
- Grid
- Zone

Used by
- UndoPerZone
- DrawPanelTreeModel
- HTMLFrameFactory
- AbstractMacroPanel
- ImpersonatePanel
- SelectionPanel
- InitiativePanel
- TokenPanel
- TokenPanelTreeModel
- ZoneMiniMapPanel
- ZoneRenderer
- ZoneView
- BaseModel
- Grid
- ModelChangeListener
- Zone
- WebAppInitiative
- WebTokenInfo


(yeah its a lot)

Merudo:
Related: the classes `CommandPanel` and `PlayerListModel` implement the deprecated `Observer` interface, which should be replaced by the EventDispatcher.

cwisniew:
[...]

Issue #987, opened 2019-12-13
Initiative Events


mikehenry-se:
**Is your feature request related to a problem? Please describe.**
I'm trying to update tokens when they get the initiative and reset all tokens a new round starts (or ends).

**Describe the solution you'd like**
I'd be happy to see two new events; e.g. onNextInitiative(Token) and onRoundStarted(TokenList) that can be implemented and which are 1) called when a token receives the initiative through the "next" button with the token, and 2) when a new round is started, with all the tokens in the initiative list as the argument. Of course, the arguments are nice-to-haves.

**Describe alternatives you've considered**
I'm doing it now with a separate macro but it would be cleaner and easier to use if it hooked into the existing initiative framework, rather than remembering to push separate buttons for this.

Thanks in advance!

Phergus:
Related to #720

selquest:
@cwisniew Is this actually in progress or tied up in an ongoing rewrite somewhere?  I'm looking to make some incremental improvements to Initiative this weekend, and I feel like there are some steps that can be taken here even before #1827 and related things.  If we're not committed to deferring this, my thought was to add support for `onInitiativeChange()` using an approach similar to the existing `onTokenMove()`.

cwisniew:
@selquest this was waiting on expanding of the event system so we could do everything in a consistent way. I guess this depends on if we are going to do a large refactor for other things, if so its not so important to wait, but that something we are still deciding on.

selquest:
I was exploring this a little further out of curiosity, and my proof of concept turned into what I think is a pretty good basic implementation.  Up to you folks whether it's worth merging of course, but I think that given how often it's come up here and on discord, it probably makes sense to get a basic implementation out.  Since it's based so heavily on `onTokenMove`, any future changes to the event mechanisms shouldn't have too much added complexity for needing to handle this one too.

selquest:
Two new initiative events added.
- `onInitiativeChangeRequest` - called when the initiative is changing, can deny the change by setting `init.denyChange` to `1`.  Unlike _onTokenMove_, this event can have macros on multiple Lib:Tokens.  All matching handlers will be called, and the change will be denied if at least one denial is received.
- `onInitiativeChange` - called once an initiative change has been approved and is being applied.  Also supports multiple Lib:Tokens.

[...]

Issue #720, opened 2019-09-18
New Events Macros


rkathey:
**Is your feature request related to a problem? Please describe.**
Add additional event macros such as those at http://lmwcs.com/rptools/wiki/Category:Event

The new events would cover things like clicking on a token or creating a new token.

**Describe the solution you'd like**
Create a new set of events not covered in the following
onCampaignLoad
onChangeSelection
onMouseOverEvent
onMultipleTokensMove
onTokenMove

**Describe alternatives you've considered**
BAU

**Additional context**
This is a request from @kayila

Merudo:
Having "onMouseOver" being integrated would be good.

JamzTheMan:
What do you mean by integrated? Like not needing a frame to hold onMouseOverEvent? (If so yes!)

Merudo:
I mean it becoming an actual event instead of being a fake one.

kayila:
Included in this, we'd like to see an onClick and onTokenCreated.

Ideally this could be done as a proper event system, and the existing event macros move to that rather than haphazardly adding special macro names as one offs.

JamzTheMan:
onCampaignStart could be useful. Would this fire when connected to a server as well or should there be a onServerJoin?  I could see these used to move players to a specific map, among other things.

Question: where should these be defined? I suggested via UI (campaign properties?)

But what should that be? Should they simply accept a macro and or UDF? Should they contain it's own macro command?

In other words, as a new user, for on onCampaignStart, can I enter a few lines of macro code like gotoMap and a broadcast ("Hello!")

Or will we require them to create a lib token somewhere and call a macro and such? (Obviously I advocate for the former and not later)

I can assist with throwing together the UI components if needed (I know some have issues with h abillie forms, it depending on when this gets done, a FX component may be better)

kayila:
So, the answer to the question of where the handlers are located is a little more complicated than just storing them in one place. I'm thinking there would be a way to store some, such as onCampaignLoad and onServerJoin on the campaign itself, no use of lib tokens. The onClick event, however, would trigger a macro on the token which was clicked on.

I think having seperate onCampaignLoad vs onServerJoin is a good idea, maybe also an onPlayerJoin/onGMJoin?

Also, these macros should probably be passed useful information, such as onClick being passed who clicked on it, onServerJoin passing the user who joined, etc.

JamzTheMan:
OnServerJoin is probably good enough as long as who joined is passed and you can if isGm it in the macro.

[...]

PR #2190, opened 2020-08-31
Add onInitiativeChangeAttempt and onInitiativeChangeCommit

selquest:
This started as an exercise in curiosity, but I think it turned into a pretty solid basic implementation based on the approach from onTokenMove.

When the initiative is changing, look for Lib:Tokens with a macro named `onInitiativeChangeAttempt`, and pass them some information about the desired change.  For each of the handlers found, check the value of `init.denyChange` to prevent the initiative change if it's been set to `1`.

If the change passes the calls for vetoes, call any Lib:token macros named `onInitiativeChangeCommit`, and pass the same information about the change.  These handlers no longer have the option to veto.

This currently effects only the next/previous initiative operations (whether triggered via macro or the Panel), meaning that any mechanism for jumping directly to a specific initiative slot (the `setCurrentInitiative()` macro, the `Make Current` action from the panel, etc.) will bypass this event.  This is largely a concession to complexity, attempting to avoid modifications to the `InitiativeList.setCurrent(int)` method that is widely used in the internals (resetting the current initiative after adding/removing, sorting the list, etc.), but also ensures that the GM always has a way of forcing an initiative change without the need to muck about with any macros.  If we decide to make other kinds of changes trigger this event, we'll need to provide some optional parameters and other mechanisms for skipping it.

Here's a sample of the JSON passed into the event macro:
```json
{
  "old":   {
    "round": 4,
    "offset": 0,
    "initiative": "22",
    "token": "2F12FEEE06B54B639228DBE56CE451AA"
  },
  "new":   {
    "round": 4,
    "offset": 1,
    "initiative": "14.1",
    "token": "5158F51707FC4397AB5E2C71C94116E5"
  },
  "direction": "NEXT"
}
```

And a (pretty nonsensical) example of an `onInitiativeChangeAttempt` implementation that prevents any attempt to move NEXT from a PC token:
```
<b>onInitiativeChange</b>:<br />
[h: jsonArgs = macro.args]
<pre>[r: json.indent(jsonArgs)]</pre>
Change Denied: [r: init.denyChange = and(isPC(json.get(json.get(jsonArgs, "old"), "token")),json.get(jsonArgs, "direction") == "NEXT")]
```

I understand we're probably planning to overhaul the initiative system at some point in the future, but I think this gets us enough bang-for-buck to be worth releasing in the meantime.  It's a fairly self-contained change in the InitiativeList model, so I don't think it adds a lot of complexity for any future plans there.  Since it's almost identical to `onTokenMove`, any event system from #1827 would be tied in here in the same way it does there.

#987
#720

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2190)


cwisniew:
Not a fan of the only first token found is called as that limits the things you can do with tracking initiative for states / spell effects etc, as everything has to be in the same macro.

selquest:
> Not a fan of the only first token found is called
Yeah, it's an unfortunate limitation - as you say, carried over from onTokenMove - but it also kind of makes sense.  Since the `onTokenMove` and `onInitiativeChange` macros can prevent the thing from happening, how should we handle it if there are multiple handlers present?  Is the event denied if ANY of them deny it?  Is it last one wins?  Is it some kind of voting mechanism?  What do we do for a handler that THOUGHT it was allowing the event to continue (and thus, set up some things with the expectation that the requested change was going to take effect)?

With a single handler implementation, we tell framework authors and GMs that they have to resolve all those questions.  Their one `onTokenMove` or `onInitiativeChange` handler can include whatever combination of sub-routines and logic they need, but at the end of the day the system is only going to consult one Go/No-Go value.

cwisniew:
> > Not a fan of the only first token found is called
> 
>  Since the `onTokenMove` and `onInitiativeChange` macros can prevent the thing from happening, how should we handle it if there are multiple handlers present? Is the event denied if ANY of them deny it? Is it last one wins? Is it some kind of voting mechanism? What do we do for a handler that THOUGHT it was allowing the event to continue (and thus, set up some things with the expectation that the requested change was going to take effect)?
> 
> With a single handler implementation, we tell framework authors and GMs that they have to resolve all those questions. Their one `onTokenMove` or `onInitiativeChange` handler can include whatever combination of sub-routines and logic they need, but at the end of the day the system is only going to consult one Go/No-Go value.
>

That means you cant get more than one token that needs to know about inventory, which is kind of why I dont want to do it until there is a proper event system or it locks MapTool into that for ever and a day.

selquest:
> That means you cant get more than one token that needs to know about initiative

Well, not really though - you can only have one token/macro that has to be the authority on whether or not initiative changes are allowed to go forward.  If that single handler wants to pass the old/new initiative info along to others (telling a spell state tracker library that token A's turn just ended, and token B's turn is about to begin, for instance), it can.  If we wanted to build-in a bit better support for that kind of thing it would be easy enough to add another couple of events like `afterInitiativeChange` that would function more like `onCampaignLoad` - every Lib:token with that macro defined could fire, because it doesn't get to abort the action.  Individual tokens could have macros like `onInitiativeGained` and `onInitiativeLost` that fire automatically, though I'd probably want a way to hide those kinds macros from players to avoid confusion.

Knowing that this hypothetical future event system is floating out there, it makes sense that we wouldn't want to add all of that extra infrastructure in now, but I wasn't really expecting the "only a single handler for yay/nay" to be the sticking point.  I'm not sure what use cases would really need more than one handler, but if it's that big of a deal would it be worthwhile to go ahead and add the `afterInitiativeChange` event alongside this? We could then fire both of those on as many Lib:tokens as have them, with the understanding that if ANY of the onChange handlers reject the change, it's blocked.  Framework authors would need to use one event macro to vote yay/nay, and another to handle any setup once they know it's definitely happening.  

To me the biggest downside of this approach is that it's actually dependent on Lib:Tokens at all - I'd prefer these kinds of handlers to be Campaign macros or on some kind of new Event Macros panel.  The other major limitation is that the handler macro fires on the client of whoever was trying to advance the initiative, and wouldn't have a mechanism to block for somebody else's input before making its decision.

> or it locks MapTool into that for ever and a day.

A lot of my reasoning for basing this on `onTokenMove` was that I don't know what kind of infrastructure we're really anticipating with whatever this new "event system" would be (or what kind of timeframe we're looking at for that), but figured it'll already need to handle a switchover of `onTokenMove`.  If we're really worried about kicking off longterm support for this though, we can always stamp it an Experimental Feature in the docs.  Basically warn framework authors that while it's here, we're not sure what it'll look like in future versions.

Azhrei:
I can foresee a campaign having a map in which individual battles are taking place and that needs one initiative type, and another map that has a large scale combat happening and it uses a different initiative type.

However, if a game system has multiple types of init like that, then they would both be implemented on the same lib token, right?

What if there are two lib tokens that both manage initiative along with myriad other aspects of the game, and a fledgling GM wants to use both?  For example, say the first lib token manages init and various combat rules, and a second lib token manages init and spell effects.  They each manage init because they each need to know the order of tokens in combat.  How does a GM combine the two?  The event proposed here doesn't have a way to notify the macro that there are others that will be executing.

Maybe the macro should be given a list of all lib tokens (as token ids) that have macros that will be executed?  Then each one would know which other ones are out there.  A naive implementation would ignore the list entirely.  A slightly better implementation could check and if the list length != 1, print a warning and abort().  A truly sophisticated one could check the names of the other lib tokens and take their actions into account, such as a ruleset lib token that recognizes the lib token for the bag-of-tricks and takes appropriate action?

selquest:
> I can foresee a campaign having a map in which individual battles are taking place and that needs one initiative type, and another map that has a large scale combat happening and it uses a different initiative type.
> However, if a game system has multiple types of init like that, then they would both be implemented on the same lib token, right?

Yeah, if your system has multiple different ways that it might want to interact with initiative, you would probably want to have one handler that chooses between the two:
```
[h: jsonArgs = macro.args]
[h, if(condition), code: {
    [macro("approach1@Lib:mysystem"): jsonArgs]
    [init.denyChange = macro.return]
}; {
    [macro("approach2@Lib:mysystem"): jsonArgs]
    [init.denyChange = macro.return]
}]
```
> What if there are two lib tokens that both manage initiative along with myriad other aspects of the game, and a fledgling GM wants to use both? ...
>Maybe the macro should be given a list of all lib tokens (as token ids) that have macros that will be executed?

I wouldn't really recommend structuring things that way just from a separation of concerns perspective, but there's no technical reason we can't have a list of all the matching macros included in the JSON that gets passed in.  Even without it, the handler macros could go find out on their own in the macro if they really wanted to.

> They each manage init because they each need to know the order of tokens in combat. How does a GM combine the two?

This part I don't really understand.  In all of these cases initiative is _managed_ by the MT InitiativeList, and exposed through macros and the Initiative Panel.  The single vs multiple handlers question boils down to how many mechanisms we expect a single campaign to include that have the ability to prevent initiative passing to the next token.  The best example I can come up with for needing multiples revolve around spell durations and such (until we get those more fully integrated with MT States anyway 😄):  You might have the main framework implemented in a system lib:token that has a handler to make sure the current token has expended all action points before ending its turn, while a separate library is tracking states and knows that this token has to make a decision about which other token gets the benefit from its Leadership Focus this round.  As currently implemented in the PR, the GM would have to make sure that the state library did NOT have an `onInitiativeChange` macro, but might adjust the macro on the main framework token to look something like:
```
[h: jsonArgs = macro.args]
[h: oldToken= json.get(json.get(jsonArgs, "old"), "token")]
[h: newToken = json.get(json.get(jsonArgs, "new"), "token")]
[h: pointsRemaining = getProperty("AP", oldToken)]
[h, macro("checkStates@Lib:states"): jsonArgs]
[h: statesNotReady = macro.return]
<!-- decide -->
[h: init.denyChange = or(pointsRemaining, statesNotReady)]
<!-- setup the next token's turn if appropriate-->
[h, if(!init.denyChange): resetProperty("AP", newToken)]
```
If we split the decision and setup parts, then this main handler gets a bit cleaner because it only has to check the remaining AP.  The `checkStates@Lib:states` macro becomes its own `onInitiativeChange` handler that gets to block the change in its own right.  Then if a change is allowed to go through, the `afterInitiativeChange` macro would fire on each of them and that's where the main lib:token would do the AP reset for the next token.

Azhrei:
> > They each manage init because they each need to know the order of tokens in combat. How does a GM combine the two?
> 
> [...] As currently implemented in the PR, the GM would have to make sure that the state library did NOT have an `onInitiativeChange` macro, but might adjust the macro on the main framework token to look something like:

Okay, that answers my question. 👍  I was trying to determine how extensive the changes might need to be when there were two drop-in lib tokens, if each one expected to be the only one being used.  If the event were propagated to both, and one of them chose to cancel the change, the other may have already been invoked and performed operations that are not easily undone.  Because the order in which the macros is invoked is undefined, it would be impossible for either one to account for the other.

Perhaps if there are multiple macros that could be invoked, none of them should be executed and a warning produced instead?  I don't recall anyone having problems with this in regards to `onTokenMove though, so maybe it's not an issue.  Or maybe there just aren't multiple drop-in tokens that use `onTokenMove` so it hasn't come up?

cwisniew:
> I wouldn't really recommend structuring things that way just from a separation of concerns perspective, but there's no technical reason we can't have a list of all the matching macros included in the JSON that gets passed in. Even without it, the handler macros could go find out on their own in the macro if they really wanted to.
> 
> > They each manage init because they each need to know the order of tokens in combat. How does a GM combine the two?
> 

Imagine the case where someone creates a lib token with spell libraries and hooks into the initiative for the duration aspect of the spells. And then someone comes along with "initiative announcer 3000" which plays a sound when its a players turn and a GM thinks, well that is useful my players are always zoning out. So GM installs this lib token but now suddenly the spell framework doesnt work.

Its not just about the people making the frameworks also have to consider those who want to use them too.

selquest:
Updated to split the `onInitiativeChange` event into two parts as discussed above.  I used the names `onInitiativeChangeAttempt` and `onInitiativeChangeCommit`, which are kind of a mouthful but I wasn't coming up with anything more succinct without being confusing.  Definitely open to suggestions there.

Each of these events now supports the presence of multiple handlers, but only the `onInitiativeChangeAttempt` handlers have veto power.  I've attached a sample campaign with a couple of dummy handlers for demonstration purposes - the _Lib:Initiative_ token denies any attempt to move `NEXT` past a PC token, and the Lib:other token offers the world's most annoying handler that asks "Are you sure?" any time you try and make a change.  In the `onInitiativeChangeCommit` handler, _Lib:Initiative_ toggles some states to highlight the active token.

[InitiativeTest.cmpgn.zip](https://github.com/RPTools/maptool/files/5164298/InitiativeTest.cmpgn.zip)

If we like this approach, I will want to complete some refactoring to keep `onTokenMove` and `onCampaignLoad` using the same infrastructure as these, simplifying future changes.  I didn't want to get too much into that while the change is still up in the air though.

cwisniew (approved):
> Reviewed 2 of 5 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2190)* status: :shipit: complete! all files reviewed, all discussions resolved

cwisniew:
> Updated to split the `onInitiativeChange` event into two parts as discussed above. I used the names `onInitiativeChangeAttempt` and `onInitiativeChangeCommit`, which are kind of a mouthful but I wasn't coming up with anything more succinct without being confusing. Definitely open to suggestions there.
> 
> Each of these events now supports the presence of multiple handlers, but only the `onInitiativeChangeAttempt` handlers have veto power. I've attached a sample campaign with a couple of dummy handlers for demonstration purposes - the _Lib:Initiative_ token denies any attempt to move `NEXT` past a PC token, and the _Lib:other_ token offers the world's most annoying handler that asks "Are you sure?" 

If you think this is the world's most annoying "Are you sure?" dialog you have never tried to uninstall IE from windows...



> If we like this approach, I will want to complete some refactoring to keep `onTokenMove` and `onCampaignLoad` using the same infrastructure as these (_edit: but not adding support for multiple onTokenMove veto handlers unless done as part of a separate issue, since that's been operating that way for a long time_), simplifying future changes. I didn't want to get too much into that while the change is still up in the air though.

This works well, as all the lib tokens get a chance to veto the change and no initiative changed event occurs unless everyone is happy. I am ok with a refactor of onTokenMove if you think you can fit it in with this change to consolidate the code for the two events, but yes breaking out the veto for that is probably too late as it will end up breaking a lot.

PR #2193, opened 2020-09-02
Implement 'getArea' for remaining Template drawings

### LINKED ISSUE #2192: Null pointer exception in 'movedOverDrawing' when moving over a template
**Describe the bug**

Calling the `movedOverDrawing` macro function for a template drawing + a path that intersects it results in a null pointer exception.

**To Reproduce**
1. Open maptool on the default 'Grasslands' map.
2. Turn on 'show coordinates', draw a radius template at 0,A
3. In the draw explorer, get the drawing GUID
4. Run the macro command: `[r: movedOverDrawing("Grasslands", "<GUID from step 3>", '[{"x": 0, "y": 0}, {"x": 50, "y": 50}]')]`

Observe a null pointer exception like so: `java.lang.NullPointerException: Cannot read field "curves" because "rhs" is null error executing expression movedOverDrawing("Grasslands", "<GUID>", '[{"x": 0, "y": 0}, {"x": 50, "y": 50}]')`.

**Expected behavior**

It should return the coordinates of the path since the path does overlap.

**MapTool Info**
- Version: current develop commit (cd482937)

 - OS: NixOS

**Additional context**

This was also reported by another user in the forum: https://forums.rptools.net/viewtopic.php?f=21&t=28824

The source of this NPE is that `movedOverDrawing` uses `getArea` to calculate path intersection, but the RadiusTemplate `getArea` returns null:

https://github.com/RPTools/maptool/blob/6d193e3f5cc635aed30730ec09fde2a210870054/src/main/java/net/rptools/maptool/model/drawing/RadiusTemplate.java#L128-L131

Most (all?) of the other template types don't implement `getArea` either.

I've already started working on this, and I expect to have a pull request before too long.

euank:
Most Template drawings did not have a getArea implementation.

This adds a straightforward implementation to each of them.

In general, the pattern for creating these was by mimicking the 'paintArea' code in
order to calculate the set of squares that would be drawn, and then
rather than paint them, merge them all into one area.

There's room for refactoring this a bit more in order to remove the
duplication between 'getArea' and 'paintArea'.
One idea would be to add a primitive to the template drawings of the
list of squares it covers, implement just that for each template, and
then have AbstractTemplate know how to implement everything else from
that.

For now, it seems pragmatic to do it with this amount of duplication and
to tackle that refactoring separately.

Fixes #2192

I do want to get back to this and do the refactor above at some point since, well, the template drawings all can really have some nicer code since they all have to be grid aligned / relatively geometric shapes.
I thought it was worth getting this up sooner though so we can merge it if we want to do the above refactor separately, or for any suggestions if there's an easier alternative path to this.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2193)


cwisniew (approved):
> Reviewed 10 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2193)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2148, opened 2020-08-03
Correctness issues

ebudai:
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2148)

PR #2289, opened 2020-10-31
Issue 2287

irisiflimsi:
NPE remedy taken from main branch (modified). I don't know where the other commits came from, they were already in my develop!?

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2289)


Phergus:
Looks to me like you created your working branch off of main (or something) instead of develop.

In any case, the changes made in #2148 should have created the instance so whatever went wrong with those needs to be fixed.

PR #2294, opened 2020-11-04
Issue 2284

irisiflimsi:
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2294)


Phergus:
I think this has been rendered obsolete by the discussion in the issue.

irisiflimsi:
It has? The discussion was about further improvements by switching to SVGs, but after some initial enthusiasm has died. The original proposal to change the text to icons would have been a valid intermediate step.

Issue #755, opened 2019-10-01
Scroll bar not always showing when needed in Image Pane of Resource Library


Phergus:
**Describe the bug**
If only a single row of thumbnails is partially outside of the window, the scrollbar doesn't appear and access to the partial thumbnails is difficult.

**To Reproduce**
1. Open a folder in the Resource Library with only a dozen or so tokens.  The Default -> Tokens folder will work.
2. Resize Resource Library window such that a couple of the tokens are partially out of the view.
3. Note lack of scrollbar.

**Expected behavior**
Expect scroll bar to appear as soon as one or more images are partially out of view.

**Screenshots**
![Capture](https://user-images.githubusercontent.com/34379254/65994770-5c16e900-e451-11e9-9085-49122b670584.JPG)

**MapTool Info**
- Version: 1.5.5-rc.2
- Install: New

**Desktop (please complete the following information):**
 - OS: Windows
 - Version; 10

**Additional context**
Users can force the scroll bar to appear by resizing window smaller or using the zoom control.

xdy:
It seems to be a bug in JScrollPane.
Even setting ScrollPaneConstants.VERTICAL_SCROLLBAR_ALWAYS doesn't make it useful, the scrollbar appears but is greyed out until one uses the same workaround as in the 'Additional context' above.
Writing our own JScrollBar might help, but, not likely to be worth the effort.

Phergus:
I don't think it is worth the effort to do that.  It's an annoyance but not critical.  We'll file this one away.

dluga93:
Turns out the problem is a miscalculation of rowCount in ImagePanel.java.
E.g. 7 images, 2 images per row -> rowCount should be 4, but it was actually 3.

Pmofmalasia:
Seems to be solved for me on 1.13.2 - single row is partially outside of the window with a scroll bar present in the screenshot: 
![image](https://github.com/RPTools/maptool/assets/79408871/fca7d330-e7e3-4c00-bc0e-7d3be7df4757)

PR #2301, opened 2020-11-09
Fix rowcount calculation

dluga93:
This change will fix issue [#755](https://github.com/RPTools/maptool/issues/755).
When scrolling through the image pane, the last row of images might not be fully displayed, even when the scroll bar is at the bottom (or in the case of the screenshot in the issue, even when the scrollbar does not appear at all).

The problem was a miscalculation of the number of rows in the pane.
E.g. 7 images, 2 images per row -> rowCount should be 4, but it was actually 3.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2301)

### DIFF
--- a/src/main/java/net/rptools/lib/swing/ImagePanel.java
+++ b/src/main/java/net/rptools/lib/swing/ImagePanel.java
@@ -392,7 +392,8 @@ public class ImagePanel extends JComponent
     if (width < gridSize + gridPadding.width * 2) {
       rowCount = model.getImageCount();
     } else {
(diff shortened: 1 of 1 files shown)


Phergus:
Do me a favor and go comment on #755. Thanks.

Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2301#-:-MLjLjFGBDm-yupX98ly:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2304, opened 2020-11-15
Add shortcut key F to flip selected tokens horizontally

MeloDub:
Hey!
I added this feature because I've seen many people asking for a faster way to flip tokens horizontally for RP purposes. To use this, just hit the F key while having one or more tokens selected with the pointer tool.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2304)


Phergus:
Just FYI, there must be an issue opened for a PR to be accepted.  As long as the F key isn't already tied to something I don't think there would be a problem with accepting this change.

PR #2324, opened 2020-11-30
Issue #2314

### LINKED ISSUE #2314: Refactor/clean up chat output generation
Presently, some of the code for generating the HTML output for chat messages is duplicated across multiple places resulting in each needing to be updated if changes are made. This could be resolved by having everything for generating the final output in one place that other bits of code would call out to. This would help prevent issues like #2177 and make things like #2273 simpler.

This would also be a good opportunity to clean up the output itself a little, using CSS classes where it makes sense. This would make it simpler to tweak things, and I recall it being mentioned that the idea of user editable CSS for the chat has been floated around so this could be a step towards that as well.

To be precise, I propose:
1. Moving code for formatting different types of messages to separate methods (something like `getFormattedSay()`, `getFormattedEmote()`, etc.) and putting them all in a single class (`MessageUtil`? Something else?) The biggest offender is probably Say messages, but I feel like it would probably make sense to include simpler things and stuff that is only used in one place for consistency, and in case that ever changes. (Likely ones include /say, /emote and /emotes, /emit, /ooc, /whisper, /gm, /self, and system messages (connected, disconnected, etc.).)
2. Removing inline styling and tags like `<b>`, `<i>`, etc. and replace them with CSS classes to be used by MessagePanel.

Irarara:
Refactored various bits and pieces of message formatting code into the class `MessageUtil`. This class also defines the CSS classes used for output. This should close #2314. 

Due to making the output more uniform and using more CSS there are a few small changes to output, for instance:
* `"self"` messages from an a macro link are now styled the same as `/self` (italicized).
* The width of the avatar image cell is slightly bigger, but it's guaranteed to be a uniform width regardless of whether there is a token image/portrait present and regardless of whether the image is 40px in width.
* `"all"` output from a macro link would previously use the player name instead of current identity if there was no token, now it uses whatever the current identity is (same as `/say`).

Notes:
* valign attribute is still necessary (for say output) because `vertical-align: top` isn't supported.
* "To gm" messages (ie those from `/gm`, etc.) are wrapped in a div with the class `to-gm`, but there are no rules defined for this class because presently they aren't formatted in any way. It just felt appropriate to include it for completeness.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2324)


Phergus:
Your commit has a spurious log file in it (because something is whacked in the build process currently).

Irarara:
Whoops, my bad. That one slipped through, should be fixed now. 😅

Phergus (approved):
:lgtm:

> Reviewed 15 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2324#-:-MONPyg_1k2j5-iKxEbS:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2337, opened 2020-12-12
Issue #2092

### LINKED ISSUE #2092: onChangeSelection events occuring when no change of selection has occurred.
**Describe the bug**
Changing a token via macro causes both an `onChangeToken` event and a `onChangeSelection` event.

**To Reproduce**
1. Load attached campaign file and click on Overlay campaign macro.
2. Right-click on Elf token and impersonate. (Generates an `onChangeSelection` event.)
3. Change a property of the token, e.g. `[r: Strength = 3d6]` or `[r: setVisible(0)]`, from chat.
4. Note an `onChangeSelection` event fires along with an `onChangeToken` event. Token remains selected.

**Expected behavior**
An `onChangeSelection` event that is triggered without an related mouse action is problematic.  Could easily create an infinite loop situation.  Actually, performing step 3 under MT 1.7.0, does exactly that but has been fixed in 1.8.0 develop.

**MapTool Info**
- Version: 1.7.0, current develop

 - OS: Windows
 - Version: 10

**Additional context**
Remove `.zip` from name.  Be prepared for errors if used under MT 1.7.0.  See #2078 .
[OverlayOnChangeSelectionMT170.cmpgn.zip](https://github.com/RPTools/maptool/files/4920186/OverlayOnChangeSelectionMT170.cmpgn.zip)

Irarara:
Fixes #2092

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2337)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLFrameFactory.java
+++ b/src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLFrameFactory.java
@@ -16,11 +16,9 @@ package net.rptools.maptool.client.ui.htmlframe;
 
 import java.util.Collections;
 import java.util.List;
-import java.util.Set;
 import net.rptools.lib.AppEvent;
 import net.rptools.lib.AppEventListener;
 import net.rptools.maptool.client.MapTool;
-import net.rptools.maptool.client.ui.commandpanel.CommandPanel;
 import net.rptools.maptool.language.I18N;
 import net.rptools.maptool.model.*;
 import net.rptools.maptool.model.Zone.Event;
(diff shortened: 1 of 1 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2337#-:-MONntUQCDm9e2SyWWhb:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

Phergus:
@Merudo  Any thoughts on this?

Merudo:
@Phergus the previous behavior, established before I joined, was to fire `onChangeSelection` and `onChangeImpersonated` both if the selection / impersonation changed, and also if the current selected / impersonated target itself changed.

With this new commit, these events will no longer fire once a selected or impersonated token itself changes. This could potentially mean that some character sheet or overlay could become out of date after token changes. Instead, `onChangeToken` should be used to keep frame/dialog/overlay updated.

Phergus:
Okay.  Will need further review on this as the previous behavior was creating infinite loops and spurious change selection events.

Phergus:
@Merudo I'm still seeing `onChangeSelection`, `onChangeToken` and `onChangeImpersonated` events in Overlays.  Did you test this with current code?

Irarara:
I think they meant that the change in behavior will likely break macros that used the old (and may I add, as far as I can tell, undocumented) behavior, which is true. 

If this change stands, it should probably be highlighted that anyone using `onChangeSelection` or `onChangeImpersonated` expecting them to fire when a selected/impersonated token changed will need to use `onChangeToken` now.

Phergus:
Got it now.

IMO, it's working correctly now and it will be much better if folks just update their dialogs/frames/overlays to use the correct events going forward.

PR #2344, opened 2020-12-14
Issue #2343

### LINKED ISSUE #2343: Trusted prefix colors will not show up in client and changing trusted prefix colors doesn't work properly
**Describe the bug**
These are really two entirely separate issues, but they warrant being fixed together.
1. The CSS used for the trusted prefix does not in work within Maptoool (Swing does not seem to support the element.class selector, despite documentation not seeming to indicate this!).
2. Only the red component from the color picker is set when changing the "Trusted Prefix Background" and "Trusted Prefix Foreground" settings.

**To Reproduce**
For 1:
1. Create a campaign macro that just outputs some text.
2. Start the server, joining as a player.
3. Click the campaign macro that was created. Observe that you can mouse over your name and see the source of the macro but the trusted prefix colors are not applied.
(Bonus, if you save the chat log, the colors will show up when viewed in a browser).

For 2:
1. Open up preferences.
2. Open the color picker for "Trusted Prefix Background" and go to the RGB tab.
3. Enter 0 for all the RGB values.
4. Confirm out of the dialog and then out of preferences.
5. Open preferences again and go back into the color picker for "Trusted Prefix Background" and check the RGB tab. Only the red value will be 0.
6. Repeat the above with "Trusted Prefix Foreground".

**Expected behavior**
The trusted prefix color should be visible and colors should be properly set.

**MapTool Info**
- Version: Development branch

 - OS: Windows
 - Version 10

Irarara:
Change to the CSS used for trusted prefixes so it works within the client and allows the colors to be properly changed. Fixes #2343

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2344)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppPreferences.java
+++ b/src/main/java/net/rptools/maptool/client/AppPreferences.java
@@ -498,8 +498,8 @@ public class AppPreferences {
 
   public static void setTrustedPrefixBG(Color color) {
     prefs.putInt(KEY_TRUSTED_PREFIX_BG_RED, color.getRed());
-    prefs.putInt(KEY_TRUSTED_PREFIX_BG_RED, color.getRed());
-    prefs.putInt(KEY_TRUSTED_PREFIX_BG_RED, color.getRed());
+    prefs.putInt(KEY_TRUSTED_PREFIX_BG_GREEN, color.getGreen());
+    prefs.putInt(KEY_TRUSTED_PREFIX_BG_BLUE, color.getBlue());
   }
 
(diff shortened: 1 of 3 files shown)


Phergus (approved):
:lgtm:

> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2344#-:-MOXpJ4w1rKY9sBLFb5-:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2350, opened 2020-12-16
Feature flip token #2349

### LINKED ISSUE #2349: Add shortcut key to flip tokens horizontally
I've seen many people asking for a faster way to flip tokens horizontally for RP purposes as doing that by right-clicking could be a little frustrating, especially if it's done very often.

Hit the F key while having one or more tokens selected with the pointer tool to flip them horizontally.

MeloDub:
Fixes #2349

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2350)


Phergus:
Instead of using the heavy `putToken()` function recommend this be done using `updateTokenProperty()` and only update just that property of the token.  See ServerCommand, ServerCommandImpl and Token.Update enum.

MeloDub:
I can't see any command related to flipping the token in `Token.Update`, but I managed to do it by creating a new command `flipX`. Is this the correct way to do it or did i miss an already existing property that i should have used?

By the way, while I'm at it, should I also change the implementation of the flip function in `AbstractTokenPopupMenu` to use the new command?

Phergus:
It's been an ongoing process to move token changes/updates to this new form and away from `putToken()`.  So, in answer to your question, yes, add a new flipX command and it would be nice to also change the popup menu at the same time.

cwisniew:
Moving to 1.9.0 as next beta for 1.8 is due in next couple of days and we want to concentrate on getting out the door. (just fixing some bugs)

Phergus (approved):
:lgtm:

> Reviewed 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2350#-:-MYm7T4ABseg7Iga0F_f:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @MeloDub)

Issue #2352, opened 2020-12-18
Add Macro Editor Syntax Highlighting for Campaign Properties


selquest:
**Is your feature request related to a problem? Please describe.**
When editing macros, direct edits to defined token properties is very convenient - but sometimes trips people up when they don't realize they've used a defined property as a local variable name.

**Describe the solution you'd like**
Defined properties can be highlighted in the macro editor to better distinguish between them and normal local vars.

**Additional context**
Ideally this syntax highlighting would be accompanied by context-dependent tooltips and/or other indications... but that would require additional functionality in the Macro Editor, whereas syntax highlighting is low-hanging fruit.

selquest:
*Before*:
![Before - No highlighting for token properties](https://user-images.githubusercontent.com/2325887/102669513-ca31f380-415c-11eb-9041-b0e199c99829.png)
*After*:
![After - Token property highlighted in macro editor](https://user-images.githubusercontent.com/2325887/102669529-d1f19800-415c-11eb-801c-49d2f676e77f.png)

Phergus:
Cool.  I was thinking about the same thing earlier when the discussion came up on Discord.

selquest:
Just checking in to see if there are any particular concerns holding up the PR, or if things are just delayed at the moment?  Making sure it doesn't slip through cracks.

Phergus:
We're on a bit of a code-freeze while there is some core infrastructure being worked on.

bubblobill:
Just one thing @selquest. You are too subtle. Any chance you could alter the highlight from orange colour to super-bold strikeout red flashing at the correct rate to induce seizures.

bubblobill:
Q: Does this pick up properties from all defined property types?

selquest:
> Q: Does this pick up properties from all defined property types?

Yes.  As mentioned in the PR description, I briefly explored the idea of restricting highlighted properties to those applicable to the actual type of the token holding the macro, but there are enough downsides to doing so (Lib:macros, `switchToken()`, Apply to Selected, etc.) that I didn't think it would actually be desirable... and it would require far more disruptive changes to make Type information available at the right spot in the current code path anyway.

> Any chance you could alter the highlight from orange colour to super-bold strikeout red flashing at the correct rate to induce seizures.

[...]

PR #2353, opened 2020-12-18
Add macro editor syntax highlighting for defined token properties.

selquest:
Adds syntax highlighting for token props.

Couple of reservations with this particular implementation:
1. not particularly efficient - a new instance of `MapToolScriptSyntax` is instantiated each time a macro editor dialog is opened, so it's not ideal to have to re-add all defined properties each time that happens.  It already does this with reserved words, special variables, etc. though so this isn't really a new problem.  Micro-optimization of this specific chunk isn't really worth the added complexity, especially without knowing what plans might exist for revised syntax highlighting approaches in MTScript2.
2. New properties _created_ during a given session will be properly highlighted the next time a macro editor dialog opens, but because the `macroFunctionTokenMap` is never cleared any properties that are _removed_ during a session will continue to be highlighted until the next restart.  A similar limitation already applies to highlighting of UDFs, so I don't see this as especially problematic.

I also toyed with the notion of restricting highlighted properties to those applicable to the actual type of the token holding the macro (for token macros), but the current code path doesn't provide that opportunity if we wanted to... and it would get complicated around things like Lib:macros anyway, so I abandoned that idea.

Fixes RPTools/maptool#2352

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2353)


cwisniew (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2353)* status: :shipit: complete! all files reviewed, all discussions resolved

cwisniew:
Moving to 1.9.0 as next beta for 1.8 is due in next couple of days and we want to concentrate on getting out the door. (just fixing some bugs)

selquest:
@Phergus any chance we can get this in 1.8.5?  I know that's still a "bugfix" release, but this is such a small change for being a pretty big QoL improvement... kind of a shame to keep putting it off.

Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2353#-:-MYIiDP67sE3tpcUP-13:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @selquest)

PR #2401, opened 2021-02-19
Create BarImageFunction to resolve #1234

### LINKED ISSUE #1234: getBarImage(barName[,value]): assetIds
Unable to get the assetId of images from Bars.

Create a function like getStateImage, but for bars; getBarImage(barName[,value]) where value is the value of bar that returns a specific image. So, 0 is empty bar, 1 is full bar and values in between would be the increments. Not specifying the value (or -1) will return all the assetIds in an array.

Not all bars are image or multi-image types, so return "" if invalid or "[]" if no value is specified.

I use multi-image bars a lot and I was designing a user input to set the values and realized I would need to put the images in a table of something if I wanted to use them in the input rather than accessing them directly from the bar.

adyoungblood:


Phergus (approved):
:lgtm:

> Reviewed 2 of 4 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2401#-:-MYtrWBmDEX03Ja2WToL:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adyoungblood)

PR #2404, opened 2021-02-20
Create functionality for #369 (move macros instead of copying)

### LINKED ISSUE #369: [Request] Move macro button instead of copy
Current, if one drags a macro button from one group to another, it is copied. This is fine, I am sure. It would be nice though, if it was possible to move it by dragging as well. 
E.g. shift-drag would move and drag would copy as it always has.

adyoungblood:
A bunch of automated translation commits got caught up in this branch, but [9654d9f](https://github.com/adyoungblood/maptool/commit/9654d9fd4b9e2ff3f25cf6087abde19376eea97f) is the only relevant commit to the issue. Resolves #369, with Ctrl + drag working to move a macro instead of copying it.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2404)


src/main/java/net/rptools/maptool/client/ui/macrobuttons/buttongroups/ButtonGroup.java
```diff
@@ -121,18 +121,31 @@ public void drop(DropTargetDropEvent event) {
...
       if (panelClass.equals("GlobalPanel")) {
         event.acceptDrop(event.getDropAction());
         tempProperties.setGroup(
             getMacroGroup()); // assign the group you are dropping it into, rather than the original
         if (!tempProperties.isDuplicateMacro("GlobalPanel", null)) {
+          if (event.getDropAction() == 1) { // DnDConstants says 1 is copy,
+            // but this is for reversed behavior (copy as default)
+            getArea().getPropertiesList().remove(oldMacroProps);
+            getArea().repaint();
+          }
```
emmebi, line 136: Suggestion:
- extract this piece of code into a small private method which can be re-used instead of repeating the same behavior
- use a constant for 1 (as you are adding a comment, make a constant instead)

adyoungblood: Don't know why I didn't do that first one, thanks! As for using the provided constants, I originally chose not to because it has counter-intuitive meaning (ACTION_COPY = 1, but the if statement is meant to be "if the token was moved"). This is a thing because when createDefaultDragGestureRecognizer is set to recognize ACTION_COPY_OR_MOVE in MacroButton, the default drag action becomes moving, with a dropAction code of 2. Holding Ctrl while dragging gives a dropAction code of 1. I'm not sure why this is, but I'm also quite unfamiliar with Swing, so it could be something obvious that I'm not aware of!

emmebi: I am not very much familiar with Swing as well, so take my advice with caution :) 

Looking at DropTargetDropEvent.java, it looks like pressing "Shift" (rather thank Ctrl) should generate an ACTION_MOVE event, which would be a little better. I think you could get rid of the comment at that point.

adyoungblood: Right now, the GestureRecognizer in MacroButton interprets a non-modified drag (no keys pressed) as ACTION_MOVE, and shift appears to do nothing. I'm really not sure why, and it changed after I had the GestureRecognizer recognize ACTION_MOVE_OR_COPY instead of ACTION_COPY. I'll dig around in a bit and see if I can maybe get it to behave as the documentation says it should (I can't get an ACTION_LINK event while dragging, for instance). In the meanwhile, it is (externally) working as intended.

Azhrei:
I'd like to actually test this on macOS before signing off on it, but my Java setup is borked and I need to reset it.  What's the play here, Phergus — merge it and let others test?

Phergus:
@Azhrei  It pretty much has to be you or some other community dev to test it out on a Mac or we merge it in and leave the Mac testing until we have a build.   How does the Mac differentiate between I'm moving this item by dragging versus I'm copying this item by dragging?

Azhrei:
In the file manager (Finder), holding down Alt changes a move operation into a copy.  However, the Mail client uses either Alt or Cmd as a copy (when moving messages between mailboxes, for example).  I'd go with Alt.

adyoungblood:
If it helps anyone, the relevant function to modify to change which keys do what is **deleteOriginalMacroIfMove**, in the ButtonGroup class. The transferModifiers field of oldData is 1024, plus certain numbers depending on which keys were pressed (Shift adds 64, Ctrl adds 128, and Alt adds 512, so Shift + Alt would be 1600). Any modifications would be a matter of adding/changing the if statements to correspond to the keys you want, and putting the correct behavior in them.

Phergus:
So most likely a check for Mac OS and then use Alt for copy on Mac and Ctrl for copy on the rest.

adyoungblood:
If you see my [message](https://cdn.discordapp.com/attachments/494598206626463755/832841102037876736/unknown.png) in the Discord, this solution isn't amazing, but I suppose I'll know if a bug report trickles in later.

Phergus:
I don't know that there is a way to do it that isn't hackish so we can go with that and if changes are needed during testing we can do so.

Phergus (approved):
:lgtm:

> Reviewed 12 of 13 files at r2, 5 of 6 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 1 of 1 files at r8.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2404#-:-MYoGe-Z1mHCxxNGP5u1:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Azhrei, @cwisniew, and @emmebi)

Phergus (approved):

Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2404#-:-MYoH1NK4U_zDWCGTDCR:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @emmebi)

Phergus:
@emmebi you need to resolve your comments on this.

emmebi (approved):

Phergus:
@emmebi  when you comment on the code you then have to follow up by resolving it or it will stay blocked.

image

Note the _pending_ above.  Click on Details to go to it.

emmebi (commented):
> Reviewed 12 of 13 files at r2, 4 of 6 files at r5, 1 of 1 files at r7, 1 of 1 files at r8.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2404)* status: all files reviewed, 1 unresolved discussion (waiting on @emmebi)

emmebi (commented):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2404)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adyoungblood)

PR #2450, opened 2021-03-10
Issue #2449

### LINKED ISSUE #2449: Text color applied to <p> elements is sometimes overriden in chat messages
**Describe the bug**
In 1.8.3,  a message (specifically, /say, /emote, or /emotes) with a `p` element that has a text color set (ex. `<p style="color: blue">Test</p>`) will sometimes show up in the chat color set by the user, instead of what is set by the style attribute. This was not true in earlier versions. 

Additionally, this doesn't happen if it's preceded by certain other elements, like `div` or `h1`.

**To Reproduce**
1. Enter `<p style="color: blue">Test</p>` into the chat (or, if your chat color is also blue, change blue to something else)
2. Observe that the output "Test" shows up in the chat color instead.

**Expected behavior**
The text in the `p` element should be the color specified by the style attribute.

**Screenshots**

**MapTool Info**
- Version: 1.8.3
- Install: New

 - OS: Windows
 - Version: 10

**Additional context**
This is a result of a change made in (#2314) and applies to messages where chat color is used (/say, /emote, /emotes). Simply, it was changed so that chat color is inserted into messages with a `font` element instead of a `span`, which didn't seem like it should really make a difference, but for `p` elements it can cause their formatting to be overridden.

Writing this much feels kind of silly since this is such a small issue, but for the sake of documenting it, and because the weird way Swing handles HTML has been a subject of personal curiosity, here's my best understanding of things for anyone else curious:

Swing applies some odd seeming transformations to HTML input to make it the way it likes. Most relevant here, spans with certain inline CSS get transformed into a `span` with a nested `font` element, and certain formatting will be moved to `font`. For example, `<span style="color:red">Text</span>` becomes `<span><font color="red">Text</font></span>`. This is why I made the change to begin with, because in theory it seemed like it should be functionally the same (and cuts out some seemingly pointless spans).

Another thing to note is Swing doesn't really allow nested `font` or `span` elements. Instead of them staying nested it essentially chops them up. So, `<font color="red">Red <font color="blue">Blue</font> Red</font>` becomes `<font color="red">Red </font><font color="blue">Blue</font><font color="red"> Red</font>`.

Finally, for some reason if you have `p` element inside a `font` element, the following transformation can happen: 
`<font color="red">Outer <p style="color:blue">Inner</p></font>`  → 
`<font color="red">Outer </font><p style="color:blue"><font color="red">Inner</font></p>`. 
The content of the `p` is wrapped in the outer `font` element, and therefore its formatting takes precedence. Hence this issue.

Why this doesn't happen if it's a `span` instead, despite the transformation that gets applied to them, is kind of a mystery to me, but it might be due to some kind of precedence. `font` might take precedence over `p`, and therefore it tries to preserve its formatting, whereas `p` takes precedence over `span` and acts like a boundary, cutting it off there. This would also explain why (most? all?) other elements aren't affected by this, they have a higher precedence than `font` and therefore also act as a boundary. This is why if `p`  follows any of these other elements the issue doesn't occur, and also the same reason the color of a message resets to black following them.

**TL;DR:** Swing is weird and if you wrap a `p` in a `font` element the formatting applied by the `font` can override whatever formatting is applied to the `p` directly. (To illustrate this, this will display "BLUE" in red in the chat panel of _any_ recent version of Maptool, but should be blue in say, a frame5: `<font color="red"><p style="color:blue">BLUE</p></font>`.)

As a final aside, there's a case to be made for this behavior actually being nice, because now `p` elements preserve chat color where they didn't before (previously, content of a `p` would show up in black and any text after would also be black). The text color can still be changed if you use `span` or `font` inside the element. That said, the behavior is counterintuitive.

Irarara:
Use span instead of font for chat color. Fixes #2449

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2450)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2450#-:-MVWUMDnDN8402rFdXgX:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Irarara)

PR #2459, opened 2021-03-13
Allow current dev play with 1.8.3

thelsing:
The current develop branch can't connect to a game hosted by 1.8.3.
This is a fix for that (#2461).

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2459)


Phergus:
PRs are only accepted if there is an open issue for the change.

cwisniew:
The code change from 1.8.3 to dev was intentional, unfortunately Dev won't work with 1.8.3 because of this. This just reintrodues. But back into the handshke

Phergus:
As change would reintroduce a bigger problem this is closed.

PR #2465, opened 2021-03-16
Add support for hyperlinks to HtmlPanelRenderer

### LINKED ISSUE #1420: Clickable links on hidden layer popups
I've seen this discussed every so often, but couldn't find an issue linked to it. I may have overlooked something, and apologize if that's the case.

I find the popup window that appears when you are in the Token layer and click a token in the Hidden layer to be one of the most useful features of Maptool. However I think that usefulness would be greatly improved if you could add working links - especially macroLinks - to that popup.

A scenario I envision is, for instance, a locked door or a trap. If you place a marker on the hidden layer you can bring about a popup with info (e.g. for D&D the DC to pick the lock or find the trap, the consequences, etc.). It would be great if there was a clickable link to run a macro to mark the door as unlocked, or the trap as activated, or a door as opened, setting or clearing VBL, automatically rolling damage etc. I can see lots of potential uses, and I get the impression it wouldn't be that difficult to implement. This would remove the need to switch between the Token and Hidden layers to initiate macros, which helps preserve the pace of the game.

I've seen some alternatives suggested, the most practical one being a docked window and a macro that updates on mouseover to get a similar effect, but this is, in my mind, less intuitive than a link and creates some accidental macro update problems.

thelsing:
Fixes #1420

I had this one laying around.  

I personally don't use HtmlPanelRenderer any more and replaced the hover with 

`HTMLFrameFactory.show(token.getName(), HTMLFrameFactory.FrameType.DIALOG,  true, "height="+ MapTool.getFrame().getHeight() * 8 / 10+";width="+ MapTool.getFrame().getWidth() * 5/10 , createHoverNote(token));`

in my current build, but the fix in the pull request is quite simple so maybe somebody finds this useful.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2465)


cwisniew (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2465)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @thelsing)

PR #2481, opened 2021-03-19
Issue #2480

### LINKED ISSUE #2480: Inline Expansion of URLs doesn't stop at whitespaces
**Describe the bug**
Presently, when URLs are turned into clickable links in the chat whitespace is not treated as a boundary, meaning if there's a URL in a message (almost) any text after the URL also gets incorporated into it. Not only does this behavior result in a link to a bad URL, but clicking on a link that includes a space in the URL simply doesn't work (though, this itself is a separate issue).

**To Reproduce**
1. Enter `https://www.rptools.net/ some text` into the chat
2. Observe that "some text" gets incorporated into the link, and that clicking on the link will pop up an error and fail to open the link in a browser.

**Expected behavior**
Whitespace characters act as boundaries for URLs.

**MapTool Info**
- Version:  Current develop (and virtually every past version I've used, I think)
- Install: Any

 - OS: Windows
 - Version 10

Irarara:
Changes the regex used for inline expansion so it doesn't capture whitespace. Fixes #2480

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2481)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/commandpanel/MessagePanel.java
+++ b/src/main/java/net/rptools/maptool/client/ui/commandpanel/MessagePanel.java
@@ -223,7 +223,8 @@ public class MessagePanel extends JPanel {
           // output = output.replaceAll("(^|\\s|>|\002)(https?://[\\w.%-/~?&+#=]+)", "$1<a
           // href='$2'>$2</a>");
           output =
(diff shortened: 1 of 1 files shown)


Azhrei (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2481)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Irarara)

Issue #2396, opened 2021-02-19
JSON format token properties error silently when used by other properties.


Akaku-Keninari:
**Describe the bug**
Token properties in JSON format attempt to parse, fail silently, then are treated as strings correctly when referenced by other token properties.

**To Reproduce**
1. Create a token with a property named `firstProperty` that is `{"A": 1, "B": 2}`
2. Create a displayable property on the same token with a value of `{json.get(firstProperty, "A")}`
3. Mouse over the token and see that it displays "1"
4. Check debug log, you should see errors like this:
```
2021-02-18 23:56:34.086 DEBUG net.rptools.maptool.model.Token - Evaluating property: 'Elevation' for token EXAMPLE(4F71C395E5004F778BFB58D9D1357590)----------------------------------------------------------------------------------
2021-02-18 23:56:34.090 DEBUG net.rptools.maptool.client.MapToolLineParser - json.get(Strength, "A")
2021-02-18 23:56:34.105 DEBUG net.rptools.maptool.model.Token - Evaluating property: 'Strength' for token EXAMPLE(4F71C395E5004F778BFB58D9D1357590)----------------------------------------------------------------------------------
2021-02-18 23:56:34.105 DEBUG net.rptools.maptool.client.MapToolLineParser -  "A": 1, "B": 2
2021-02-18 23:56:34.107 DEBUG net.rptools.maptool.client.MapToolLineParser - net.rptools.parser.ParserException: line 1:4: unexpected char: ':'
```

**Expected behavior**
It shouldn't spend the time to error out silently. This time adds up on frameworks that have extreme amounts of jsons and calls to other properties in their statsheets.

**MapTool Info**
- Version: 1.8.3
- Install: New

**Desktop (please complete the following information):**
 - OS: Windows
 - Version 10

**Additional context**
Attached is zip file that's actually a .cmpgn file set up already.

[json error.zip](https://github.com/RPTools/maptool/files/6008477/json.error.zip)

Link to a campaign that slows to a crawl because of this:
https://drive.google.com/file/d/1Z3iGg0oKnAJV6vcCxmpoAmhC7BTiiuha/view

Getting the json with `getRawProperty()` bypasses this error but has limitations (will not pull defaults) and is messy to hack into existing code.
Messy hack to 'fix' the issue on a token in linked campaign:
`[h: names = getAllPropertyNames(getPropertyType(),"json")]
[foreach(prop, names, "<br>"), code:
{
	[r: prop]
[...]

Issue #1560, opened 2020-04-15
Campaign default property bugged with {} assignment


Merudo:
**Describe the bug**
Setting up a default property value with `{}` assignment doesn't work anymore since 1.5.12.

**To Reproduce**
1. Set the following default property:
```
var1:10
*var2:{var2=var1}
```
2. Put your mouse over a token. You will see `{"var1":"var2"}` in the stat sheet.
3. In 1.5.10, the stat sheet shows `10`

Note that the bug doesn't occur if there is no variable assignment within the `{}`. So, 

```
var1:10
*var2:{var1}
```
works fine.

**Expected behavior**
MapTool returns the value, not the assignment code.

**MapTool Info**
- Version: 1.5.12+
- Install: New

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: 10

bubblobill:
Does this work with {var2}={var1} or [r:{var2}={var1}]?

Merudo:
Don't work:
`*var2:{var2}={var1}`, creates a loop

Work: 
`*var2:[r:{var2}={var1}]`
`*var2:[r:var2=var1]`

Phergus:
This didn't really work in 1.5.10 either.   If you have your properties defined as in the OP but no default for `prop1` then when you open the Edit Token dialog to put in a value for `prop1` the formula in `prop2` is set to empty.  Now it doesn't matter what `prop1` is.

Merudo:
Discussion on the forum: https://forums.rptools.net/viewtopic.php?f=6&t=28676

Merudo:
I think the problem is that the `{a=b}` type of campaign properties are confused for a json object `{a:b}`.

Merudo:
PR #1871 fixes this.

The order of operation was reversed - the property should be parser as code and then potentially parsed as a json - doing it in the opposite order leads to `{a=b}` being parsed as a json.

Azhrei:
Is this the only place where code can be enclosed in braces?  Everywhere else requires the use of `[r: ...]` to achieve the same effect, right?

If so, why even allow the braces here?

Merudo:
> Is this the only place where code can be enclosed in braces? Everywhere else requires the use of `[r: ...]` to achieve the same effect, right?
> 
> If so, why even allow the braces here?

Nope, everywhere else `[r: ]` and `{}` give identical results.

Phergus:
[...]

PR #2483, opened 2021-03-20
Improve performance of token property evaluation

selquest:
- Improves performance evaluating JSON object properties by detecting properly formed JSON before attempting to parse as MTScript.  Using strict evaluation for this initial check maintains the improvement of #1560 without incurring the performance cost of trying to parse JSON objects as MTScript first.
- Collect performance data while preparing the statSheet popup.
- Removes unnecessary double evaluation of properties on the statSheet popup (pre-dated `getEvaluatedProperty()`)
- Add unit test to confirm proper evaluation of Token properties, including the malformed JSON that Gson shouldn't really be supporting but is, and that we'll continue to support to maintain backwards-compatibility.

~~Planning to add another change, converting default value lookup to use a Map instead of iterative comparisons across an entire list of definitions.~~  This planned follow-on has been moved to its own issue, so this PR is now ready.

RPTools/maptool#2396

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2483)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/MapToolVariableResolver.java
+++ b/src/main/java/net/rptools/maptool/client/MapToolVariableResolver.java
@@ -334,8 +334,12 @@ public class MapToolVariableResolver implements VariableResolver {
   }
 
   protected void updateTokenProperty(Token token, String varname, String value) {
-    MapTool.serverCommand()
-        .updateTokenProperty(tokenInContext, Token.Update.setProperty, varname, value);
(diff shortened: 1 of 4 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2483#-:-MWQ3mYRB0C_m-asrB60:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @selquest)

PR #2503, opened 2021-03-29
onChangeToken version 2

TheWoopWooper:
Provides additional information to a macro that is triggered using the onChangeToken event, such as update source, and any parameters that are easily representable in string form.

For example: "tokenId=5CEA82FC008847AEAEB3265F0C6B7125;update=setProperty;parameters=Speed,4;"

A user that wishes to use version 2 may include _'version="2"'_ in their onChangeToken link tag, otherwise, defaults to the original implementation.

---

In order to make sure existing macros that relied on this don't fail, I decided to supply the new implementation only if the macro writer has added version=2 to their link tag. I hope this solution is ok.

If it's possible to get this merged in some form it would be much appreciated! This feature would really help me with optimising my framework. It auto-updates some html forms with new data and since the current version of MT provides no way to know what data changed, has to process every display property which takes upwards of 800ms to complete. With this change the update can run at around 25ms.

I've also removed the requirement that the frame be visible, users would switch to another tab and come back and find the data was out of date. If the macro writer wishes not to update non-visible frames they can simply inspect the frame properties. I don't imagine anyone uses this with the expectation that frames in other tabs would not be counted.

P.S. Do I need to make a ticket for this too?

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2503)


Phergus:
> P.S. Do I need to make a ticket for this too?

Yes.  A feature request (or bug report) ticket must be opened before a PR will be accepted.

TheWoopWooper:
Sorry, wasn't sure on the protocol, here you go: https://github.com/RPTools/maptool/issues/2505

cwisniew:
Questions about this PR
1. Why do we need to change the name of the call back function if the link also has to specify the attribute version=2. The attribute should be enough to make sure the callback is called with correct parameters. 
2. I think this changed properties are better passed as a json array as to many potential conflicts with characters in property names otherwise

TheWoopWooper:
New version up, rebased up to date. Uses json now as requested.

cwisniew (approved):
> Reviewed 12 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2503)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @TheWoopWooper)

cwisniew (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2503)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @TheWoopWooper)

cwisniew (approved):
> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2503)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @TheWoopWooper)

Phergus:
No action by OP.

PR #2507, opened 2021-03-30
Issue #2506

### LINKED ISSUE #2506: An asset with an empty ID ("") will be stuck as transferring
**Describe the bug**
If the `src` of an `img` points to an asset URL that is empty (ie a string of just `asset://`) the asset will get stuck as transferring and never update. This will also make `ImageManager.getImageAndWait()` hang.

**To Reproduce**
1. Enter `` into the chat window.
2. Observe that it displays `TRANSFERING_IMAGE` (the question mark), indicating that it is stuck transferring.

**Expected behavior**
An asset with a empty ID should be marked as broken, like any other invalid ID.

**MapTool Info**
- Version: Current Develop
- Install: Any

 - OS: Windows
 - Version 10

**Additional context**
The issue is ultimately because of this:
https://github.com/RPTools/maptool/blob/98cdf98188d2eb6de7724199af7af2c53b9a9d9e/src/main/java/net/rptools/maptool/server/ServerMethodHandler.java#L440-L443

Doing an early exit if the id is empty causes the asset to never update to broken. The simplest fix would be to just remove that part of the condition, and as far as I can tell does not introduce any issues (trying to fetch something with an ID of `""` will resolve the same way any other invalid ID does).

Irarara:
Fixes #2506

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2507)

### DIFF
--- a/src/main/java/net/rptools/maptool/server/ServerMethodHandler.java
+++ b/src/main/java/net/rptools/maptool/server/ServerMethodHandler.java
@@ -438,7 +438,7 @@ public class ServerMethodHandler extends AbstractMethodHandler implements Server
   }
 
   public void getAsset(MD5Key assetID) {
-    if (assetID == null || assetID.toString().length() == 0) {
+    if (assetID == null) {
       return;
     }
     try {


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2507#-:-MXCImUG8IgfQbpVMf_A:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Irarara)

Issue #2441, opened 2021-03-07
return(0) renders an error trace


wolph42:
**Describe the bug**
[h:return(0)] in the code renders an error trace

**To Reproduce**
Create new macro on lib:test
put it: 
[h:return(0)]
run it. 

**Resulting behavior**
Error trace : (new)@Lib:test

**Expected behavior**
nothing (as in no output)

**MapTool Info**
- Version: 1.8.3
- Install: New

**Desktop (please complete the following information):**
 - OS: win10

MeloDub:
Working on this.

Phergus:
Tested.  Calling a macro that has [h: return(0)] no longer produces an error.  Return is now `null`.

selquest:
I'm not sure these changes are working as intended, and I'm a little confused by some of the choices.  This appears to be overloading the existing `macro.return` special variable (normally used to convey a result back to a caller) as a flag to control whether the exception produced in the `return()` function should be ignored.  Reuse aside, do we know for certain that we never want `macro.return` to contain other information while we're making this decision?  Also, we're now creating an empty Result in the new catch block in `MapToolLineParser` - does this still allow the rest of `return()` functionality to operate?

I haven't tested this thoroughly myself yet, but @wolph42  has demonstrated on Discord some confusing behavior of `return()`, and this jumped out at me as the only mention of `return()` on the 1.8.4 Change Log.

@Phergus does reviewing an already closed-PR make sense?  Do we have a re-open workflow if it turns out these changes do need to be reverted?  Should further investigation be a new issue (they may be anyway - not sure Wolph's issue is actually directly related)?  I don't know what our SOP is here.

JonathanGabor:
#2513 appears to be caused by this.

Phergus:
@selquest we don't really have an SOP for this but generally have just opened a new bug issue referencing the previous issue/pr that way anyone looking at the old ones will see the new mentions at the bottom of the discussion.

PR #2515, opened 2021-04-08
Added macro.catchReturn

### LINKED ISSUE #2513: "return(0,0)" does not return if "macro.return=1" has been set by a previous statement or macro/udf call.
**Bug Description**
The following code snippet aborts with "must not happen" but should never reach the assert:

```
[h: macro.return = 1]
[h: return(0, 0)]
[h: assert(0, "must not happen")]

```
This is not real code, but a boiled down sample of minimum lines to show the error.

**MapTool Info**
Version 1.8.4 (new install) shows this bug, 1.8.3 works as expected.

**Desktop**
 - Linux [Gentoo]

**Additional Info**
Replacing the first line with  `[h: macro.return = 0]` or `[h: macro.return = 2]` or `[h: macro.return = "anything not == 1"]` or replacing it with a call to another macro/UDF (wich does not set macro.return=1) works as I expect, i.e not failing with "must not happen".

**For the curious**
The real code which runs into this error is:

```
[h: '<!-- ==== void switchToCurrent() ==== -->']
[h, if(she_isDebug()): she_debug(getMacroName())]

[h: currentKey = "activeName_" + getPlayerName()]

[h: "<!-- check Library -->"]
[h: selected = getLibProperty(currentKey, "Lib:sheSettings")]
[h, if(she_hasTokenRights(selected)), code: {
    [switchToken(selected)]
    [return(0, 0)]
};{}]

[h: "<!-- check Impersonated -->"]
[h: selected = getImpersonated()]
[h, if(she_hasTokenRights(selected)), code: {
    [switchToken(selected)]
    [setLibProperty(currentKey, selected, "Lib:sheSettings")]
    [return(0, 0)]
};{}]

[h: "<!-- check Selected -->"]
[h: selected = getSelected()]
[h, if(listCount(selected) == 1 && she_hasTokenRights(selected)), code: {
    [switchToken(selected)]
    [setLibProperty(currentKey, selected, "Lib:sheSettings")]
    [return(0, 0)]
};{}]

[h: assert(0, "Kein Token verk&ouml;rpert oder ausgew&auml;hlt!")]
```

The UDF `she_hasTokenRights(selected)` returns true (by setting its macro.return=1), the "true" branch of the if gets executed, but `return(0,0)` did not return. In fact all "true" branches get executed, all `return(0,0)` get executed but did nothing and finally the assert throws its error. This was quite tricky to pin down ;-), because if any other macro/UDF (e.g. `she_debug("something")`) which does not return "1" is called before `return(0,0)` everything works. I assume that's because the UDF "she_debug" does not set "macro.return" at all and thus apparently clearing it.

JonathanGabor:
Fixes #2513
See also recent discussion on #2441

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2515)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
+++ b/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
@@ -1050,7 +1050,7 @@ public class MapToolLineParser {
       return result;
     } catch (ReturnFunctionException e) {
       log.debug(e);
-      boolean catchReturn = BigDecimal.ONE.equals(resolver.getVariable("macro.return"));
+      boolean catchReturn = BigDecimal.ONE.equals(resolver.getVariable("macro.catchReturn"));
       if (!catchReturn) throw e;
 
       // return an empty result to not collide with tooltips
--- a/src/main/java/net/rptools/maptool/client/MapToolVariableResolver.java
+++ b/src/main/java/net/rptools/maptool/client/MapToolVariableResolver.java
@@ -98,6 +98,7 @@ public class MapToolVariableResolver implements VariableResolver {
       this.setVariable("macro.args", "");
       this.setVariable("macro.catchAbort", BigDecimal.ZERO);
       this.setVariable("macro.return", BigDecimal.ZERO);
+      this.setVariable("macro.catchReturn", BigDecimal.ZERO);
       this.setVariable("macro.catchAssert", BigDecimal.ZERO);
       this.setVariable("macro.args.num", BigDecimal.ZERO);
       this.setVariable(TokenMoveFunctions.ON_TOKEN_MOVE_DENY_VARIABLE, 0);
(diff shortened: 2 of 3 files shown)


selquest:
I still haven't had a chance to look into this more, so I may just be out of the loop - `macro.catchReturn` makes a lot more sense to me than the initial overloading of `macro.return`, but I'm still confused about the overall plan here.

As I understand it, our desired behaviors from `return()`:
- `return(0)` - end the current macro, setting `macro.return` to `null` for the calling context.
- `return(0, "someValue")` - end the current macro, setting `macro.return` to `"someValue"` for the calling context.
- `return(1)` - effectively a no-op, current macro continues executing as though the return statement was not present

Based on the wiki, I wouldn't expect an error trace to be shown in any of these situations:
> If you prefer to display an error message when exiting the macro see the assert() function or if you want to abort the flow of overall macro execution (e.g. in case of an error) see the abort() function.

So why do we need `macro.catchReturn`?  And if we **do** need it, are we sure that creating a new empty Result object is the correct behavior in the MapToolLineParser's catch block?  At the very least, shouldn't we be returning the `returnValue` parameter (was that attached to the ReturnException?)?

Some execution patterns to be sure to test off the top of my head:
- using `return(0)` in chat
- using `return(0)` in macro called from chat
- using `return(0)` in UDF called from chat
- repeat above including a `returnValue` param
- repeat above called from macro button
- repeat above called from UDF

PR #2516, opened 2021-04-08
Removed macro.return override

### LINKED ISSUE #2513: "return(0,0)" does not return if "macro.return=1" has been set by a previous statement or macro/udf call.

MeloDub:
Fixes for return() function. 
Closes #2513

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2516)


selquest:
So what's the state of #2441 after this change?  No longer overriding `macro.return` is good, but have we achieved the overall desired `return()` behavior - basically behaving as it did in 1.7, but eliminating the error call stack introduced in 1.8.3?

Copying comment from the (now closed) PR with other discussion: #2515:

As I understand it, our desired behaviors from `return()`:
- `return(0)` - end the current macro, setting `macro.return` to `null` for the calling context.
- `return(0, "someValue")` - end the current macro, setting `macro.return` to `"someValue"` for the calling context.
- `return(1)` - effectively a no-op, current macro continues executing as though the return statement was not present

Based on the wiki, I wouldn't expect an error trace to be shown in any of these situations:
> If you prefer to display an error message when exiting the macro see the assert() function or if you want to abort the flow of overall macro execution (e.g. in case of an error) see the abort() function.

Some execution patterns to be sure to test off the top of my head:
- using `return(0)` in chat
- using `return(0)` in macro called from chat
- using `return(0)` in UDF called from chat
- repeat above including a `returnValue` param
- repeat above called from macro button
- repeat above called from UDF

Phergus:
@MeloDub 
> One question: what is UDF?

User-Defined Function - see [defineFunction()](https://wiki.rptools.info/index.php/defineFunction)

MeloDub:
I ran some other test, it seems that an exception is thrown when calling `return(0)` from an UDF (chat or macro button):
`java.lang.NullPointerException: Cannot invoke "String.replaceAll(String, String)" because "output" is null`

This exception is not thrown if I put a return value in the macro (i.e. `return(0, 0)` or `return(0, "test")`). 
Was this error already present before 1.8? I can confirm that this happens without my fixes too.

Phergus:
It's a different error in 1.7.0 but, yes, using `return(0)` in MT 1.7.0 throws an exception when used in a UDF:
`java.lang.NullPointerException error executing expression afunc().`

Phergus:
@MeloDub you need to comment on this issue so that it can be assigned to you.

Phergus (approved):
:lgtm:

> Reviewed 2 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2516#-:-MYHMZjl6_jGTp7JW84k:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @MeloDub)

Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2516#-:-MYHMaSG0FJZE2wRFqES:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @MeloDub)

PR #2520, opened 2021-04-10
Switch to Graalvm js and update JSAPI - closes #2519

### LINKED ISSUE #2519: Re-Add Missing Classes for JS-API and migrate to GraalVM-JS
Commit 73fa8c89443cc58992d702d0b51f6608eea8ebba removed several unused classes, but caught `net.rptools.maptool.client.script.javascript.api.*` in the crossfire.  This broke the Nashorn-based javascript macro system.  

Additionally, Nashorn itself is going away (removed in OpenJDK JDK 15), so simply re-adding the removed classes would be a temporary solution at best.  There is an alternative Javascript implementation, available from OpenJDK 11 on, called GraalVM-JS, which is generally superior to Nashorn, and can easily be used in Maptool in its place.

Switch from Nashorn to GraalVM-JS, and re-add the missing interface/wrapper classes, updated for use with the new JS engine.

Just re-add the old classes.  They would work without modification, at least until MT switches to JDK 15 (which will likely be soon due to unrelated problems).
Just switch to GraalVM-JS, and remove the built-in reference to `MTScript` which causes `js.eval` to immediately fail.  This would be functional, but essentially leave the javascript only able to do basic math/string manipulation.  Without the API wrapper classes, the map, tokens, drawings, and even client information is inaccessible to JS, and the JS side cannot post messages or roll dice.

perkinslr:
As explained in #2519, this switches to GraalVM-JS for the javascript implementation, and re-adds the broken/missing JSAPI classes.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2520)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -141,7 +141,7 @@ runtime {
                'jdk.dynalink',
                'jdk.jfr',
                'jdk.jsobject',
-               'jdk.scripting.nashorn',
+               'org.graalvm.js',
                'jdk.unsupported',
                'jdk.unsupported.desktop',
                'jdk.xml.dom',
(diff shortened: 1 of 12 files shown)


cwisniew (approved):
> Reviewed 12 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2520)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @perkinslr)

PR #2532, opened 2021-04-13
Unify code for fetching images for asset:// URLs

### LINKED ISSUE #2334: Trying to display an image in HTML with a bad size parameter causes client to hang
**Describe the bug**
The way the size parameter/suffix (the `-##` value attached asset IDs used to scale them when displaying them in HTML) is processed results in the client hanging if the value cannot be parsed as an integer. Additionally, if the parameter is very large the client will also throw an exception and hang (there's no positive constraints on the value).

**To Reproduce**
1. Open a new campaign
2. Drop the Dragon token on the map
3. Enter `` into the chat. The client will hang on an exception.
4. Repeat the above, but enter `` instead. The client will again hang on an exception.

**Expected behavior**
The client handles the bad input without hanging.

**MapTool Info**
- Version: Development branch (1.8)

 - OS: Windows
 - Version 10

A thought, it might be easier to simply only allow images to be scaled down this way, rather than having to set an arbitrary constraint (the client already seems to handle dropping images that are excessively large). Is scaling images up something people regularly end up doing? (I didn't even know images _could_ be scaled up that way until mucking around.)

### LINKED ISSUE #2493: Images with size parameter don't scale correctly in HTML5 panels
**Describe the bug**
An asset fetched with the `-#` suffix normally scales the largest side of an image to the given value and maintains its aspect ratio. In HTML5/JFX panels (Frame5/Dialog5/Overlay), it scales both dimensions to that size instead.

**To Reproduce**
1. Drop a token (for the following macro, I used the Elf Token in the resource library)
2. Run the following in the chat:
```
[frame("Frame"): { <img src="[r:getTokenImage(100,"Elf")]"> }]
[frame5("Frame5"): { <img src="[r:getTokenImage(100,"Elf")]"> }]
```
3. Observe how in the frame5 the image is stretched.

**Expected behavior**
Images should display the same way across different types of panels.

**Screenshots**

**MapTool Info**
- Version: Current develop
- Install: Any

 - OS: Windows
 - Version 10

**Additional context**
The source of this issue is that in [`AssetURLStreamHandler.getInputStream()`](https://github.com/RPTools/maptool/blob/a3cbf627869c7926ff719e70829ec0abb1d2e3b8/src/main/java/net/rptools/maptool/client/AssetURLStreamHandler.java#L55) the size value is assigned to both `scaleW` and `scaleH` (which the image is scaled to later if either is > 0). 

I can work on this, but beyond simply fixing the bug, what I'd like to do is add a utility method to handle this stuff to ensure the behavior is consistent (something to just parse the asset URL, maybe scale, and return an image). If there's no objection to this, there's one thing I was wondering about. 

`AssetURLStreamHandler.getInputStream()` uniquely supports height and width query parameters. If support for these is kept, then other HTML panels would support that too (well, without an extra argument to ignore parameters). This seems undocumented and I couldn't find anything internal that takes advantage of this. Left to my own devices I'd opt to just keep the functionality just in case (though, it doesn't strike me as especially useful), but if anyone has any particular feelings about it, let me know.

Irarara:
This adds `ImageManager.getImageFromUrl()` and changes `AssetURLStreamHandler.getInputStream()`, `ImageLoaderCache.get()`, and `HTMLPanelImageCache.get()` to use it when fetching images from `asset://` URLs. 

This fixes #2493 and fixes #2334.

There are a few behavioral changes, though they mostly should not impact the way things are now outside of fixing the aforementioned issues:
* `Asset://` URLs now support `width` and `height` query parameters in all contexts (these only worked in HTML5 panels before)
  * If these parameters are present, valid or not, the size parameter (`-##`) is ignored.
* Images now only scale down (returning images unscaled if any parameter is greater than native size)
* Invalid parameter values (a value <= 0 or a non-integer value) return `BROKEN_IMAGE`

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2532)


Phergus (approved):
:lgtm:

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2532#-:-MYytWiNCzGcJWEMU1o3:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Irarara)

PR #2553, opened 2021-04-19
Allow direct execution of chat commands

### LINKED ISSUE #2550: Don't add internally sent commands to the chat's command history
Certain functionality within Maptool works by sending text commands to the chat, such as impersonating tokens or auto executing macros, and these commands get added to the chat command history the same as regular user input. This can make it  annoying when trying to get back to something you entered yourself, and it feels like the history should be reserved for text input submitted by a user.

Don't add internally sent commands to the chat's command history.

None.

### LINKED ISSUE #2551: Certain actions will clear the content of the chat's text area
**Describe the bug**
When performing certain actions (such as, impersonating and unimpersonating tokens, clicking an auto executing macros, using Speech on a token, etc.), any text currently entered into the chat's text area is cleared. This can easily result in something a user is in the middle of typing being lost and is rather unexpected behavior. 

**To Reproduce**
1. Place a token.
2. Type anything into the chat text area.
3. Impersonate the token (by the right click menu, impersonate window, etc.)
4. Observe that the text area is now empty

2-4 can be repeated with different actions instead of impersonating. Unimpersonating, running an auto executing macro, or using speech from a token's context menu all have the same behavior.

**Expected behavior**
The content of the text area should be preserved.

**MapTool Info**
- Version: Any
- Install: Any

 - OS: Windows
 - Version: 10

**Additional context**
This issue is because certain things work by submitting commands to the chat. In most cases, this means setting the text of the chat's text area to the command and then calling `commitCommand()`, overwriting whatever was in the text area. There is already `quickCommit()` that has the option to preserve the current input, but it's not widely used and when used the text is sometimes explicitly not preserved. Genuinely curious if there's a reason to want to clear the user's input.

The simplest way to address this would be simply to use `quickCommit()` when submitting commands. This works fine. An alternative however would be just skipping the middleman and allowing commands to be passed directly to `commitCommand()`, with the parameterless `commitCommand()` just passing in the content of the text area. This would also easily allow for different behavior when submitting from the chat vs something internal (useful for, say, #2550).

Irarara:
This changes `commitCommand()` so that commands can be passed into it directly, avoiding manipulating the chat's text area. This makes it so when executing commands internally the current content of the text area will not be overwritten and the command will not be added to command history. This also makes `quickCommit()` redundant, so I removed it.

This should close #2550 and close #2551.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2553)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/macro/impl/RunTokenSpeechMacro.java
+++ b/src/main/java/net/rptools/maptool/client/macro/impl/RunTokenSpeechMacro.java
@@ -48,11 +48,7 @@ public class RunTokenSpeechMacro implements Macro {
       if (tmacro == null) {
         continue;
       }
-      MapTool.getFrame()
-          .getCommandPanel()
-          .getCommandTextArea()
-          .setText("/im " + token.getId() + ": " + tmacro);
-      MapTool.getFrame().getCommandPanel().commitCommand();
+      MapTool.getFrame().getCommandPanel().commitCommand("/im " + token.getId() + ": " + tmacro);
     }
   }
 }
--- a/src/main/java/net/rptools/maptool/client/ui/AbstractTokenPopupMenu.java
+++ b/src/main/java/net/rptools/maptool/client/ui/AbstractTokenPopupMenu.java
@@ -43,7 +43,6 @@ import javax.swing.KeyStroke;
 import javax.swing.SwingConstants;
 import javax.swing.filechooser.FileFilter;
 import javax.swing.filechooser.FileNameExtensionFilter;
-import javax.swing.text.JTextComponent;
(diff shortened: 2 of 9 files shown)


cwisniew (approved):
> Reviewed 9 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2553)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Irarara)

Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2553#-:-MYlzuT_3p0JUUlCvMSW:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Irarara)

Issue #2583, opened 2021-04-26
Some items are not internationalized


fw:
**Is your feature request related to a problem? Please describe.**
Some texts are not internationalized. Many of them are dropdown list components, such as Token Size selection.

I couldn't determined category for this issue. Possibly, this may be fitted to place in bugs.

**Describe the solution you'd like**
To add missing i18n properties.

**Additional context**
Here are items that is not internationalized I have noticed.

#### Token Edit dialog
* Each items within Method selection at VBL Optimization in VBL Tab
* Each items within Shape, Size and Terrain Modifier selection and Ignored Terrain combo box in Config tab

#### Campaign Properties dialog
* Each items within Type and Corner selection in State tab
* Each items within Type and Side selection in Bar tab

#### Preferences
* Each items within New Map Grid Type and Light, Movement metric, Duplicate Token Numbering, Show Numbering on and New Token Naming in  the Interactions tab

#### Map Properties dialog
* Each items within AI Cell Rounding and Light

#### Token Context Menu
* Size categories (Medium, Small, and so on) within Size submenu on Token and Map Explorer

#### Draw Explorer
* All items in context menu of drawings

#### Drawing Tools Palette
* All tool-tips (such as Snap to grid, Line Type,  Eraser, Pen Width, Opacity)

#### Status Bar
* Tool-tips for Zoom Level and Used Memory

#### Map
* Following text labels: "Map not visible to players" and "Player View"

Phergus:
Could be either but Feature is good enough.

Irarara:
This is great. I've been keeping a list of things I've noticed as well and been wanting to tackle some of this stuff.

I suppose that makes this issue as good a place as any to ask a few questions I've had related to adding/editing i18n strings:
1. Is there a naming scheme that should be followed for keys or otherwise any preference about how things should be organized? I get the general gist of it but it's not always consistent. Specific example, the entire context menu for drawings needs to be added, but context menu items seem to be all over (token context menu items are`token.popup.menu.*`, init panel context menu items are `initPanel.*`, etc.) I can always just pick something to go with, but I just want to make sure I'm not going to make things disorganized.
    1. If you have something like say, a dialog window that's opened with a menu option, is it preferred to associate those things together (like, `foo.bar.dialog.*`) or organize them separately (both seem to be done?).
2. If I understand right, when the source strings are changed the strings for other languages are updated to reflect them. Is there a way to get around that if you're trying to change something that shouldn't affect the translations?
[...]

PR #2618, opened 2021-05-03
Localize Campaign Properties dropdowns

Irarara:
Part of #2583.

This adds strings for the state type, state corner, bar type, and bar side dropdowns to `i18n.properties`.

This feels slightly hacky, or at least, it feels like it could be done in a nicer way, but I'm under the impression a lot of these dialogs are going to be reworked at some point anyway so it didn't seem too worthwhile to mess around with their behavior too much.

With this change,  `JComboBox` items in form XML will be localized by `FormPanelI18N`. With state type, I stuck the original English name in the `name` value, and for state corner I stuck the enum name in the `name` value, because those strings are used in `TokenStatesController` (which previously relied on checking the label instead). Previously, the names were all `listitem`. This wasn't necessary for bar type or bar side, because `TokenBarController` references the selection by index.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2618)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2618#-:-MZyL6TlDZomEOd645QP:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Irarara)

PR #2624, opened 2021-05-06
Localize Preferences dialog drop downs

Irarara:
Part of #2583.

This adds localization support for the remaining drop downs in the preferences dialog (new map grid type, movement metric, duplicate token numbering, show numbering on, and new token naming). For the combos where the items were just strings, I added a little class (`LocalizedComboItem`) that stores the localized name and the the preference option values (ie, the old labels).

 I also added localization support for the New Token Naming "Use "Creature"" option, since it made sense to do that at the same time.

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppPreferences.java
+++ b/src/main/java/net/rptools/maptool/client/AppPreferences.java
@@ -879,7 +879,7 @@ public class AppPreferences {
   }
 
   public static void setMovementMetric(WalkerMetric metric) {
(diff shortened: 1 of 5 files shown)


Phergus (approved):
:lgtm:

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2624#-:-M_5e_T4Dn9fWcxOdYSL:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Irarara)

PR #2662, opened 2021-05-19
Flush DrawableRenderers on DRAWABLE_ADDED and DRAWABLE_REMOVED events

### LINKED ISSUE #2661: Changes made to drawings are not immediately reflected on map
**Describe the bug**
When manipulating drawings through the context menu in the Draw Explorer panel, such as changes to their order (Bring to Front/Send to Back) or pen settings (through Set Properties) the changes are not reflected until one of the following things happens:
1. The zoom level is changed
2. A drawing on the same layer is added
3. A drawing on the same layer is deleted
4. A drawing is moved to or from the same layer (this is really just 2 and 3)

This goes for the client the action is originating from and any connected clients.

**To Reproduce**
1. Add a drawing
2. Set a different fill color
3. In the Draw Explorer right click on the drawing and select "Set Property"
4. Observe nothing happens
5. Zoom in or out (or do any of the other things mentioned above) and observe that the drawing updates to the new fill color

For order, just do the above but layer two drawings and use Bring to Front and Send to Back from the context menu. There will be no update until one of the things mentioned is done.

**Expected behavior**
Changes to drawings should be reflected on all clients without any extra user input.

**MapTool Info**
- Version: Current develop
- Install: Any

 - OS: Windows
 - Version 10

Irarara:
Fixes #2661. This makes it so `DRAWABLE_ADDED` and `DRAWABLE_REMOVED` events guarantee that the `DrawableRenderer` will be flushed. Previously, the renderers only flushed if the number of drawings on a layer changed, if the zoom level changed, or if `flush()` was manually called elsewhere.

There's a few ways the issue could have been addressed, but I noticed the `DRAWABLE_ADDED` and `DRAWABLE_REMOVED` events already existed* and using them seemed like a natural way of going about this. Rather than flushing every time the events are fired this sets a flag to indicate that the renderer should be flushed the next time `renderDrawables()` is called (to avoid a bunch of redundant flush calls with batch deletions, etc.).

`DiskBasedPartitionedDrawableRenderer` seems to be unused, but since I modified the underlying interface I just made it work the same just in case.

*Note: updating a drawing fires `DRAWABLE_ADDED`. Should there be a `DRAWABLE_UPDATED` event? It didn't seem like there was much point making any changes there right now since they weren't even being put to use, but, just a thought.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2662)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/DiskBasedPartitionedDrawableRenderer.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/DiskBasedPartitionedDrawableRenderer.java
@@ -47,12 +47,13 @@ public class DiskBasedPartitionedDrawableRenderer implements DrawableRenderer {
 
   private final Map<String, BufferedImage> chunkMap = new HashMap<String, BufferedImage>();
 
-  private double lastDrawableCount;
   private double lastScale;
   private Rectangle lastViewport;
 
   private int horizontalChunkCount;
   private int verticalChunkCount;
+
+  private boolean dirty = false;
 
   static {
     try {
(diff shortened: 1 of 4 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2662#-:-MaEnLOa5k5HftxgYa6G:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Irarara)

PR #2682, opened 2021-05-27
Add workaround for layout persistance issues with macro frames

### LINKED ISSUE #2485: Macro-generated frames do not remember their previous position
**Describe the bug**
Before 1.8.3, the frame created by the frame5() roll option would remember where it was last docked. On closing and reopening MapTool, creating the same frame (which I believe was determined by the frame title, unsure) would then create the frame where it was last placed. Since 1.8.3, MapTool no longer remembers the frame's last position and creates it floating, as if it had been created for the first time. Note that this only occurs with restarting MapTool, closing the frame itself and then recreating it without closing MapTool correctly places the frame.

**To Reproduce**
1. Create a frame from a macro using the frame5() roll option
2. Dock it in a position you like
3. Save the campaign and close MapTool
4. Reopen MapTool and rerun the macro
5. The frame will not be in the same place as it was docked when the campaign was saved.

**Expected behavior**
The frame should be created in the same place as it was docked when the campaign was saved.

**MapTool Info**
- Version: 1.8.4
- Install: New

 - OS: Windows 10

Irarara:
Fixes #2485. As detailed in the discussion around the issue, the layout information for frames that are not added to the `DockingManager` before layout is loaded is not retained. This works around that by saving the names of macro frames to a file (frames.dat, just a `NULL` delimited list of the names) when Maptool exits and then, when launching, using those names to generate placeholder frames before the layout is loaded so that the layout information is retained. Really, I'd consider this more of a "temporary" workaround to be removed if the issue is ever actually addressed within the library.

In my testing, it seems to do the trick, layout seems to restore the same as you'd expect pre-1.8, and I also haven't observed any issue with the encoding of the frame names. I did once observe frames opening incorrectly (they opened docked somewhere I know I hadn't put them, and when opening them in 1.7 they were in the right spot), but I haven't been able to reproduce this so it may have just been due to something I was doing at the time messing with things. Also of note, if the frame is floating its position is not restored, _but_ this is also the case for macro frames in 1.7.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2682)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/MapToolFrame.java
+++ b/src/main/java/net/rptools/maptool/client/ui/MapToolFrame.java
@@ -22,16 +22,16 @@ import java.awt.image.BufferedImage;
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
+import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.*;
 import java.util.List;
-import java.util.Map;
-import java.util.Observable;
-import java.util.Observer;
-import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.stream.Stream;
 import javax.swing.*;
+import javax.swing.Timer;
 import javax.swing.border.BevelBorder;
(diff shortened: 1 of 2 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2682#-:-Maj9D_GD8KPXbujCUtc:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Irarara)

PR #2685, opened 2021-05-29
Make the external IP address determination be parallel.

### LINKED ISSUE #2684: MapTool determines the external address by hitting every configured site.
**Describe the bug**
A MapTool client can take a very long time to connect to a server, if any service used to determine the external IP address of the server is slow to respond. This can take so long that MapTool appears to hang. A similar issue exists with the Connection Details dialog, though that can be closed without impacting other MapTool functionality. 

**To Reproduce**
Since this is a matter of handling unreliable external resources, reproducing the issue relies on one of the configured sites to be having issues. In that case, the issue can be seen by opening the Connection Details dialog and observing how long it takes to determine the external address. Alternatively, a server can be started and a client then connects to it, and the client will appear to hang.

**Expected behavior**
MapTool should use the result of the fastest service and should ignore other responses.

**Screenshots**
This is what a user sees when trying to find their connection details:
image
A thing is true for the Connect to Server dialog, but in that case the IP address discovery blocks the UI completely.

**MapTool Info**
- Version: 1.8.3, also reproduced on `develop`.
- Install: New

 - OS: Linux Mint
 - Version 20

**Additional context**
In my particular case, I was having trouble with only one service: `http://icanhazip.com/`. Incidentally, this is the last service on the list (found in `ip-check.txt`). So MapTool went through four good services, got the same IP address four times, but still contacted this service as well and waited for it to respond.

kwvanderlinde:
This change introduces `getAddressAsync()` as an asynchronous alternative to `getAddress()`. `getAddressAsync()` will attempt to read an IP address from all URLs concurrently, and will only resolve once one of them returns successfully. The original `getAddress()` is now a wrapper around `getAddressAsync()` that waits for the IP address to be returned, and times out after a while otherwise.

Fixes #2684.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2685)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/MapToolRegistry.java
+++ b/src/main/java/net/rptools/maptool/client/MapToolRegistry.java
@@ -28,6 +28,11 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.TimeZone;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import net.rptools.maptool.server.MapToolServer;
 import okhttp3.Call;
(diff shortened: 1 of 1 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2685#-:-Matu3Ro13Z_EVgVXrae:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

Issue #2379, opened 2021-01-23
AI Pathfinding works incorrectly in closed area if it is most top-most MBL area on map


melek:
**Describe the bug**
With AI Pathfinding enabled, the top-most (lowest 'Y' position value) enclosed area of MBL (or VBL if VBL Pathfinding is enabled) will behave as a thicker boundary than it should all around the interior of the shape, effectively 'shrinking' the pathfinding area.

* This issue only affects closed shapes with boundaries at or slightly within the closed shape's grid intended cells.
* Any line/non-closed shape higher on the map (lower 'Y' position value) will correct the error for other shapes on the map. 
* Having MBL (or VBL boundaries if VBL Pathfinding is enabled) boundaries that extend slightly into surrounding cells also corrects the issue.
* If multiple shapes share a 'Topmost'/lowest Y value, the left-most (lowest X value) will be the broken shape.

**To Reproduce**
1. Open a new campaign. Make sure the map has a SQUARE grid.
2. Draw a grid-aligned (hold CTRL) rectangle with the Draw tool (this is just for visual aid), at least 4 cells by 4 cells in size.
3. Draw a grid aligned (hold CTRL) HOLLOW RECTANGLE of MBL perfectly around the drawn area.
4. Place any token in the center area of the shape and enable AI Pathfinding. 
5. Move the token. Note that you cannot path directly to the edges of the shape.
6. Repeat steps 2-5 with another shape ABOVE (lower Y position value) the current shape. 
7. Notice that the NEW shape blocks movement incorrectly near edges, but the original shape now works as it should.

- You can test the same with VBL and enabling VBL pathfinding.  Switching VBL pathfinding on and off may affect which closed shape is the 'topmost' and thus broken.
- If you add a non-closed shape MBL (or VBL if VBL pathfinding is enabled) above all other shapes, all AI pathfinding will work correctly.

**Expected behavior**
MBL/VBL pathfinding should not act incorrectly based on the position of the MBL/VBL, especially not relative to other MBL/VBL shapes.

**Screenshots**
![Screenshot of issue](https://user-images.githubusercontent.com/220149/105613510-d28ed700-5d77-11eb-811c-e56c4ff4d774.png)

**MapTool Info**
- Version: 1.7.0 and 1.8b2
- Install: New installs alongside many older versions.

**Desktop (please complete the following information):**
 - OS: Windows 10

**Additional context**
This came up in a game while I was DMing, so it affects players during a hosted game as well.

[...]

PR #2692, opened 2021-06-01
Fix AI pathfinding in the top-leftmost MBL box.

kwvanderlinde:
This is a workaround to address #2379.

Since `PathIterator` returns all holes prior to returning all shells, we reverse the order to work better with JST's assumption that holes follow shells. Although this still does not fully fulfill JST's assumptions, it is sufficient for JST to correctly identify all holes when combined with buffering.

This change relies on the order of `PathIterator`. Although the order of `PathIterator` appears to be consistent, I haven't found this documented anywhere. So this change isn't really a proper fix as much as it is a workaround for JST's behaviour as it stands with the current behaviour of `PathIterator`. A robust solution would require fixes in JST to reduce or correct assumptions about `PathIterator` behaviour.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2692)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/walker/astar/AbstractAStarWalker.java
+++ b/src/main/java/net/rptools/maptool/client/walker/astar/AbstractAStarWalker.java
@@ -162,8 +162,9 @@ public abstract class AbstractAStarWalker extends AbstractZoneWalker {
       try {
         vblGeometry =
             shapeReader
-                .read(vbl.getPathIterator(null))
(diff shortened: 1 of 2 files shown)


cwisniew (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2692)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

Issue #2696, opened 2021-06-03
AI pathfinding behaves suboptimally and inconsistently when using 1-2-1 movement with token terrain modfiers


kwvanderlinde:
**Describe the bug**
When using 1-2-1 movement, the pathfinding sometimes chooses to go through tokens with a terrain cost when going around is more optimal. In other cases, it avoids the terrain by multiple cells when not necessary. Moreover, the difference in behaviour can be quite drastic depending on the route taking to get to the problematic terrain.

**To Reproduce**
1. Ensure that "Movement metric" is set to "One-Two-One" in the preferences dialog
2. Create a new map with a square grid. I use 5.0 cell distance, but this shouldn't matter.
3. Create a line of 5 terrain tokens as shown in the attached screenshots. Set each token's Terrain Modifier to Add 5.0
4. Create one more token to use for pathfinding
5. Place the pathfinding token above and to the left of the terrain tokens, with one diagonal grid space between (shown in the screenshots). Then try each of these scenarios:
    1. Drag the token along the line of terrain tokens, to any position between the terrain tokens (or an odd number of cell past the last terrain token). Notice that the path goes straight through the entire line of tokens.
    2. Drag the token one cell down, set a waypoint (right click), then drag to the same destinations as the previous step. Notice the path avoids all terrain tokens and results in a cheaper (even optimal) route.
    3. Drag the token along the line of terrain tokens, to any position coinciding with a terrain token (or an even number of cells past the last one). Notice that the route dodges all the terrain tokens and gives an optimal route.
7. Now place the token immediately above and to the left of the terrain tokens. Then try each of these scenarios:
    1. Drag the token along the line of terrain tokens, to any position between the terrain tokens. Notice that the path goes straight through the entire line of tokens.
    2. Drag the token one cell down, set a waypoint (right click), then drag diagonally until right next to the destination, set another waypoint, then move one cell to the right to get to one of the same destination as the previous step. Notice this path is forced to avoid all terrain tokens and results in an optimal route.
    3. Drag the token along the line of terrain tokens, to any position coinciding with a terrain token. Notice that the route goes through all the terrain tokens.
    4. Drag the token one cell down, set a waypoint, then drag it onto a terrain token. Notice that this path avoid all but the last terrain token and gives and optimal route.
    5. Drag the token to the location immediately below and to the right of the last terrain token. Notice the path avoid all terrain tokens, and even veers further away for the final two terrain tokens.
[...]

PR #2713, opened 2021-06-07
Fix and tweak various aspects of the A* walker, especially relating to 1-2-1 movement on square grids.

kwvanderlinde:
There are two big changes for correctness:
1. Modified nodes are reinserted into the priority queue. Without this, we weren't getting the node with the lowest f-score, which causes some nodes to be closed prematurely. This is a general fix that applies to all metrics and all grids.
2. The g-score now agrees with the 1-2-1 metric, which is also displayed to the user. Previously the g-score was counted as 1-1.5-1, which caused A* to consider some sub-optimal paths as though they were optimal.

A few minor, technical changes are also in here:
1. The heuristic function (`hScore()`) is now admissible for square grids. It now *rewards* paths that are more in-line with the start-goal direction, rather than *punishing* out-of-line paths.
2. Path aesthetics have been improved in some common cases by making the heuristic function more accurate under 1-2-1 (it now accounts for the number of diagonals already taken in the path). This extra accuracy allows the tie breaker to have an effect in more cases.
3. Concurrency issues with the A* debug labels have been fixed by having them added and removed through the AWT EventQueue.

This fixes most of the errant behaviour in #2696.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2713)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/walker/astar/AStarCellPoint.java
+++ b/src/main/java/net/rptools/maptool/client/walker/astar/AStarCellPoint.java
@@ -56,18 +56,56 @@ public class AStarCellPoint extends CellPoint implements Comparable<AStarCellPoi
     isAStarCanceled = aStarCanceled;
   }
 
+  /**
+   * Create an A* node from a position.
+   *
+   * <p>The node will have no scores or data carried over from a cell point.
+   *
+   * @param x The x cell position of the node.
+   * @param y The y cell position of the node.
+   */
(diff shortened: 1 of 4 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2713#-:-Mbc5is6APeSnoD3jtnN:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #2718, opened 2021-06-09
Prepare JTS geometry prior to pathfinding

### LINKED ISSUE #2716: Pathfinding performance significantly degrades in the presence of complex MBL.
**Describe the bug**
When dragging a token through complex MBL on a moderately sized map grid, pathfinding is slow and can easily time out.

**To Reproduce**
1. Open the attached maze map, that has lots of MBL.
2. Drag the token from the maze entrance, straight up to the centre area. Gives a path length ~730.
3. Wait for some time. Depending on system specs and maybe the phase of the moon, it may complete after a few seconds or take long enough to time out.

**Expected behavior**
The pathfinding should be relatively quick.

**MapTool Info**
- Version: 1.8.3, `develop`
- Install: New

 - OS: Linux Mint
 - Version 20

**Additional context**
Pathfinding spends a significant amount of time checking whether movement is allowed from one cell to the next. While individually these checks are not very heavy, they are performed so many times that they are the bulk of the pathfinding cost when moving through complex MBL.

The way the MBL check is done is that a line segment is "drawn" between the centers of each cell, and then JTS is asked whether the line intersects with the MBL. Thankfully, repeated intersection against a given geometry is a case that is optimized by JTS. It just requires moving to `PreparedGeometry` instead of `Geometry`.

Maze map (rename to `labyrinth.rpmap`): [labyrinth.zip](https://github.com/RPTools/maptool/files/6619428/labyrinth.zip)

kwvanderlinde:
Since MBL doesn't change during any given pathfinding operation, prepared geometry is used to speed up the intersection
calculations done against it.

Fixes #2716.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2718)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/walker/astar/AbstractAStarWalker.java
+++ b/src/main/java/net/rptools/maptool/client/walker/astar/AbstractAStarWalker.java
@@ -41,9 +41,10 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.locationtech.jts.awt.ShapeReader;
 import org.locationtech.jts.geom.Coordinate;
-import org.locationtech.jts.geom.Geometry;
 import org.locationtech.jts.geom.GeometryFactory;
 import org.locationtech.jts.geom.LineString;
+import org.locationtech.jts.geom.prep.PreparedGeometry;
+import org.locationtech.jts.geom.prep.PreparedGeometryFactory;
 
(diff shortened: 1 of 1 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2718#-:-MbkYoM31jQkBgh2-maa:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #2721, opened 2021-06-09
Make path costs well-defined by distinguishing paths with even/odd diagonals.

kwvanderlinde:
Finishes #2696.

The core of this change is to augment A* nodes with a flag that indicates whether the node was arrived at via a path with an odd number of diagonals, specifically under 1-2-1 movement. This flag is also now part of the equality of the node. The effect is that each cell can have up to two nodes associated with it, one for an odd number of diagonals and one for an even number.

Most of the work in this change was in refactoring `AStarCellPoint`. This class served a few distinct purposes:
1. To be queued for use in A*, with path costs and estimates.
2. To associate terrain modifiers with cells.
3. To cache whether moves into a cell are allowed or are blocked by MBL/VBL.

(1) is the aspect we want to modify, while (2) and (3) are incidental uses of `AStarCellPoint` that rely on its equality being the same as for `CellPoint`. Since we seek to change the equality, it was necessary to break (2) and (3) into their own things. This is how things stand now:
1. Retained as `AStarCellPoint`, but no longer inheriting from `CellPoint` and only containing information needed for A* itself (parent node, cost estimates, ...). I would have like it to be renamed to `AStarNode`, but that resulted serialization issues, so for now the name is kept the same.
2. Now represented as a mapping from `CellPoint` to a new record `TerrainModifier`. This has a small perk of no longer needing to iterate over the entire set of terrain modifiers for each cell.
3. Now represented as a nested mapping, first from target cells, then from source cells, and maps to whether the move from the source to the target is valid. This is actually the same structure as before, just without needing to go through an `AStarCellPoint` anymore. Debug rendering was modified to use the new representation.

Now that each cell has two nodes, the debug labels have been updated to display in two columns per cell. The left cell is for when the path has an even number of diagonals, and the right cell for when the path has an odd number. Only one column will be used when not using 1-2-1 movement.

There is also a small bugfix here. Previously, a cell in the token's footprint would be closed if a move from that cell to *any* of the current node's neighbors would be blocked by VBL. This is despite the fact that other valid moves could be made from that cell, and that A* may still expect the cell to be open. This caused issues, for example, when running a large token along the inside top edge of a hollow rectangle of MBL. It did not really affect medium or smaller tokens since the only footprint cell is the current cell, which would end up closed anyways as part of A*. But now that two nodes are associated with each cell, this bug became a more obvious problem since both nodes for the *current cell* ended up closed even though better paths could still be found through the cell. This is now been fixed so that nodes are only closed by being processed by A*, or by determining that the cell is completely blocked thanks to terrain.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2721)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
@@ -83,7 +83,6 @@ import net.rptools.maptool.client.ui.token.AbstractTokenOverlay;
 import net.rptools.maptool.client.ui.token.BarTokenOverlay;
 import net.rptools.maptool.client.ui.token.NewTokenDialog;
 import net.rptools.maptool.client.walker.ZoneWalker;
-import net.rptools.maptool.client.walker.astar.AStarCellPoint;
 import net.rptools.maptool.language.I18N;
 import net.rptools.maptool.model.AbstractPoint;
 import net.rptools.maptool.model.Asset;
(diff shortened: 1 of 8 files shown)


Phergus (approved):
:lgtm:

> Reviewed 8 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2721#-:-MbmQo0DEnfN7cXMJosc:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

Phergus:
@kwvanderlinde   Oops.  Missed spotless.

kwvanderlinde:
Sorry, I thought I ran that after my final edit. Fixed now.

Phergus (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2721)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #2734, opened 2021-06-12
Reduce geometry calculations in A* implementation

### LINKED ISSUE #2729: MBL geometry is recalculated and checked too often when pathfinding, leading to an unfortunate user experience
**Describe the bug**
When a token is dragged with pathfinding enabled, every time the goal cell is changed the MBL/VBL AWT geometry is converted to JTS geometry. When there is lots of geometry, this can take a long time and leaves the user without any feedback that the drag is in progress. Also, when checking moves against MBL, the checks are cached but only when the token is near MBL, meaning expensive geometry checks are still done in the common case of moving through unobstructed cells. This leaves the user waiting longer than necessary for feedback about the path.

**To Reproduce**
1. Enable pathfinding
2. Import this map: [50x50 Cave 100px.dd2vtt.zip](https://github.com/RPTools/maptool/files/6640037/50x50.Cave.100px.dd2vtt.zip)
3. Place a token in the top-left of the map.
4. Try dragging the token to a neighboring cell.
    1. Without releasing the token, wait at that cell for a while (a second or more) and a path should be calculated.
    2. Without releasing the token, move a cell over and observe the same wait time.
5. Try dragging the token to a point across the map. Wait noticeably longer for a path to show up.

**Expected behavior**
When MBL/VBL does not change, the pathfinding algorithm should minimize the number of new geometry operations, since these are relatively expensive.

**Screenshots**
Dragging one cell takes about a second on my machine for any feedback to be presented to the user: 
large-vbl-processing-time-cropped
And then continuing the drag one cell over also costs a second:
large-vbl-processing-time2-cropped

A longer move like this takes about twice as long due to the great number of MBL intersection checks: large-vbl-long-move

**MapTool Info**
- Version: develop
- Install: New

 - OS: Linux Mint
 - Version 20

**Additional context**

kwvanderlinde:
Fixes #2729.

When the MBL doesn't change, we don't have do the heavy work of recomputing the corresponding JTS geometry. Additionally, all MBL-checked moves are now cached which decreases the number of geometry checks even further. We also don't do a bounds check to see whether cells contain any MBL, as that proved to be heavier than just letting JTS do the actual MBL check.

To avoid consistency issues, the move cache is now cleared whenever the MBL changes.

This change does not do anything to address the initial delay when dragging a token. This could have been solved by adding some static state to the walker, but that still wouldn't address the delay on the very first drag, or any delay after changing MBL. A proper solution would add some complexity, and may not even be considered worthwhile.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2734)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2734#-:-McFgaY_B874v7ZYVmuH:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #2738, opened 2021-06-15
Setting program name on macOS

mikehenry-se:
On macOS, the name shown in the "MapTool" menu of the menubar is the name of the default class "LaunchInstructions" used to start MapTool. This is less than ideal and this fix sets the name of the application to display "MapTool" rather than the default.

This fix has only been tested and verified on macOS but should work on other platforms as well since it uses an JVM -X argument, which are typically ignored unless supported by the particular platform JVM implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2738)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -122,7 +122,7 @@ run {
     // Add  -Djavax.net.debug=all to debug networking issues
 
     if (osdetector.os.is('osx')) {
-        applicationDefaultJvmArgs += [ "-Xdock:name=MapTool" ]
+        applicationDefaultJvmArgs += [ '-Xdock:name=' + project.name + developerRelease ]
     }
 
     if (System.getProperty("exec.args") != null) {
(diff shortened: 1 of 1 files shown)


Phergus:
I may have been a bit too subtle previously.  We need an Issue opened first that details the problem to be solved and then a PR can be opened to address that Issue.

Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2738#-:-McFgCi526cD7PtLhVJd:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mikehenry-se)

Phergus:
This option does not work on Windows and actually keeps the JVM from launching.

Phergus (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2738)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mikehenry-se)

Issue #2029, opened 2020-06-18
Remove undesirable files from wikiDocExtraction from soure control.


cwisniew:
**Describe the bug**
There are several "working" files in RPTools/maptool/build-resources/wikiDocExtraction that should never have been checked into source control in the first place and should be removed.

**Expected behavior**
These files would not have been included in the first place if procedures / requests not to / good practices had been followed.

These files should be removed and relevant entries added to git ignore.

**Additional context**
Files in question
processed.txt
wiki-dump.xml
wiki-has.txt
jamz-has.txt

Azhrei:
Actually, the whole directory shouldn’t be part of MapTool but should be a new repo.  You might as well go whole hog and move all of them.

cwisniew:
Done

PR #2752, opened 2021-06-18
Update macro editor help docs from wiki

Azhrei:
@melek has more things to edit on the wiki to add doc for macro events and special variables that are missing from this list.  But that requires new wiki templates, requires updating pages, requires updates to these scripts, and then all of that must be executed and the output snagged for use.

That means this PR catches us up to date, but there will be a few more to add for the next build.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2752)


cwisniew:
As discussed in #2029 the dump xml and intermediate working files shouldn't be part of source controlled files.

Azhrei:
Crud, I didn't catch that the wiki-dump was in there.  I'll add it to the `.gitignore`.

The working files, however, are very useful when the script(s) don't change but the wiki does.  It helps to know what the scripts produced last time when they need tweaks.  I run the scripts without making any changes, then look to see what they generated that's different.  Think of them like the gradle wrapper — generated files that are convenient from one execution to the next.

Also discussed in #2029 was moving the whole thing to a new repo.  Should we do that?  How about `wiki2properties`?

cwisniew:
Please also remove wiki-updates.adoc 
Either the updates should be made or an issue should be logged so it has visibility and can be worked on by anyone rather than burying a list of tasks to be done somewhere in a document in source control

PR #2784, opened 2021-06-23
Remove redundant PathIterator segments to avoid self-intersections

### LINKED ISSUE #2783: Conversion from AWT geometry to JTS geometry is not robust.
**Describe the bug**
An AWT `Area`'s `PathIterator` can yield near-zero length segments that are seen as self-intersections by JTS. This can cause surprising behaviour for nested polylines, as JTS cannot consistently identify the interior and exterior of a shape. It also makes VBL/MBL sensitive to changes in the JTS implementation.

**To Reproduce**
JTS 1.16.1 appears to be relatively lax about the sort of geometry we send it. I haven't been able to reproduce the issue unless on JTS 1.18.x.

1. Check out commit `c82be16e2f7d75471debb377fe55f15b6cb89e32`.
2. Run MapTool: `./gradlew run`
3. Import this DungeonDraft map: 
[50x50 Cave 100px.dd2vtt.zip](https://github.com/RPTools/maptool/files/6704800/50x50.Cave.100px.dd2vtt.zip)
4. Draw a hollow rectangle of MBL around the entire map.
5. Place a token inside the caves.
6. Try dragging the token. Notice that the path length remains at 0.

**Expected behavior**
It should be possible to use AI navigation with polylines under JTS 1.18.

**Screenshots**
By converting the JTS geometry back to AWT geometry, we can see that JTS considers the caves to be solid: broken-jts-geometry

We should expect that the caves are actually hollow like this: expected-jts-geometry

**MapTool Info**
- Version: develop
- Install: New

 - OS: Linux Mint
 - Version 20

**Additional context**
This issue was revealed after an attempt to upgrade to JTS 1.18.1 in #2589, and that was reverted shortly after in #2759. Fixing this will allow upgrading to future versions of JTS.

kwvanderlinde:
A weakness in AWT's `PathIterator` is that it sometimes yields miniscule segments that are essentially zero-length. When processed by JTS, these segments are interpreted as a self-intersection, which it tries to handle intelligently. Unfortunately, doing so is difficult especially when shapes are nested.

This changes removes these near-zero-length `SEG_LINETO` segments. Doing so also the possibility that a given ring no longer contains any `SEG_LINETO`. These degenerate rings are also removed.

In order to more easily discover broken geometry in the future. the `Geometry::buffer()` call is now done after the logging for invalid geometry. Since `Geometry::buffer()` always creates valid geometry, the logging was not being triggered even when tested with broken geometry.

Fixes #2783.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2784)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/walker/astar/AbstractAStarWalker.java
+++ b/src/main/java/net/rptools/maptool/client/walker/astar/AbstractAStarWalker.java
@@ -49,6 +49,7 @@ import org.locationtech.jts.geom.GeometryFactory;
 import org.locationtech.jts.geom.LineString;
 import org.locationtech.jts.geom.prep.PreparedGeometry;
 import org.locationtech.jts.geom.prep.PreparedGeometryFactory;
+import org.locationtech.jts.operation.valid.IsValidOp;
 
 public abstract class AbstractAStarWalker extends AbstractZoneWalker {
   private record TerrainModifier(Token.TerrainModifierOperation operation, double value) {}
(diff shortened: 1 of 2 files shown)


cwisniew (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2784)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

Issue #2785, opened 2021-06-24
Macros moved/copied between Macro Groups & Panels are always player-editable


melek:
**Describe the bug**
When moving a macro between a macro group, or when copying a macro from one panel to another, the 'Allow Players to Edit Macro' checkbox is always reset to 'checked'.

**To Reproduce**
1. Create a campaign macro and edit it so 'Allow Players to Edit Macro' is **unchecked**. Rename `Locked` or similar.
2. **Copy to Token Error**: Create a token.
3. Drag the `Locked` macro to the token's Selection panel. It will copy. 
4. Edit the copied `Locked` macro on the token.
5. See error: 'Allow Players to Edit Macro' is **checked**.
2. **Move to Group Error**: Create a second campaign macro in a new macro group from the first.
3. Drag the `Locked` campaign macro to the new macro group. It will move from one group to another (neat!).
4. Edit the moved `Locked` macro on the campaign panel.
5. See error: 'Allow Players to Edit Macro' is **checked**.

**Expected behavior**
All settings of the moved/copied macros are the same.

**MapTool Info**
- Version: 1.9.2
- Install: Upgrade 

**Desktop (please complete the following information):**
 - Windows 10
 
**Additional context**
- Duplicating a macro from the macro button context menu works correctly / does not do this.
- The current behavior is arguably reasonable for players connected to a server when dragging macros from the Campaign Panel to their own token panels. It is complicated here - players shouldn't be able to view the code of locked/trusted macros, but they also shouldn't be able to put a trusted macro on a token it wasn't intended for.

adyoungblood:
Ah. That would be my fault, I'm the one who made moving the default. I'll look into it right now!

Phergus:
I bet Melek has his preferences set to Players Can Edit Macros.

adyoungblood:
This is interesting, the code is already written to use the Players Can Edit Macros setting when creating a new macro, instead of whatever the setting was on the original macro. In that sense, I think this issue would be more of fix than a change?

adyoungblood:
> players shouldn't be able to view the code of locked/trusted macros, but they also shouldn't be able to put a trusted macro on a token it wasn't intended for.

Looking at the code, it should be possible to stop players from even moving trusted macros, while fixing the original issue and retaining intended functionality. The only question is how to convey to a player that that macro cannot be moved (I'm thinking some kind of exception).

cwisniew:
[...]

PR #2787, opened 2021-06-25
Fix issues with moving and copying macros for #2785

adyoungblood:
The bug where non-player-editable macros could be moved and suddenly become player-editable is fixed. More importantly though, players can no longer move macros into or out of the Campaign or GM panels. There is definitely reason to allow players to copy macros _out_ of the Campaign panel, but usually a GM might want to keep those macros to themselves. These changes (should) remove any need to restrict the moving/copying of non-player-editable macros in a more general fashion.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2787)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/macrobuttons/buttongroups/ButtonGroup.java
+++ b/src/main/java/net/rptools/maptool/client/ui/macrobuttons/buttongroups/ButtonGroup.java
@@ -131,7 +131,8 @@ public class ButtonGroup extends AbstractButtonGroup {
       if (!MapTool.getPlayer().isGM()
           && (panelClass.equals("CampaignPanel")
               || panelClass.equals("GmPanel")
-              || data.panelClass.equals("CampaignPanel")
+              || (data.panelClass.equals("CampaignPanel")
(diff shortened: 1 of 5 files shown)


adyoungblood:
Relevant [issue](https://github.com/RPTools/maptool/issues/2785)

melek:
> There is definitely reason to allow players to copy macros out of the Campaign panel, but usually a GM might want to keep those macros to themselves. 

Like you say, I (and maybe others) do put a few macros on the Campaign panel that players are encouraged to copy to their tokens (namely, I have a Spellbook macro there that they can copy for easy access). 

Personally, keeping the trusted status/player editable setting consistent would be sufficient for me - I would like to see that as a server setting if it is changed.

Phergus (approved):
:lgtm:

> Reviewed 4 of 5 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2787#-:-Md2j_b31aQDHYwiyDX1:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adyoungblood)

PR #2811, opened 2021-06-28
Fix log dir always being in user.home/.maptool-rptools

### LINKED ISSUE #2786: Change logs directory (Linux)
**Describe the bug**
Changing the Data Directory in the MapTool config does not change where the logs end up.

**To Reproduce**
1. Change Data Directory
2. Run MapTool
3. Note logs still go to `<userhome>/.maptool-rptools/logs`

**Expected behavior**
Logs should go to the MapTool Data Directory.

**Screenshots**

**MapTool Info**
- Version: Unknown

OS: Arch Linux
I installed MapTool [from the AUR](https://aur.archlinux.org/packages/maptool-bin/)

**Additional context**
Changed to Bug from original Question issue.

> I was able to move almost everything from `~/.maptool-rptools` to a different directory by changing Data Directory in MapTool preferences. However, `~/.maptool-rptools/logs/` persists. Is this intentional? Is there a way to relocate the logs directory as well?

Irarara:
Fixes #2786.

As per the discussion in the issue, the `appHome` property has been replaced with something more appropriately named  (`MAPTOOL_LOGDIR`) and the property is set before the logger is initialized so it can be used to set the logger's path. I also just removed the line that was for logging the value of `appHome` since it didn't seem to be adding much (the log file path is still logged).

One thing I noticed is in `getAppHome()` there is a chance of it attempting to use the logger before it's initialized when it's unable to create the data dir ([here](https://github.com/Irarara/maptool/blob/da0fa49790dc37d7140335c97c3c2267c868b2a1/src/main/java/net/rptools/maptool/client/AppUtil.java#L125)). I didn't touch this in case there's something I'm missing, but if the data dir can't be created there should be nowhere to log to anyway, right? Should that bit just be removed? To note though, that being reached doesn't actually affect anything, because being unable to create the data dir results in a "Failed to launch JVM" error either way (would be nice if it popped a proper error message, but that's a separate matter).

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2811)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppUtil.java
+++ b/src/main/java/net/rptools/maptool/client/AppUtil.java
@@ -44,13 +44,19 @@ import org.apache.logging.log4j.Logger;
 
 /** This class provides utility functions for maptool client. */
 public class AppUtil {
-
   public static final String DEFAULT_DATADIR_NAME = ".maptool";
   public static final String DATADIR_PROPERTY_NAME = "MAPTOOL_DATADIR";
-  private static final Logger log = LogManager.getLogger(AppUtil.class);
+  public static final String LOGDIR_PROPERTY_NAME = "MAPTOOL_LOGDIR";
   private static final String CLIENT_ID_FILE = "client-id";
   private static final String CONFIG_SUB_DIR = "config";
   private static final String APP_HOME_CONFIG_FILENAME = "maptool.cfg";
+
+  static {
+    // Don't move this. This MUST be set before the logger is initialized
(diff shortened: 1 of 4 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2811#-:-MdIBYTC8gh4JcBptn4Q:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Irarara)

PR #2816, opened 2021-06-28
Handle exceptions thrown when trying to get the logs directory at launch

### LINKED ISSUE #2812: Bad data dir paths result in an unhelpful error message when launching Maptool
**Describe the bug**
If `MAPTOOL_DATADIR` is set to a bad path or the data dir otherwise can't be created the user gets the message "Failed to launch JVM" when launching Maptool and no other information. Exceptions during launch are already supposed to be handled so that more meaningful errors are returned (#1678), but in this case they slip through.

**To Reproduce**
1. In `MapTool.cfg`, set the `-DMAPTOOL_DATADIR` System Property to a bad path (something with an illegal character or that you don't have write access to, etc.)
2. Launch Maptool
3. Observe that "Failed to launch JVM" pops up.

**Expected behavior**
A meaningful error message should be displayed.

**Screenshots**

**MapTool Info**
- Version: develop/1.9.2 (but probably basically any)
- Install: Any

 - OS: Windows
 - Version 10

Irarara:
Fixes #2812.

Moved setting the `MAPTOOL_LOGDIR` property and initializing the logger into a method (`initLogging()`) that is called at launch so any exceptions that might get thrown can actually be handled. `getAppHome()` is still called first in `LaunchInstructions.main()`, and while it may look superfluous (and it kind of is), it's there so that if the data dir itself can't be created the error shown will reflect that (as opposed to being about the logs dir specifically), it just seems clearer that way.

Additionally, `exit()` is now called when an exception is caught so that it exits properly, and I removed a redundant check for the data dir in `MapTool.main()`.

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppUtil.java
+++ b/src/main/java/net/rptools/maptool/client/AppUtil.java
@@ -51,12 +51,7 @@ public class AppUtil {
   private static final String CONFIG_SUB_DIR = "config";
   private static final String APP_HOME_CONFIG_FILENAME = "maptool.cfg";
 
-  static {
-    // Don't move this. This MUST be set before the logger is initialized
-    System.setProperty(LOGDIR_PROPERTY_NAME, getAppHome("logs").getAbsolutePath());
-  }
-
-  private static final Logger log = LogManager.getLogger(AppUtil.class);
+  private static Logger log;
 
   /** Returns true if currently running on a Windows based operating system. */
   public static boolean WINDOWS =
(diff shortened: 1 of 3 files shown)


Phergus (approved):
:lgtm:

> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2816#-:-MdJ44Ym8Z7IAH41VM4M:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Irarara)

Issue #2844, opened 2021-07-05
MBL is not strict


elfofthefire:
**Describe the bug**
A player can move into a region that is marked as MBL (but not VBL.)  The player is shown a 0 for their move distance, and after releasing the drag, everyone's maptool freezes for a moment and then the token is placed in the blocked off area.

**To Reproduce**
Create a MBL only block.  Create a token, snapped to grid.  Assign token to player.  Player can drag token into MBL area.

**Expected behavior**
The movement is blocked.

**MapTool Info**
- Version: 1.9.2

**Desktop (please complete the following information):**
 - OS: Windows
 - Version 10

Phergus:
Thanks for jumping in to help with MapTool.  Volunteers help keep the development moving forward.  However, you really should talk to the developers somewhere between opening an issue and submitting a fix. 

The current MBL behavior is intentional as far as allowing players to force movement into the area.  It's only there to influence the pathfinding code for normal, i.e. walking, movement and not to actually stop tokens from being moved into an area.  There are several reasons for this.

- A glass wall that would stop normal movement but wouldn't stop something that uses short range teleport. Normal characters would have to walk around it.
- A pit that would need to be navigated around for normal characters but wouldn't inhibit a flying character.  Or perhaps a character wants to jump into it as they have an ability that lets them ignore falling damage.
- A burning area that any normal character would walk around but those with some kind of fireproofing could ignore.
- And many other examples...

Forcing the GM to handle those types of scenarios just slows things down.

Note also that current develop branch code already has a fix for the long delay when dropping a token in a block of MBL.

For general contact with the RPTools team, the MapTool Discord server is a good place to go: https://discord.gg/hbn2bfn

cwisniew:
I guess we could add some sort of toggle (only settable by GM) that enables/disables this behaviour for movement, that way its not all or nothing and the GM can decide how much it annoys them either way :)

Phergus:
Probably need a preference setting for the default state and either use that in the server policy or add yet another checkbox to the start server dialog that defaults to the preference setting.

FullBleed:
I don't have a problem with more server checkboxes... but I am wondering if they could use some organization and categorization so it doesn't seem quite so overwhelming and random when you're selecting them.  Something like this maybe?

Options
 
     Tokens
           Strict Token Management
           Allow Players to Impersonate any Token
           Movement Metric

     Vision/Blocking/FoW
          Use Individual Views
               Use Individual FoW
[...]

PR #2845, opened 2021-07-05
Enforcement of MBL

elfofthefire:
Change to prevent players from dropping owned (snap-to-grid) tokens into MBL only areas.  Before, the token's path would fail, show as 0 distance on-screen, then cause the server and clients to all freeze for a moment, but would then be moved to the MBL blocked area.

Now, only the GM will be allowed to drag a token into a MBL area.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2845)


elfofthefire:
fixes #2844

Phergus:
No response from author.  Closing.

PR #2856, opened 2021-07-11
Only update token type if the type has actually changed when editing a token

### LINKED ISSUE #2849: Has Sight always ticked for PC tokens
**Describe the bug**
When a token is set as a PC in the top right of the properties and saved, it automatically ticks "Has Sight" for the token every time the properties window is closed. This used to occur when the token was first set to PC, but it was possible to untick "Has Sight" and leave it off. Now, it appears to be ticking "Has Sight" regardless and it's not possible to has it off. This makes maps with manually revealed fog of war impossible to use

**To Reproduce**
1. Create a new campaign and token
2. Note that "Has Sight" is unticked
3. Change the token to be a PC
4. Save the properties, reopen the properties and note that "Has Sight" is ticked
5. Untick "Has Sight", save and close the properties
6. Reopen the properties, note that "Has Sight" has reticked itself - this is the bug

**Expected behavior**
"Has Sight" should remain unticked once it has been unticked

**MapTool Info**
- Version: 1.9.2
- Install: New

 - OS: Windows
 - Version 10

Irarara:
Fixes #2849. 

As the title says, this changes it so token type will only be updated if the token type actually changes to avoid `hasSight` getting set to `true` each time a PC token is edited. Also added a comment to document that side effect in `setType()`. Thanks for pointing me to the source of the issue, @Phergus!

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2856)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/token/EditTokenDialog.java
+++ b/src/main/java/net/rptools/maptool/client/ui/token/EditTokenDialog.java
@@ -607,7 +607,10 @@ public class EditTokenDialog extends AbeillePanel<Token> {
       return false;
     }
     // TYPE
-    token.setType((Token.Type) getTypeCombo().getSelectedItem());
+    // Only update this if it actually changed
+    if(getTypeCombo().getSelectedItem() != token.getType()) {
+      token.setType((Token.Type) getTypeCombo().getSelectedItem());
+    }
 
     // SIZE
     token.setSnapToScale(getSizeCombo().getSelectedIndex() != 0);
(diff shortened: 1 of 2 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2856#-:-MeNAzgi3gxGdDiAu4ao:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Irarara)

Issue #2876, opened 2021-08-08
add webrtc protocol for communication


thelsing:
**Is your feature request related to a problem? Please describe.**
The networking setup is a bit complicated for non-IT-persons and often an issue with provider-NAT and similar stuff even for IT-persons. 

**Describe the solution you'd like**
WebRTC hopefully offers a "just works" solution for the networking. Things like audio and video-calls could also be added to MT based on WebRTC.

**Describe alternatives you've considered**
Keep everthing as it is and live with it.

thelsing:
IMO checking that socket communication still works should be enough. Webrtc is only an additional option it doesn't replace socket communication.

Phergus:
No additional issues reported. Closing.

PR #2877, opened 2021-08-08
WebRTC (addresses #2876)

thelsing:
This adds experimental WebRTC support. It has to be enabled in the "Help"-Menu for Client and Server. For signalling I wrote a small node-js server: https://github.com/thelsing/mt-signaling 
The signaling server is hard coded to my free azure app-server instance. (ws://mt-test2.azurewebsites.net ) The machine is quite limited but should be enough for testing.

I did quite a bit of refactoring to support webRTC:
- I broke the inheritance of the Client/Server-Connection classes and replaced it with interfaces and decorator pattern.
- I refactored the Handshake-class to a message based approach and made handshaking asynchronous. The messages are defined in protobuf-file format. The gradle generates the necessary java code from them.
- I added names to all threads I could find. This makes debugging threads much easier.
- I added a Mapper class to map protobuf <-> pojo

I still need to do some minor cleanups and add the webrtc natives for macos and linux, but I would like to do this after I got some comments about the code.

#2876 

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2877)


Phergus:
@cwisniew   Should we maybe put this in a feature branch for testing?

thelsing:
IMO checking that socket communication still works should be enough. Webrtc is only an additional option it doesn't replace socket communication.

cwisniew:
Some comments.
I would prefer it if interfaces didn't start with I<Name> just because they are interfaces, it doesn't add much and arguably as its inconsistent with other interfaces in the code base is going to be more confusing if someone is expecting other interfaces to follow that naming standard.

cwisniew:
Added branch feature-webrtc

Created Repo https://github.com/RPTools/maptool-protobuf

thelsing:
I merged the mt-protobuf repo and added the handshake.proto file to it and renamed all I* interfaces. I did an test 
"git subtree push --prefix src/main/proto my-mt-proto develop"
This took quite a while and I had all commit of MT in my MT-probuf repository. So made a new clean branch of develop, squash merged the webrtc branch, removed the proto dir before commit and redid the subtree stuff. Now it worked as expected.

So we have to make sure to never ever have commit that touch both the proto dir and other files.

I had to "git push -f" my changes so you have the recreate the feature-webrtc branch.

cwisniew:
branch feature-webrtc recreated

cwisniew:
I think it might be easier just to use git sub modules, sure it means cloning requires an extra flag but then at least we dont have to worry about all the PRs coming back with commits messed up

thelsing:
With submodules contributors have to split pull-requests that touch proto-files. And you still have to make sure that the changed .gitmodule doesn't get into a pull request. 

I tried submodules in a different projects and didn't like them.

cwisniew:
But as you said if we have to make sure that we never touch the proto and MT files at the same time for git sub tree how do we even manage that? From a developer perspective its never going to happen, someone changes token, they are not going to commit then change the proto file majority of the time they will do it in one commit.
You cant push the work of checking every commit and rejecting those that don't follow this rule back on to @Phergus  and myself thats just unreasonable, and the consequences of getting it wrong means fiddly and potentially long time for recovery. If peoples changes are rejected with no easy way to fix it thats also going to annoy them...

thelsing:
There is also https://github.com/ingydotnet/git-subrepo . But I didn't test it yet.

cwisniew:
> There is also [ingydotnet/git-subrepo](https://github.com/ingydotnet/git-subrepo) . But I didn't test it yet.

I looked at it, and it seems that would require contributors to use command line (won't work through their IDEs / GitHub desktop etc). Which for many I am guessing is going to be more of a stumbling block than 2 PRs.

thelsing:
This just keeps subtree and submodule a options.

subtree:
+ development in MT-repo is easier
+ pull requests with changes of proto files work a they do now (only one pull request)
- subtree pushes are quite slow (currently abount 10 min on my quite old laptop)
- if you have the subtree dir and other files in the same commit, you push the preceding commits of the other files also to the protobuf repo. Those commits are NOPs there as they don't touch anything in the protobuf repo. This could probably be checked via a github action on pull-requests.

submodule:
+ easy to set up
- contributors have to change the submodule to their local fork if the change anything on the proto files. They have to make sure that they don't put this change into the pull request. We also have to check for this on pull requests.
- contributors have to make a different pull-request for the proto-files. This has to be accepted first. Only after this a contributor can submit a pull request that builds. (He can now push an update to the correct commit hash in the submodule file). There is a risk that the follow up pull request of the MT-repo will never come. 

My vote is still on subtree. But I can also live with submodules.

cwisniew:
I think the other disadvantage for subtree is we are not in a situation where we will have to make changes to MT protos that don't get pushed up to the proto repo too, so there is always going to be "double handling". Unfortunately there is no clear winner :( 

Maybe we do just for now include the proto files in MapTool registry then when a significant amount of the proto work has been done we can get a better handle on the workflow (plus once its nearly all done the number of large changes that touch both is going to be far less significant). Once we have a better handle on it we can make a better decision on which of subtree/module/gradle dependency would work best.

thelsing:
That's ok too. In the pull request I "git-subtree-added" the mt-proto repo. So we can push back there if we want. There are even github actions that can do so automatically.

cwisniew:
@thelsing do you mind just resubmitting against feature-webrtc branch with all in one repo, I tried changing the branch in this pull request but that lead to all sorts of funkiness in the review as its somehow outta whack.

Thanks

thelsing:
Further discussions in #2892 (Resubmit as requested.)

PR #2925, opened 2021-08-28
webrtc bugfix and signaling via rptools.net

thelsing:
Changes:
- fixes bug where only one client can connect to server
- changed signaling server to rp-tools.net

With this changes webrtc be reasonable stable.

For #2876 

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2925)


Phergus:
With these changes it is working for me now.  Didn't do much beyond move some tokens around, run a few macros, add a drawing and some VBL.

Issue #2741, opened 2021-06-16
<select> element with multiple attribute containing an <option> element prevents form from being submitted


dark-ether:
**Describe the bug**

 a \<select\> element with the multiple attribute will, if it contains an \<option\>\<\/option\>\ element, breaks the submit button which submits the form.

**To Reproduce**
1. in a new campaign create a token called Lib:test
2. in the Lib:test token create 2 macros called link and test
3. in the link macro copy  and then paste `[r:broadcast(macro.args)]`
4. in the test macro copy and  then paste  
```
[h:link = macroLinkText("link@Lib:test","all")]

[frame5("test"):{

	<form action="[r:link]" method="json">
	
	<select id="test" name="test" multiple>
			<option value="option">option</option>
	</select><br/>
	
	<input name="submit" type="submit" value="sucess" name="submit">
	
	</form>
}]
```
5. click in the test macro and in the opened frame click the button

**Expected behavior**
after doing these steps a json element containing the key value pair submit:sucess should appear in chat 

**MapTool Info**
- Version: 1.9.0 and 1.9.2
- Install: the Deb package

**Desktop (please complete the following information):**
 - OS: linux Pop! Os 
 - Version 20.10

**Additional context**
 submit buttons which submit other forms in the same frame still work

kwvanderlinde:
There are several things going on here, some of which is expected and some of wh.

First of all, there are a couple of changes that should be made to your code:
1. Use the `r` option on the `foreach` so that it doesn't try print tooltips or anything.
2. Change the "link" macro to broadcast instead of roll:
   ```
   [h: broadcast(macro.args)]
   ```
   Otherwise the output won't make it to chat.

But even with those fixed, there is certainly an issue with `multiple`. I did some debugging on Linux Mint 20, and found that [this line](https://github.com/RPTools/maptool/blob/14353f889a5452f45a3ce05eecb6115a340a2e92/src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLWebViewManager.java#L620) throws an exception. After adding a catch-all, this is what I found:
```
java.lang.UnsatisfiedLinkError: 'long com.sun.webkit.dom.HTMLOptionsCollectionImpl.itemImpl(long, int)'
	at javafx.web/com.sun.webkit.dom.HTMLOptionsCollectionImpl.itemImpl(Native Method)
[...]

PR #2933, opened 2021-09-01
Work around missing JavaFX symbols on linux.

kwvanderlinde:
Works around #2741.

On linux, the JavaFX native webkit libraries are missing the necessary symbol that provies the implementation of `HTMLOptionsCollectionImpl::itemImpl`. As a result, we are unable to access the items returned by `HTMLSelectElement::getOptions()`, and thus can't use it to determine which options of a `<select>` are selected.
    
To workaround this, we take a dependency on the `HTMLSelectElementImpl`, the JavaFX implementation of `HTMLSelectElement`. It has a method `item()` that is not exposed in the `HTMLSelectElement` interface, but which provides access to the associated options. Ideally we'll remove this once JavaFX is fixed, so we don't keep a dependency on internals.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2933)


cwisniew (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2933)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

Issue #2155, opened 2020-08-07
FogUtil perf improvement


ebudai:
Path2D is faster than Area, it can be used in calculateVisibility()

Phergus:
@ebudai  Please use one of the templates and be more descriptive.

Pretty sure that Path2D has been explored in the past and was not used for _reasons_.   Perhaps this has changed now that we are on Java 14.   @JamzTheMan ?  @cwisniew ?

ebudai:
My tests show a 5% improvement.  I only got the idea looking at another area of the code, but if it introduces a bug, certainly close this.

Also, I will resubmit the issue if you like.

JamzTheMan:
Area is the slowest. I found the same and if we ripped area out everywhere we would see more improvements.

In fact, using the jst lib may even yours better performance.

ebudai:
I'll have a look at incorporating JTS.

ebudai:
I have to update Hessian IO, but its used by the clientserver project.  Fine to update that one also?

Phergus:
First step is to open a ticket on the clientserver project explaining the reason to update Hessian IO.  Once clientserver has been updated, a release can be done for it at which point the MapTool build can be updated to use the new release.

Second step will of course be testing that all aspects of client/server communication are working correctly.

Azhrei:
I've tried updating the Hessian server-side stuff before.  It was a whole can of worms, with one error cascading into other areas.  So instead I've been patching the code when it breaks (as PHP has gone from v5.2 to 7.1).

I'd rather get rid of the Hessian library for talking to the web site and use JSON/REST instead.  We'd have to keep existing Hessian in place on the server for older clients, but we could switch to JSON and a new URL for newer code.

ebudai:
I made a fork, and had no issues updating to the latest Hessian.

Azhrei:
The error was not related to the updating — that's just copying files into place.  It was related to running the PHP code.

I suppose you could modify the URL in the MapTool class so it points to your own private web server...?

ebudai:
It sounds as if updating the website and related comms code should be its own issue, which would block this one.

cwisniew:
> I made a fork, and had no issues updating to the latest Hessian.
> 
> edit: that I know of - the error I was getting went away. I did not test any client/server comms - is there an easy way for me to check that out?

@ebudai I guess there is two parts to test

[...]

PR #2934, opened 2021-09-02
Change `VisibleAreaSegment::getPath()` to return `Path2D`.

kwvanderlinde:
Addresses #2155, at least partially.

Measurements were showing that `PointerTool::exposeFoW()` was spending about 60% of its time in `VisibleAreaSegment::getPath()`. This methods constructs an `Area` from a path, but the callers only care about getting the bounding box of the `Area`, and not the `Area` itself.

This change just skips the creation of the `Area` and instead returns the much cheaper `Path2D` that is created. Since the conversion to `Area` is quite expensive, this eliminates almost all of the running time of `getPath()`, and yields about a 2x speedup for `exposeFoW()`.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2934)


cwisniew (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2934)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #2948, opened 2021-09-07
Use JTS for calculating visibility

kwvanderlinde:
Relates to #2155 

For the most part, `VisibleAreaSegment` no longer deals in `Area`s, but `Geometry`s. `FogUtil::calculateVisibility()` takes advantage of this through `PreparedGeometry`. Although `PreparedGeometry` does not help with the many potential union operations, it does offer significant speed ups to the containment and intersection checks. JTS also helps somewhat with the union operations of plain `Geometry`s compared to AWT `Area`s.

The boundary of `VisibleAreaSegment` was previously provided by `getPath()`. This has been removed in favour of only providing a bounding box, as that is all `getPath()` was used for. We also don't bother constructing a boundary path anymore and instead calculate the bounding box just from the points in the boundary.

Some redundant AWT `Area` operations were also identified in `exposeLastPath()`. Snap-to-grid tokens would reveal from each position twice, while free-moving tokens would reveal from the final position once for each point in the path. A free-moving token would typically only have two points in its path, unless it is dragged together with a snap-to-grid token in which case it may have many more points in its path. The two cases have now been distinguished and modified so that snap-to-grid tokens only reveal once for each position, and free-moving tokens only reveal once from their final position.  The types were also cleaned up and an invalid supressed unchecked cast was removed in favour of generality.

Testing was done using snap-to-grid tokens from specific start and end points two particular maps, with Fog of War automatically exposed at all cells (i.e., not only at waypoints). The first map was a 70x70 dd2vtt import containing a hand drawn cave system. The second was a 100x100 dd2vtt import of dense randomly generated caves. All-in-all, there was about a 23% improvement in `exposeLastPath()` times for the 70x70 map, and about a 38.5% improvement for the 100x100 map.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2948)


cwisniew (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2948)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2948#-:-Mixcsfs3PFUh_iP8m_V:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

Issue #2260, opened 2020-10-12
Offset grid: pasted tokens no longer go in the cell we want


dorpond:
**Describe the bug**
If I adjust the grid, specifically moving the offset of it, pasted tokens will not go where they should.

**To Reproduce**
- Create a new map. Use defaults, doesn't matter.
- Adjust Grid. No need to change the size of the grid, just move the grid offset around some.
- Paste a token. It no longer goes in the cell you are pointing at.

**Expected behavior**
pasted items should always go into the cell your mouse is currently pointing at.

**MapTool Info**
- Version: 1.7.0 (EXE)
[edit] Notice it also happens with 1.5.6

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version [10 1909]

Phergus:
I thought this was fixed in 1.8 but it's still there.

Phergus:
Still present in 1.9.0

kwvanderlinde:
This bug should be fixed since 1.10.

Phergus:
It is except for one weird case.  If you have tokens on the map, select one and hit Copy, then adjust the grid so that the offsets are both negative some amount and now paste a token it is offset by one cell from the mouse pointer.

Also weird, I can set the offset manually to positive values but dragging the grid only results in negative offsets.  When I manually set the offset to a positive value and pasted a previously copied token it went where I pointed.

kwvanderlinde:
> It is except for one weird case. If you have tokens on the map, select one and hit Copy, then adjust the grid so that the offsets are both negative some amount and now paste a token it is offset by one cell from the mouse pointer.
> 
> Also weird, I can set the offset manually to positive values but dragging the grid only results in negative offsets. When I manually set the offset to a positive value and pasted a previously copied token it went where I pointed.

It also effects positive offsets, we just get lucky and those tend to land in the same cell anyways. But crank it to be larger than the grid size and tokens will be pasted the lower-right of the cursor.

kwvanderlinde:
There's one more odd thing I found, though it might be by design.

Tokens that are snap-to-grid will use their cell offset from the top-left token when pasted. This means if we adjust the grid size between copying and pasting, the tokens will spread out or be pulled in accordingly, keeping the number of cells between them constant. But we don't scale other token positions, so the relative positioning between snap-to-grid and non-snap-to-grid can be surprisingly changed.

E.g., I start with a 50x50 grid with one non-snap-to-grid token:
[...]

PR #2959, opened 2021-09-10
Fix for pasted tokens not being placed in the correct cell

kwvanderlinde:
Addresses #2260.

There are two separate problems that together lead to some rather inconsistent behaviour:
1. Converting from `ZonePoint`to `CellPoint` needs to round to the nearest cell, which has slightly different logic for
   positive and negative positions. The choice was being done on the unshifted `ZonePoint` which does not correspond to
   whether the position is logically negative or positive. Thsi resulted in off-by-one errors when x or y is near zero.
2. When copying tokens, their position relative to the "top-left" token is remembered. Upon pasting, it is assumed that
   this relative position can be treated as a `ZonePoint`, including being converted to a `CellPoint`. This fails to
   account for the grid offset (or rather, the grid offset is applied during the conversion when it shouldn't
   be). Essentially, we're treating a vector as a point, and vice-versa.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2959)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppActions.java
+++ b/src/main/java/net/rptools/maptool/client/AppActions.java
@@ -159,7 +159,16 @@ import org.apache.logging.log4j.Logger;
 public class AppActions {
   private static final Logger log = LogManager.getLogger(AppActions.class);
 
+  /**
+   * Hold copies of tokens for a future paste.
+   *
+   * <p>Note that the tokens stored in this set have had their positions modified to be measured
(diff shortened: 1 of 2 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2959#-:-MjEaAtT624ibktOR27i:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

Issue #2970, opened 2021-09-12
Campaign properties can't be opened when a hex-shaped light is defined


kwvanderlinde:
**Describe the bug**
If a hex-shaped light is added to in the Campaign Properties, the Campaign Properties cannot be opened again because an NPE is thrown.

**To Reproduce**
1. Open a blank campaign.
2. Open the Campaign Properties dialog and go to the Light tab.
3. Edit the "Candle" entry to be a hex-shaped light:
    ```
    Candle - 5: hex 5 10#000000
    ```
4. Push the OK button
5. Try opening the Campaign Properties dialog again.
6. See an exception dialog.

**Expected behavior**
Hex lights can be defined without breaking the Campaign Properties dialog.

**MapTool Info**
- Version: 1.9.3
- Install: New

**Desktop (please complete the following information):**
 - OS: Linux Mint
 - Version 20

**Additional context**
<details>
    <summary>Exception details</summary>

```
java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because "lastShape" is null
	at net.rptools.maptool.client.ui.campaignproperties.CampaignPropertiesDialog.updateLightPanel(CampaignPropertiesDialog.java:378)
	at net.rptools.maptool.client.ui.campaignproperties.CampaignPropertiesDialog.copyCampaignToUI(CampaignPropertiesDialog.java:225)
	at net.rptools.maptool.client.ui.campaignproperties.CampaignPropertiesDialog.setCampaign(CampaignPropertiesDialog.java:212)
	at net.rptools.maptool.client.AppActions$72.executeAction(AppActions.java:3025)
	at net.rptools.maptool.client.AppActions$ClientAction.execute(AppActions.java:3396)
	at net.rptools.maptool.client.AppActions$ClientAction.actionPerformed(AppActions.java:3428)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1972)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2313)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:405)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
	at java.desktop/javax.swing.AbstractButton.doClick(AbstractButton.java:374)
	at de.muntjak.tinylookandfeel.TinyMenuItemUI.doClick(Unknown Source)
[...]

PR #2971, opened 2021-09-12
Handle HEX when updating the Light tab in Campaign Properties

kwvanderlinde:
Addresses #2970

"hex" is accepted as a valid shape when editing lights, and hex-shaped lights will work when attached to a token. But when a hex light is defined, the campaign properties dialog cannot longer be opened due to a `NullPointerException` This changes fixes that by adding `HEX` as a recognized shape when updating the lights.

There is also a slight restructuring: if we encounter an unrecognized light shape, we now deliberately throw a `RuntimeException` rather than allowing a `NullPointerException` further along. This has the perk that we don't need to initialize `shape` to `null`, because Java recognizes that all code paths will initialize `shape` before it is read.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2971)


cwisniew (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2971)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

cwisniew (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2971)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #2973, opened 2021-09-14
Fix JTS geometry errors in calculateVisibility()

kwvanderlinde:
Improves #2155 

There are lots of ways in which cleared area geometry can become invalid, even though it is composed of many valid polygons. Operations like unions are particularly susceptible to breaking as a result. To avoid issues with unions of cleared area while also avoiding excessive buffering, we now build a `GeometryCollection` containing all of the cleared area geometry. We then `buffer(0)` that, which has the effect of creating a single valid geometry without growing the geometry at all.
    
In order to make this work robustly without having to buffer each individual cleared area, we no longer try to union individual cleared areas together and instead only union the many cleared areas in a single `UnaryUnionOp`.  Although this can theoretically limit the number of culled segments, performance actually improves in practice.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2973)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/FogUtil.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/FogUtil.java
@@ -59,6 +59,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.locationtech.jts.awt.ShapeWriter;
 import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.geom.GeometryFactory;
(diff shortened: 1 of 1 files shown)


cwisniew (changes requested):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2973)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

cwisniew (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2973)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

Issue #2755, opened 2021-06-19
Add two-way VBL/terrain walls


adyoungblood:
**Is your feature request related to a problem? Please describe.**
When setting up new maps with Fog of War enabled, I am always forced to choose between obscuring the art of the walls with accurate VBL, or creating VBL that approximates what should be there, while showing as much of the actual walls as possible. Also, I have to put an X across objects such as pillars or tall boxes, which gets the job done, but looks ugly.

**Describe the solution you'd like**
Terrain walls, straight out of [Foundry](https://foundryvtt.com/article/walls/).

**Describe alternatives you've considered**
- Doing what I've always done (which works, sorta)
- Allowing players to simply see the whole map
- Manually removing the FoW from the walls I wish the players to see (time consuming)

**Additional context**
Let's say I have my test subject Agatha the hag standing in the Yawning Portal, looking towards the bar wall.
![image](https://user-images.githubusercontent.com/16092423/122628010-59ff1e80-d068-11eb-8244-a88ebca2e592.png)
In the GM view, I can see the wooden wall that blocks her vision. However, in the player view, the wall is simply a fog-colored rectangle:
![image](https://user-images.githubusercontent.com/16092423/122628029-7602c000-d068-11eb-9bef-a1a2aa9318ad.png)
I would rather she (and by extension, my players) be able to see the nice wall art underneath.

Additionally, on a map such as this one
![image](https://user-images.githubusercontent.com/16092423/122628054-a34f6e00-d068-11eb-9514-7bd8547b640a.png)
it's rather difficult to place VBL that is accurate to the map. but also reveals what's back there (specifically, the place where the outer wall merges with the inner wall).

emmebi:
A few things which you can try here:
1. use a VBL/MBL in the middle of the wall to block the movement and the line of sight
2. put a VBL on the whole wall, and then reveal it using the Fog of War tool; this requires some effort, but provides some good results.

adyoungblood:
I've typically gone with the first method, although the second method does seem rather appealing now that you mention it's . Nevertheless, I believe that terrain VBL (and perhaps more advanced VBL in general, although that's a separate topic) could be useful to most users.

adyoungblood:
[...]

PR #2976, opened 2021-09-18
Issue 2755 terrain vbl

kwvanderlinde:
Implements #2755. Intentionally leaving as a draft for now as there is remaining work to do.

This change implements "terrain VBL", a new kind of VBL that allows vision into but not out of the area. Using terrain VBL allows players to see walls, pillar, landscapes, etc., while still not allowing them to see behind the VBL.

This PR has the following features:
1. Add a new kind of topology called "terrain VBL" which allows vision into the area, but not out.
2. If the option is selected for AI to navigate around VBL, then the AI will also navigate around terrain VBL.
3. A new toggle button has been added to the toolbar to enable drawing terrain VBL instead of regular VBL.
4. Tokens can be marked as terrain or not terrain.
    * If marked as terrain, the token's VBL is terrain VBL, otherwise it is regular VBL.
5. Terrain VBL can be transferred to/from terrain tokens.
    * Terrain VBL cannot be transferred to/from non-terrain tokens
    * Regular VBL cannot be transferred to/from terrain tokens.
6. Existing VBL functions (`drawVBL()`, `eraseVBL()`, `getVBL()`) now accept/return both regular and terrain VBL.
    * Terrain VBL is identified by `"terrain": 1` in the shape JSON.
    * Regular VBL is identified by `"terrain": 0` in the shape JSON, or if the `"terrain"` key is missing.
7. Existing token VBL functions (`getTokenVBL()`, `setTokenVBL()`, `transferVBL()`) respect whether the token is a terrain token.
    * `transferVBL` respects the rules in (4). So it only operates on terrain VBL if the token is a terrain token. Otherwise, it only operates on regular VBL.
    * If `setTokenVBL()` is given a shape with a `"terrain"` key, then it will only apply terrain VBL if the token is a terrain token, and will only apply regular VBL if the token is a non-terrain token.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2976)


Phergus:
Merged to branch feature-terrain-vbl for testing.

PR #2995, opened 2021-10-01
Reserve namespaces for future RPTools libraries

### LINKED ISSUE #2994: Reserve namespaces for future RPTools libraries
With the addition of lib:// URIs we should reserve some namespaces so we can add builtin/standard libraries in the future and not have to worry about conflicts.

The following lib:tokens should not allow players/GMs to set the allows URI access flag.

Lib tokens with the following names (after the lib: part)
* rptools
* maptool
* maptools (because people often call it maptools)
* internal
* builtin
* standard

Lib tokens starting with the following names (after the lib: part)
* rptools.
* maptool.
* maptools. (because people often call it maptools)
* tokentool.
* net.rptools.
* internal.
* _
* builtin.
* standard.
* .

Lib:Tokens with these names would still function as they currently do so that we don't create any compatibility issues, but should not work for lib:// functionality (and possibly other future functionality)

Not reserving the names which will lead to problems in the future if we want to use them.

**Additional context**
Once library are more flexible and offer more functionality we will probably want to include standard ones with MapTool, it also provides a clear indication if it's a standard library from the RPTools team.

cwisniew:
resolves #2994

### Description of the Change

Stops players from setting the allows URI access flag for lib:tokens with names that match one of the following.

Lib:tokens with the following names (after the lib: part)
* rptools
* maptool
* maptools (because people often call it maptools)
* internal
* builtin
* standard

Lib:tokens starting with the following names (after the lib: part)
* rptools.
* maptool.
* maptools. (because people often call it maptools)
* tokentool.
* net.rptools.
* internal.
* _
* builtin.
* standard.
* .

This will stop the player/GM from setting this flag in the UI and macro functions.

CleanShot 2021-10-01 at 22 18 00@2x

CleanShot 2021-10-01 at 22 16 42@2x

CleanShot 2021-10-01 at 22 16 33@2x

CleanShot 2021-10-01 at 22 16 20@2x

CleanShot 2021-10-01 at 22 16 13@2x

### Possible Drawbacks
There is a small chance that some groups have already built a release to use from the develop branch and are using the above names and will have to make changes before moving to the 1.10 release.

### Documentation Notes
See description for list of criteria where the URI access can not be set on a lib:token.

### Release Notes
N/A (this is a restriction to change in upcoming release not a feature or bug fix in of itself)

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2995)


Phergus (approved):
:lgtm:

> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/2995#-:-Mkxn8AB5haMeCyjjjBT:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cwisniew)

Issue #3011, opened 2021-10-06
js.eval, js.evalNS, and js.evalURI use case sensitive comparisons for determining which to run


cwisniew:
**Describe the bug**
The code for js.eval, js.evalNS, and js.evalURI all check `functionName` using a case sensitive comparison. As the parser does not check for case when calling the functions this leads to confusing issues if the user attempts to run one of these functions using the wrong case (e.g. js.evalns()).

**To Reproduce**
Try to execute js.evalURI as js.evaluri and you will get an error as it attempts to parse the URI as javascript.

**Expected behaviour**
To be consistent with the parser and other MTScript functions these should accept any case.

**MapTool Info**
- Latest develop branch

kwvanderlinde:
It's worth noting that this isn't specific to the JS functions. E.g., `getvbl()` will similarly generate an "Unknown function" error. In fact, it seems most of our functions use case-sensitive comparisions based on this quick estimate:
```
> grep -rn . -e 'functionName' | grep '.equals(' --color=always | wc -l
289
> grep -rn . -e 'functionName' | grep '.equalsIgnoreCase(' --color=always | wc -l
153
```

Issue #3010, opened 2021-10-06
Inform user if URI access is not available when trying to use js.evalURI


cwisniew:
**Is your feature request related to a problem? Please describe.**
When a user uses js.evalURI() to run javascript code MapTool should let them know if the lib:token in question does not have Allow URI Access checked.

**Describe the solution you'd like**
If a user uses js.evalURI() and the lib:token for that URI does not have the "Allow URI Access" flag selected then MapTool should let the user know that this is the case.

**Describe alternatives you've considered**
Don't do it :)

PR #3012, opened 2021-10-06
Updates for lib:token js.evalURI

cwisniew:
Addresses #3010 
Addresses #3011 

### Description of the Change
If a user attempts to use js.evalURI() to access a lib:token that exists but does not have the 'Allow URI Access' flag set they will get an error message informing them that that flag is not set.

Also updated 
* code to fix the bug (#3011) where using lowercase produces unexpected results.
* Cleaned up error messages that are returned when trying to execute from a lib:// URI so that they only display error message and not exception where it makes sense.

### Documentation Notes
No additional documentation required this is just displaying a more informative error message for something that was already displaying an error and aligning macro case with what users already expect.

### Release Notes
N/A

### DIFF
--- a/src/main/java/net/rptools/maptool/model/framework/LibraryToken.java
+++ b/src/main/java/net/rptools/maptool/model/framework/LibraryToken.java
@@ -23,11 +23,12 @@ import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import java.util.stream.Collectors;
 import net.rptools.maptool.client.MapTool;
+import net.rptools.maptool.language.I18N;
 import net.rptools.maptool.model.GUID;
 import net.rptools.maptool.model.MacroButtonProperties;
 import net.rptools.maptool.model.Token;
+import net.rptools.maptool.model.framework.LibraryNotValidException.Reason;
 import net.rptools.maptool.util.threads.ThreadExecutionHelper;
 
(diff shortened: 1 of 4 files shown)


src/main/java/net/rptools/maptool/model/framework/LibraryToken.java
```diff
@@ -162,9 +163,16 @@ private static Library findLibrary(URL path) {
       List<Token> tokensFiltered =
           zone.getTokensFiltered(t -> name.equalsIgnoreCase(t.getName())).stream()
               .filter(Token::getAllowURIAccess)
-              .collect(Collectors.toList());
+              .toList();
       if (tokensFiltered.size() > 0) {
         return new LibraryToken(tokensFiltered.get(0).getId());
+      } else { // Check to see if URI access is not allowed and throw error
+        List<Token> tokens =
+            zone.getTokensFiltered(t -> name.equalsIgnoreCase(t.getName())).stream().toList();
```
kwvanderlinde, line 171: Since we only check whether or not the list is empty, there's no need to convert to a stream and back.

Also, we could save the unmodifiable list above (the one returned by the first call to `getTokensFiltered()`) and avoid regenerating it here with repeated logic.

cwisniew: You are correct I was trying to get it done in a hurry over lunch and fell afoul of the dreaded cut and paste/modify slightly curse. I will update after work tonight.

PR #3018, opened 2021-10-07
GraalVM JS refactoring.

perkinslr:
#2519 

### Description of the Change

`js.eval` now uses a trusted anonymous namespace if called from a trusted MTScript context.
`MapTool.clientInfo.clientId` is properly annotated for javascript access.
`Token.get<X/Y>` now properly enforce owner-only or trusted-namespace restriction.
`MapTool.getSelected` (and related functions) moved into `MapTool.tokens.getSelected`
`MapTool.tokens.getMapTokens` can be given a map name to search.

### Possible Drawbacks

Existing js.eval scripts using `MapTool.getSelected` or similar will need to migrate to `MapTool.tokens.getSelected`.

### Release Notes
- Fixed an issue where `MapTool.clientInfo.clientId` was not accessible
- Added an option to specify a map by name to `MapTool.tokens.getMapTokens`
- Moved `MapTool.getSelected`, `MapTool.getSelectedTokens`, and `MapTool.getTokenByID` to `MapTool.tokens.*`

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3018)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/MacroJavaScriptBridge.java
+++ b/src/main/java/net/rptools/maptool/client/functions/MacroJavaScriptBridge.java
@@ -20,6 +20,7 @@ import java.math.BigDecimal;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.*;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
 import javax.script.ScriptException;
 import net.rptools.maptool.client.MapTool;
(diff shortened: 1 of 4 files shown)


cwisniew:
@Phergus Unless you can foresee any bad juju from this, I am ok with the change of namespace for the two functions, due to nashorn vs graalvm differences there will be minor tweaks people have to make anyway so this (and the fact there was no JS last build) is perfect chance to make that change that is easy for people to fix in their scripts.

PR #3040, opened 2021-10-12
Implement map terrain VBL

kwvanderlinde:
Implements #2755

### Description of the Change
Terrain VBL is much like VBL, except that vision can enter terrain VBL but not leave it. This has the effect of showing the map beneath the terrain VBL while still blocking what is behind it.

New features in this change:
- Can draw Terrain VBL on the map by hand, just like VBL and MBL. The 3-state VBL/MBL button is now a group of 5
  separate toggles for the VBL, Terrain VBL, MBL, MBL & VBL, and MBL & Terrain VBL. This makes topology mode selection
  more akin to topology shape selection, and is easier for the user to work with than cycling through five states on a
  single toggle button. It also has the perk that the implementation doesn't need to hack five states into a two-state
  toggle button.
- Can draw Terrain VBL on the map using new macro functions `drawTerrainVBL()`, `eraseTerrainVBL()` and
  `getTerrainVBL()`. These function analogously to `drawVBL()`, `eraseVBL()` and `getTerrainVBL()` respectively. Some
  refactoring was done in this area to reduce duplicated code and simplify the flow of the code when drawing
  shapes. Specifically, the `Topology_Functions.add*Topology()` methods have been replaced with
  `Topology_Functions.make*()` methods that only create `Area`s, while the caller is reponsible for adding the topology
  to the map.

Some minor bugs have also been fixed:
- When calling `getVBL()` and similar with an array of shapes, passing an empty array would cause a null pointer
  exception. This no longer occurs. Instead, an empty area is returned.
- `TokenVBL.renderVbl()` has been removed in favour of `renderTopology()`. This avoids a bug where transferring token
  VBL to the map depends on the current topology mode rather than always working with VBL.

Some implementation details have changed to support terrain VBL:
- `AreaTree.getOceanAt()` always returns an ocean for valid topology, even if the point is within an island. This
  enables vision to work within an area of terrain VBL, whereas the same case always blocked vision with regular VBL.
- `VisibleAreaSegment` can either be created from the front side or back side of an `AreaContainer`. Which side gets
  used depends on whether terrain VBL is in use and whether the point is within a given island or ocean.
- `GeometryFactory` instances are shared between `FogUtil` and `VisibleAreaSegment` in order to enforce consistent
  precision models. This is important as the default precision is far too prone to slight differences between what ought
  to be equal points. At the same time, we don't want to be too lax or else obviously distinct points can appear the
  same. We currently use 10e-6 as the precision.
- `net.rptools.maptool.client.ui.zone` package had the classes `AreaData`, `AreaMeta`, and `AreaFace`. These are
  redundant analogues to the classes under `net.rptools.maptool.client.ui.zone.vbl`. They were also unused except in
  some expository tests, and thus have been removed.

### Possible Drawbacks

- The group of topology mode buttons in the toolbar may a surprise to users already familiar with the 3-state toggle.
- The new implementation has more geometry to process and so users could see a performance hit with FoW.
- Increased use of JTS could introduce consistency errors.

### Documentation Notes
Terrain VBL is useful for marking obstacles that should block a token's vision, but where the GM still wants the players to be able to see the obstacle. Some typical examples would be:
- A pillar in a dungeon
- A large boulder on a battle map
- Houses in town when the PCs are outside.

Here is a concrete example. An enemy (Troll) hides from a PC (Elf) behind a dungeon pillar. This is what the GM sees, where the cyan area is terrain VBL:
image
The player owning Elf cannot see the troll, but can see the pillar itself. This makes it clear to the player that a pillar is blocking Efl's vision, and also lets them now the size and shape of the obstacle:
image
By contrast, if we had used VBL instead of Terrain VBL, the player would not be able to see what kind of obstacle is blocking Elf's vision. The pillar would just look like a "hole" in Elf's vision:
image

It may not always be desirable to show the entire obstacle. In that case, we can modify the example so that only the boundary of the pillar is covered in terrain VBL. Since the back edge of terrain VBL always blocks vision, this will have the effect of only revealing the near side of the pillar boundary, while hiding the center and the back side of the pillar. This is what the new terrain VBL looks like
image
Now the player can see the near side of the pillar, but can't see the backside, and still can't see anything behind the pillar:
image

### Release Notes

- Added Terrain VBL which blocks vision from the back instead of the front.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3040)


cwisniew:
@kwvanderlinde there are some merge conflicts with the change that updated to case insensitive checks for function names.

Issue #3037, opened 2021-10-11
Error with image url within frame5 environment when using getImage()/getTokenImage() with size parameter


Reznoov:
**Describe the bug**
When the source url of an image element is gotten using getImage() or getTokenImage() with the size parameter set, the url is not read correctly within a frame5 or any of its varieties. The result is no image is displayed.

This problem does not persist when using the legacy frame or dialog or a pure chat output.

**To Reproduce**
The following code does not work:
```[fram5("test"): {
<img src="[r:getTokenImage(100)]">
}]```

The following codes work:
```[fram5("test"): {
<img src="[r:getTokenImage()]">
}]```
```[frame("test"): {
<img src="[r:getTokenImage(100)]">
}]```

**Expected behavior**
The image should be displayed properly with resizing.

**MapTool Info**
- Version: 1.10.0

**Desktop (please complete the following information):**
All

**Additional context**
Previous MapTool version, image was displayed correctly.

PR #3046, opened 2021-10-14
The raw asset change breaks several existing functions for webp

cwisniew:

Addresses remaining part of #3037

This possibly affects other image types as well as webp, when returning an asset to be displayed in
html/html5 views the image will not be able to be displayed.

### Description of the Change
Unless the query parameter raw=true is specified, all assets fetched via the asset:// URL now default to the pre 1.10 behaviour where it will not fetch the raw asset. 

### Possible Drawbacks

This change will always fetch the png version of an asset unless the raw=true query flag is present. Unfortunately this introduces a small compatibility issues for anyone that started using animated GIFs between release of 1.10.0 and 1.10.2 but given the size of the compatibility issue of not requiring the
explicit request for raw breaking other exisiting functions this is by far the lesser of two evils.

### Documentation Notes
```html
[h: rawImg = getTokenImage() + "?raw=true"]
<img src="[r: rawImg]"/>
```

or 
```html
<img src="[r: getTokenImage()]?raw=true"/>
```

This is also detailed in Issue #3037 comments.

### Release Notes
- getTokenImage(), tblImage() etc work with webp images again.
- fetching a raw asset from URL requires ?raw=true query parameter

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3046)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AssetURLStreamHandler.java
+++ b/src/main/java/net/rptools/maptool/client/AssetURLStreamHandler.java
@@ -21,6 +21,7 @@ import java.io.InputStream;
 import java.net.URL;
 import java.net.URLConnection;
 import java.net.URLStreamHandler;
+import java.util.Arrays;
 import net.rptools.lib.MD5Key;
 import net.rptools.lib.image.ImageUtil;
 import net.rptools.maptool.model.AssetManager;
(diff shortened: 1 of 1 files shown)


cwisniew:
I am planning to delete the branch after release so no big deal its not 1.10.2 :)

Issue #233, opened 2019-02-11
Player is able to advance initiative with Spacebar?


Azhrei:
In tonight's game, a player was able to advance initiative by pressing the Spacebar.  It's not clear where the mouse focus was, although the player disconnecting and reconnecting fixed the problem.  I tried activating the Init Panel and hitting the Spacebar and it had no effect; other players reported the same.  So it was just this one player.  We're sure they were logged in as a Player and not GM (the GM is using a password to prevent accidentally logging in the wrong way).

Phergus:
OS?  Java version?  MapTool version?

Azhrei:
No idea; I didn't grab any of that good info as it was the middle of a session.  I've sent him a ping on our Discord server so I'll add it here when I have it.  We were using 1418 at the time, so it's the embedded Java.

leowalter:
Skimming through old issues and found this.  

My group struggled with this for a very long time.  The initiative would sometimes randomly advance without any of us knowing why.  We found it was when someone hit the space bar to active the shared pointer.  

Here is how I duplicate the problem.  When a player owns two consecutive tokens on the initiative, if they click the next button for the first token, then hit the space bar before clicking anywhere else, they will advance the initiative.  If a GM clicks the Next button and then hits the space bar before clicking anywhere else, they can advance the initiative indefinitely just using the space bar.

dorpond:
I confirmed this behavior with 1.8.4 as well. 

In fact, I just opened one of my campaigns, and as the GM, I advanced the initiative using the next button in the initiative panel, and then if I immediately hit the space bar, it will advance the initiative, as if it is actually clicking down on the next button, because every space literally clicks on the button (as it shows the animation that it is clicked). 

Another observation: if I click any of the buttons in the initiative panel, and then hit the space bar, the space does the same to them as well - the space bar becomes the left mouse button on the last button clicked.

Phergus:
This is still present in MT 1.10.  Hitting the space-bar activates the last UI element that was clicked on.  I believe this is the default behavior for JButtons.  The Initiative window just happens to be the only one were there are buttons that have actions where this matters.  So if you click on Pause Initiative, each press of space-bar will toggle the Pause status.  Same for advance/rewind initiative.

PR #3066, opened 2021-10-16
Update InitiativePanel.java

Phergus:

Corrects problem seen in #233 

### Description of the Change

Remove space bar from the input map for the Initiative panel buttons so that folks don't accidentally advance/rewind/pause initiative.

### Possible Drawbacks

Maybe someone wants space-bar to advance initiative?

### Documentation Notes

Hitting space-bar will no longer activate the last button clicked on the Initiative panel.

### Release Notes

- Hitting space-bar will no longer activate the last button clicked on the Initiative panel.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3066)


kwvanderlinde:
The change looks good to me as-is. That said, do we also want to disable other keys that behave similarly (such as Enter)? Or is only Space considered special here because it's also used for showing a pointer?

Phergus:
> 
> That said, do we also want to disable other keys that behave similarly (such as Enter)? Or is only Space considered special here because it's also used for showing a pointer?

I suspect the space-bar is the usual culprit for that very reason but perhaps blocking out the Enter key would be a good idea as well.  Are there other keys that would trip the buttons?

Phergus:
Excluding `ENTER` key as well.

PR #3125, opened 2021-10-30
2777 disable map select UI mergefixed

### LINKED ISSUE #2777: Server Option to Remove or Disable "Map Select" UI from Clients.
We can create a much better Map Select feature for ourselves in our frameworks with macros.  For example, I can make some maps visible to some players and not to others.  But even if we do, the built-in one is there and makes us have to ask players not to use it or ignore it.  In my case, in order to "disable" the built-in Map  Select I have to make all the maps "Not Visible to Players."  Which means I have to have a permanent "Map not Visible to Players" banner across the top of my maps all the time, even though the map is visible to them if they are on it.

*Describe the solution you'd like**
I would like to be able to remove or disable the built in "Map Select" from clients.  I assume this would be a server setting, but there may be other options I'm not aware of.

We can do as I'm doing and make all maps not visible to players.  But, if this is going to be the only solution, then I would like to have a way to disable the "Map not visible to Players" banner across the screen.

Map is Visible

michaelperino:
### Documentation Notes
image
image
image

### Release Notes

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3125)


cwisniew (approved):
> Reviewed 15 of 16 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3125)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @michaelperino)

thelsing:
Did you test if it works with fullscreen mode?

PR #3137, opened 2021-11-03
Startup lock token edit 2001

### LINKED ISSUE #1104: Option to not allow players to access and use their Library window to drop things onto the map.
Players can drag and drop their own tokens onto the map from their own Library window during a game.  As a GM, I would like to have an option to disallow players access to their Library window -- or at least not be able to drag and drop their own tokens onto the map.

I've had some players thoughtlessly drop random tokens onto the map, and that damages the gaming experience for everyone else.

When starting the server, one of the options presented is a checkbox to allow the GM to **disable** player ability to -- or as an option we set in Preferences:

1. access their own Library window

2. drag and drop tokens onto the map

I don't know of an alternative other than to ask players not to do it when they do -- and they so far comply (but it's still slightly awkward for everybody).  It would be better to just disable access.

**Additional context**

### LINKED ISSUE #2001: Add option to have "Lock Player Token Editor" checked as the startup default.
I always want to have "Lock Player Token Editor" (found under the "Tools" menu) checked -- and I keep forgetting to check it.

I would like to have MapTool startup with "Lock Player Token Editor" checked as the default.  I suggest adding an option to Preferences on the Application tab to enable GMs to set a default.

I keep trying to remember to set that to checked each time I start MapTool to run a game, but I keep forgetting to do it.

**Additional context**
The "Lock Player Token Editor" feature is critical because it prevents players from editing their tokens, which can create concurrency problems if the GM is trying to modify the token through his Token Editor or via macros.

michaelperino:
Resolves #2001 Resolves #1104

### Description of the Change

-Option to lock player movement and token editing on launch of server that persists through stop and start of server.
-Option to block players from dropping tokens onto map. Also blanks out library window but leaves it there in case players have a custom window arrangement they like.

### Possible Drawbacks

The start server dialog is now longer than it was last month with more options GMs must read.

### Documentation Notes

Implemented change to getInfo("server") to track blocking of player asset insertion to https://wiki.rptools.info/index.php?title=getInfo. I guess here's what a blanked out library pane looks like. 
image

### Release Notes

- Implemented options for GM to control player token insertion, movement, and token editing on server startup.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3137)


cwisniew:
@michaelperino Can we maybe add a label saying something like "Adding Tokens Disabled by GM"  to the library otherwise I foresee a lot of "All my tokens disappeared" support posts/messages :) Alternatively if they can't drag and drop do we need to clear the window?

michaelperino:
> 
> 
> @michaelperino Can we maybe add a label saying something like "Adding Tokens Disabled by GM" to the library otherwise I foresee a lot of "All my tokens disappeared" support posts/messages :) Alternatively if they can't drag and drop do we need to clear the window?

I like that, I'll take a look at that tomorrow evening probably.

FullBleed:
Was thinking the same thing... only it should probably say, "Adding _Assets_ Disabled by GM."

michaelperino:
Oh and the reason I cleared the window is so that players had a clear indication that something changed without having a banner pop up every time they try to drop something.

FullBleed:
> 
> 
> Oh and the reason I cleared the window is so that players had a clear indication that something changed without having a banner pop up every time they try to drop something.

Only reason I'd see a need to not clear the window is if there was a "Request" option where it would prompt a GM to allow a particular asset drop.

Issue #3140, opened 2021-11-04
[Bug]: Campaign Bars can't be added or edited if Language settings are set to German. 


flaedda:
### Describe the Bug

In Maptool 1.10.2 & 1.10.4 I'm not able to add or modify Campaign Bars when the Language in Maptool is set to German. Setting the Language to English solve the Problem. 

### To Reproduce

Open  Maptool 1.10.4 with Language set to german. 
Go to Campaign Settings and to the Bars Tab.
Try to add a new Bar.

### Expected Behaviour

I expect a new Campaign Bar after clicking add. 

### Screenshots

![image](https://user-images.githubusercontent.com/77189248/140339296-e0bc401d-1fe8-473b-9170-1e66a0a8c8b9.png)

### MapTool Info

1.10.4

### Desktop

Wi8ndows 10 (German Version)

Phergus:
Problem will occur in any language where the Bar Type has been translated such as Italian and Portuguese.

PR #2862, opened 2021-07-14
Give a proper name to each bar side option 

kwvanderlinde:
It's not correct top use translated text when resolving enum values using `valueOf()`. This change gives an untranslated name to each option for `tokenBarSide`, and that name is used when calling `Side.valueOf()`.

Fixes #2861

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/2862)

PR #3141, opened 2021-11-04
Update TokenBarController.java

Phergus:

Fix for #3140 

### Description of the Change

Changed code to use the i18n provided strings for bar type comparison.

### Possible Drawbacks

Unknown.

### Documentation Notes

n/a

### Release Notes

- With using languages with translated bar types users were unable to add new bars. Fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3141)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/campaignproperties/TokenBarController.java
+++ b/src/main/java/net/rptools/maptool/client/ui/campaignproperties/TokenBarController.java
@@ -63,6 +63,7 @@ import net.rptools.maptool.client.ui.token.MultipleImageBarTokenOverlay;
 import net.rptools.maptool.client.ui.token.SingleImageBarTokenOverlay;
 import net.rptools.maptool.client.ui.token.TwoImageBarTokenOverlay;
 import net.rptools.maptool.client.ui.token.TwoToneBarTokenOverlay;
+import net.rptools.maptool.language.I18N;
 import net.rptools.maptool.model.Campaign;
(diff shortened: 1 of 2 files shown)


kwvanderlinde:
I just want to reference to a similar fix that I made a few months ago in #2862. We've gone two different directions, me with giving each list item a name, and you with using the translated names. Personally, I think giving each list item a technical name is better than depending on I18N for decisions. With that said, I've also just found out that Abeille doesn't let you set the names for list items. So such a change would require manually editing the XML (like I did by mistake), and that's also not ideal.

For maintainability reasons, we should pick one way to do it and stick with it. It doesn't really matter to me whether that's by I18N or by using names, but if we go with I18N it makes sense to also update the `Side` logic to match.

Phergus:
Yeah, you're right.  Was being lazy.  Redone.

Did you even check if editing the dialog in Abeille leaves the names alone?

kwvanderlinde:
For most edits, yes it does. If you edit a list item, that particular item will lose its name.

kwvanderlinde (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3141)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cwisniew)

Phergus:
So we'll just have to be mindful if someone touches the dialog down the road.

Issue #3077, opened 2021-10-19
Add initial support for "Add-On" libraries.


cwisniew:
This is a [ZenHub](https://www.zenhub.com/) Epic

**Summary of the Epic.**
Add support for "Add-On" libraries. 
This will allow GMs to include add-ons in the campaign that can contain MT macro script, JavaScript, HTML, CSS, Images, Text files etc.

This epic is the initial steps and doesn't encompass the JavaScript API also required for a more complete add-on solution, which will be added in subsequent version (these things take time).

**Describe the solution**
See attached issues (via [ZenHub](https://www.zenhub.com/) or plug in)for further information

**Additional context**
See attached issues (via [ZenHub](https://www.zenhub.com/) or plug in)for further information

Phergus:
Initial work completed.

Issue #1348, opened 2020-03-03
Method for storing HTML / CSS/ JavaScript / Audio / etc in campaign


cwisniew:
**Create a Method for storing HTML / CSS/ JavaScript in campaign.**
The campaign file needs to be able to store these text types to make it easier to create HTML forms/character sheets etc for the new HTML5 frame. 

This should be extended to Audio and all sorts of supportable assets.

This depends on the work done for #1346

**Describe alternatives you've considered**
This could all be done in token properties or tables for the text types but its a messy messy kludge.

Phergus:
@cwisniew is this considered complete or is there more to be done?

cwisniew:
> @cwisniew is this considered complete or is there more to be done?

Probably can be considered complete. Any additional changes can get a new issue

Issue #1346, opened 2020-03-03
Expand Asset class to support things other than Images


cwisniew:
**The Asset class should be able to support multiple types of assets.**
At the moment the Asset type is being used to store other unknown binary types for the Hero Lab support, but this should be expanded so that it can store things like text (and be aware its text/sound/etc).

JamzTheMan:
FWIW this should probably include Asset Library for better integrations of file filter types. Or another issue?

It's only related in the sense that there are a lot of assumptions in the code that Assets should be an image type but currently it knows about .por and .PDF files and probably will/should know about other binaries like audio files as well.

cwisniew:
> FWIW this should probably include Asset Library for better integrations of file filter types. Or another issue?
> 
Yeah already considered it, its part of #1348 and #1349

bubblobill:
Using this would be able to have snd: tokens in the same way we have img: tokens?

Phergus:
Yes.  That's one of the goals.

Phergus:
Foundational work completed.

PR #3158, opened 2021-11-09
Addons part2

cwisniew:
Addresses #1346 
Addresses #1348 
Addresses #3077

As well as the above it also provides the initial groundwork for many other changes in the future.

### Description of the Change

Apologies for the size of the change but there was a large number of things that needed a refactor or an abstraction layer in the current code base just to get something that was functional.

This implements Add-Ons that have almost all the current features of lib:tokens (things that are missing is they don't live on the map, and they don't have macro buttons on any of the panels). 
To achieve this it also adds the following
* Asset class is now aware of what type of asset it points to and can handle many types of asset a lot better
* There is now a generic data store that is not tied to any single entity (well apart from the campaign I guess) which will allow for easy future tasks like adding data to maps without having to write new code to send it to the clients/save it to campaign. (export / import functionality will still have to be addressed but can at least be generic once written).

***This is still a work in progress to garner feedback and get something near feature complete for 1.11 out that can be tested sooner rather than later. Error handling and a few other things still need to be improved.***

Details for how add-ons work can be found here https://rptools-doc.craigs-stuff.net/blog/add-on-libraries
A small (very contrived) campaign and add-on I was using for testing can be found here https://github.com/cwisniew/test-maptool-add-on-lib

### Possible Drawbacks

* This is going to need a lot of testing... a lot of testing, especially by those that have heavy macro libraries as there was a almost complete refactor on how even token libraries get called/access from macros. 
* There will also need to be a fair amount of testing around client server connections.
* At some stage external tools (like IDE plugins) are probably going to have to be thought about to generate the json files/zip addons etc to make it easier for people.
* Except in the simplest case its not possible for for the export lib:token functionality to produce an add-on that works out of the box,  users are going to have to make some changes to what is exported. But really this is a little short term pain for long term gain for them. 

### Documentation Notes
Details for how add-ons work can be found here https://rptools-doc.craigs-stuff.net/blog/add-on-libraries
A small (very contrived) campaign and add-on I was using for testing can be found here https://github.com/cwisniew/test-maptool-add-on-lib

### Release Notes
Adds support for Add-Ons.
Adds support for expanded Asset classes.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3158)


Phergus (approved):
:lgtm:

> Reviewed 91 of 91 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3158#-:-Mo8xkZr5TKuI0APiM4K:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cwisniew)

Issue #3164, opened 2021-11-11
[Feature]:  Add a method to read static data from Add-Ons


cwisniew:
### Feature Request

Add-Ons can contain files which are useful for storing static data (json file(s) with look up tables or SRD information for example). There needs to be a way to get this file.

### The Solution you'd like

A new Macro Function "data.getStaticData(namespace, path)" which returns the information contained in the file that is at the specified path. 

If the macro function is run from the add-on it should be able to access any file, if it is not run from the token then it should only be able to access files in the public/ directory and only if the URI access flag is set.

### Alternatives that you've considered.

Not doing it and drowning under a sea of feature requests.

Phergus:
Tested.  Working as described.

PR #3165, opened 2021-11-11
add data.getStaticData macro function

cwisniew:
Addresses #3164 3164

### Description of the Change
Add the new macro function data.getStaticData(namespace, path) that will return the data at the specified path (or asset:// if t can't be converted to a macro script type). 

If the function is run from the add-on then it is able to access any file, if it is not run from the add-on then it is only able to access files in the public/ dir and only if allow URI access is enabled.

This does not work for Lib:Tokens.

Also due to the fact that MTScript doesnt have a null value if the path does not exist an empty string will be returned, but this shouldn't be a big issue as its not likely that add writers will be including empty strings as static data.

An Example has been added to  https://github.com/cwisniew/test-maptool-add-on-lib

### Possible Drawbacks
None that I can think of

### Documentation Notes
See above

### Release Notes
Added Macro Script function to read static data from add-ons

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3165)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
+++ b/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
@@ -1541,7 +1541,7 @@ public class MapToolLineParser {
    * @return the current context.
    */
   public MapToolMacroContext getContext() {
-    return contextStack.peek();
+    return contextStack.size() > 0 ? contextStack.peek() : null;
   }
 
   /**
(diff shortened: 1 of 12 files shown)


Phergus (approved):
:lgtm:

> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3165#-:-MoE6AVX7eBMT7zGc8Qa:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cwisniew)

PR #3192, opened 2021-11-19
Is number fix3191

### LINKED ISSUE #3191: [Bug]: isNumber() failing on padded strings
### Describe the Bug

Doing `isNumber(" 5")` returns false.

### To Reproduce

1. Run `isNumber(" 5")` from chat or a macro. Note the space in front of `5`.
2. Returns 0.

### Expected Behaviour

Should return `1` for true as was the behavior in MT 1.7.0 and prior.

### Screenshots

_No response_

### MapTool Info

1.9.0 to 1.11.0-alpha.5

### Desktop

Windows 10

### Additional Context

Known working in 1.7.0.  Not working in 1.9.0 to 1.11.0-alpha.5.

Phergus:

Resolves #3191 

### Description of the Change

As part of the test, whitespace padding is removed as is the first leading '+' character if present.

### Possible Drawbacks

Macros written after MT 1.7 and using `isNumber()` may produce different results now.

### Documentation Notes

`isNumber()` trims whitespace from the passed in string and strips off the first leading `'+'` character if present before evaluating whether or not the string contains a valid number.

### Release Notes

- Macro function `isNumber()` will now evaluate strings with leading/trailing whitespace or single leading `+` characters returning true for numeric strings.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3192)


Azhrei:
The docs for [Integer.parseInt()](https://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html#parseInt(java.lang.String)) indicate that a single leading minus sign or plus sign will be handled properly.  If that's true, then stripping off a plus sign would allow a number such as `++5` to be treated as positive 5 and `+-5` to be treated as negative 5.

Be sure to add tests for strings such as those.

(I'm basing this on the docs for [NumberUtils.isParseable()](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/math/NumberUtils.html#isParsable-java.lang.String-) and its references.)

Phergus:
@Azhrei passing `+5` to NumberUtils.isParseable() returns false so, with the code as submitted, passing `++5` will return false after the first `'+'` is stripped off.  Doesn't make sense to me to expect strings with an arbitrary number of `'+'` characters to be treated as numbers.  The code for `isParseable()` explicitly handles a leading minus sign but returns false with 2 or more.

kwvanderlinde (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3192)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cwisniew)

Issue #3236, opened 2021-12-02
[Bug]: HTML5 cannot pass `{body:` to `fetch`


perkinslr:
### Describe the Bug

The `fetch` API uses `Request` objects internally, which return a promise for their body argument.  This promise argument is passed directly to the underlying `xmlhttprequest.send` call, which is then serialized as `[object]` to the macro's body.

### To Reproduce

In an html5 frame, call
```js
async function test() {
  let x = await fetch("macro:myMacro@myLib", {body: "some body", method: "POST"})
  await x.text()
}

test()
```
In `myMacro@myLib`, `macro.args` is `[object]` instead of `"some body"`.

### Expected Behaviour

In the target macro, `macro.args` is `"some body"`

### MapTool Info

develop branch from github

### Desktop

Gentoo

Phergus:
Assumed to be working.

Issue #3235, opened 2021-12-02
[Bug]: XMLHttpRequest invalid UTF-16 response from lib://


perkinslr:
### Describe the Bug

The new `XMLHttpRequest` implementation improperly decodes `lib://` responses as UTF-16, which produces gibberish output.

### To Reproduce

In a frame5, execute this script, with any text data in the target URI

```js
    const x = new XMLHttpRequest()
    x.open("GET", "lib://Test/macro/test.json")
    x.onreadystatechange = () => {
        console.log(x.readyState)
        if (x.readyState == 4) {
            console.log("XMLHTTPRequest Response:" + x.response)
        }
    }
    x.send()
```

### Expected Behaviour

The content of `test.json@lib:Test` should be broadcast to the chat box.
Instead, a string of gibberish text is broadcast.
```XMLHTTPRequest Response:笢景漢㨠㌲�```

### MapTool Info

develop branch from github

### Desktop

Gentoo Linux

Phergus:
Assumed to be working.

PR #3237, opened 2021-12-02
Fix XMLHttpRequest issues

perkinslr:
#3236 
#3235 

### Description of the Change

Changes the decoding for lib:// targets to UTF-8
Changes `fetch` to an `async` function, and internally `await`s the body of the Request object.  
Filters out the pseudo-header `:Status` from the responseHeaders for the `fetch` API if the underlying response fails.

### Possible Drawbacks

Binary data returned from lib:// targets will be improperly decoded.  Binary data should be stored in asset:// instead.

### Documentation Notes
Sending `body:` argument via `fetch` API now works.
`lib://` URIs now return sensible output.

### Release Notes

Examples:

- Sending `body:` argument via `fetch` API now works.
- `lib://` URIs now return sensible output.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3237)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/htmlframe/MTXMLHttpRequest.java
+++ b/src/main/java/net/rptools/maptool/client/ui/htmlframe/MTXMLHttpRequest.java
@@ -56,7 +56,9 @@ public class MTXMLHttpRequest {
 
     responseHeaders.forEach(
         (key, value) -> {
-          jheaders.setMember(key, value);
+          if (!key.equals(":Status")) {
+            jheaders.setMember(key, value);
+          }
         });
   }
 
--- a/src/main/java/net/rptools/maptool/model/library/url/RequestHandler.java
+++ b/src/main/java/net/rptools/maptool/model/library/url/RequestHandler.java
@@ -121,7 +121,7 @@ public class RequestHandler {
       }
       responseHeaders.put(":Status", "200");
       try {
-        c.complete(new String(stream.readAllBytes(), StandardCharsets.UTF_16));
+        c.complete(new String(stream.readAllBytes(), StandardCharsets.UTF_8));
         return c;
       } catch (IOException e) {
         responseHeaders.put(":Status", "500 Internal Exception");
--- a/src/main/resources/net/rptools/maptool/client/html5/javascript/XMLHttpRequest.js
+++ b/src/main/resources/net/rptools/maptool/client/html5/javascript/XMLHttpRequest.js
@@ -107,7 +107,7 @@ class XMLHttpRequest {
 }
 
 
-function fetch(target, optionObject) {
+async function fetch(target, optionObject) {
     let request;
     if (target instanceof Request) {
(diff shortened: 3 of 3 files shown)


src/main/java/net/rptools/maptool/client/ui/htmlframe/MTXMLHttpRequest.java
```diff
@@ -56,7 +56,9 @@ public void _getResponseHeaders(JSObject jheaders) {
 
     responseHeaders.forEach(
         (key, value) -> {
-          jheaders.setMember(key, value);
+          if (!key.equals(":Status")) {
```
kwvanderlinde, line 59: Since this isn't mentioned explicitly in the bug reports, I have a couple questions:
1. What problems does it cause for this pseudo-header to be present?
2. Are there other pseudo-headers that should also be filtered out to avoid similar problems?

perkinslr: Oops!  That should have been included in the `fetch` bug report.  
`:Status` is an intentionally invalid HTTP header which lets the macro writer explicitly set the response code.  It is also used internally to get the status code from the request processing out to the wrapper layer.  It *could* use an optional, or similar data type, but exposing those to the macro writers would be a bit of a trick.  By specification, invalid HTTP headers *must not* be returned to the client, so it *must* be stripped out before the headers are returned.  If the response is wrapped in a javascript `Response` object (as the `fetch` API does), `Response`'s constructor will throw an exception on invalid headers.  Note this is different from undefined / extension headers (valid headers must not start with `:`).  

The problem is the header filtering happens when the response body is received.  If there is an exception thrown between when the header default values are set and the response body is received, the body-received function is not called, so the headers do not get filtered.  When `fetch` then tries to handle the failed request, *it* chokes on the invalid header.

So to sum up:

1. It breaks the `fetch` API
2. No.

That said, I *think* it might still have a problem *outside* the `fetch` API on internal server errors.  Probably need to refactor it slightly to always explicitly filter the headers *without* calling the body-received function.

perkinslr: Hm, nope, I can't see *why* the `MTXMLHttpRequest::recv` doesn't get called before the headers get retrieved.  Especially given that the critical `onreadystatechange` call which triggers the javascript-side header processing happens *after* the headers should be handled.

kwvanderlinde (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3237)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @perkinslr)

kwvanderlinde (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3237)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @perkinslr)

PR #3255, opened 2021-12-05
Replace hessian with protobuf

### LINKED ISSUE #3254: [Feature]: move from hessian to protobuf
### Feature Request

To address security concerns of hessian serialisation and to allow easier interoperability with other tools like apps for initiative tracking or similar stuff it would be nice to move from hessian to protobuf. 

A nice site effect is, that we would have a defined representation of the data. So we could store the protobuf objects on disk or in a DB an get rid of java serialisation as on disk format later.

### The Solution you'd like

Remove everything hessian related and replace it with protobuf.

_No response_

### Additional Context

_No response_

thelsing:
resolves #3254

### Description of the Change
This changes to format of communication between client and server to protobuf. It defines data transfer objects and mappings between them and the MT-model object. 

### Possible Drawbacks
Side effects could be many. Forgotten null values are most likely. protobuf doesn't support null for strings. So there could still be some cases that I missed. I think we should test all message on at a time.

### Release Notes
- moved serialization format from hessian to protobuf

This is still WIP. I need to go over the mappings and remove additional accessors and constructors that were needed when mapping code was in the Mapper class.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3255)


thelsing (commented):
Cleaned up changed permissions for properties/constructors. This should be ready for test now.

kwvanderlinde:
One request I have is to squash these commits prior to merging. A clean git history is quite helpful when bisecting, etc.

thelsing:
You can squash all those commits on merging in github (Button "Squash and merge") Is there any reason to do this prior?

kwvanderlinde:
It's the same thing really. I'm just leaving the note because I don't do merges but I do dig through history somewhat routinely :smile:

cwisniew (approved):
> Reviewed 120 of 127 files at r1, 1 of 1 files at r2, 6 of 6 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3255)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @thelsing)

PR #3258, opened 2021-12-06
Fixes disappearing templates on server start

### LINKED ISSUE #3256: [Bug]: Starting server will destroy certain templates
### Describe the Bug

When starting to host a server, any template of the following types will be 'destroyed' or blanked:
- radius;
- radius centered on grid;
- line;
- line centered on grid;
- wall.

These templates will still be visible on the draw explorer, but they will not be visible on the map, and their preview in the draw explorer is also not visible. Double clicking their entry in the draw explorer will center the view at (0, 0) of the map.

Templates of other types (cone, burst, blast) as well as drawings function as normal.

### To Reproduce

1. Open Maptools.
2. Draw something, like, one instance per template type.
3. Start a server.
4. Cone, burst, blast templates are still visible, while all other templates disappear.

### Expected Behaviour

Templates should not disappear.

### Screenshots

_No response_

### MapTool Info

Version: 1.11.3. I have observed similar behavior down to version 1.9.x.

### Desktop

Windows 10

### Additional Context

Starting the server also randomly brings you to a random map in the campaign. Annoying, but I don't know if that is associated with this behavior.

cwisniew:
Fixes #3256 

### Description of the Change
Templates record the id of the zone that they are in, when a server is started the zone gets a new id (in the copy constructor) so the templates no longer render properly as the zone they belong to can't be found. The least problematic solution is to update the zone id on copy. Retaining the zone id on copy has too many things it could potentially break.

### Possible Drawbacks
None foreseen

### Documentation Notes
N/A

### Release Notes
The radius, radius centered on grid, line, line centered on grid, wall templates will no longer disappear on server start.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3258)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/Zone.java
+++ b/src/main/java/net/rptools/maptool/model/Zone.java
@@ -32,6 +32,7 @@ import net.rptools.maptool.client.ui.zone.ZoneView;
 import net.rptools.maptool.language.I18N;
 import net.rptools.maptool.model.InitiativeList.TokenInitiative;
 import net.rptools.maptool.model.Token.TerrainModifierOperation;
+import net.rptools.maptool.model.drawing.AbstractTemplate;
 import net.rptools.maptool.model.drawing.Drawable;
 import net.rptools.maptool.model.drawing.DrawableColorPaint;
 import net.rptools.maptool.model.drawing.DrawablePaint;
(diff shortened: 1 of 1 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3258#-:-MqEuwyw9ZDTjvLtYx8p:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cwisniew)

PR #3260, opened 2021-12-08
Fix so Edit Token Dialog does not keep state from previous tokens

### LINKED ISSUE #3234: [Bug]: Token VBL carries from one token to another.
### Describe the Bug

If I generate VBL on a token, then drag a new token onto the map, the new token has the VBL from the previous token.  

### To Reproduce

1. Edit a token.
2. Generate some VBL
3. Exit the token editor (Ok button)
4. Drag a new token onto the map
5. Edit the new token.
6. Observe that it has the VBL already set to the VBL of the prior token.

### Expected Behaviour

I expected a new token would not have any VBL.

### Screenshots

_No response_

### MapTool Info

1.10.4, 1.11.0

### Desktop

Linux Ubuntu

### Additional Context

Issue exists on 1.10.4 and 1.11.0 so it probably has been there a while.

kwvanderlinde:

Fixes #3234

### Description of the Change

There were a couple issues to resolve:
- When a token's VBL is first generated in the dialog, a flag is set so that the VBL will be regenerated on various UI changes (e.g., changing the sensitivity, method or level). This flag was not being reset when a new token is edited, meaning the new token would automatically have VBL generated and applied to it even though the user never asked for it.
- When the Edit Token Dialog is opened, the various UI elements are initialized to match the internal state of the token VBL panel model. But since the token VBL panel wasn't reset for the new token until _after_ this initialization, the state was being set according to the previous panel state and not the new panel state.
- The **Inverse VBL Color Selection** checkbox state was not being set when the dialog is opened.

There is also a cleanup change to reset the calculated point counts to 0 even when the VBL is null, which is the case for new tokens. It doesn't look like this change is currently observable, but given that there are some conditions on the number of points being non-zero, this just reduces the possibility of future bugs due to the point count being preserved from prior uses of the dialog.

### Possible Drawbacks

It's possible that the something has a subtle dependency on the order of `bind()` vs `reset()` so the VBL could now be broken in some other way.

### Documentation Notes

N/A

### Release Notes

- Fixed Edit Token Dialog to properly update the internal and UI state of the VBL panel.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3260)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3260#-:-MqOw2w_EMBtlwImlA_I:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #3272, opened 2021-12-12
Fix for Server pushing assets when it shouldnt

### LINKED ISSUE #3271: [Bug]:  MapTool Server is currently pushing the full asset instead of allowing client to pull
### Describe the Bug

MapTool Server is currently pushing the full asset instead of allowing client to pull, this bypasses the sending of assets in chunks. Which could cause more noticeable lag on s server with slow connection and big assets.

### To Reproduce

It just happens, you don't need to reproduce it.

### Expected Behaviour

The asset should be sent as an empty array will result in  the client will fetch the asset via getAsset which will then be sent in chunks.

### Screenshots

_No response_

### MapTool Info

1.11.3

### Desktop

All

### Additional Context

The code for pushing the asset is still in MapTool but its pushed with an empty array. This is so we can in the future push smaller assets which it doesn't make much sense to ask the client to fetch if they dont have it.

cwisniew:
Fixes #3271

### Description of the Change
Server will now not push asset contents.

### Possible Drawbacks
None foreseeable, this should be reverting back to older behaviour

### Documentation Notes
N/A

### Release Notes
Fixes for server overzealous asset push

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3272)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/Asset.java
+++ b/src/main/java/net/rptools/maptool/model/Asset.java
@@ -752,7 +752,7 @@ public final class Asset {
     return new Asset(
         asset.getMd5Key(),
         asset.getName(),
-        asset.getData(),
+        null, // Always null for now, in future we may want to add the data for small assets.
         asset.getType(),
         asset.getExtension(),
         asset.getData() == null);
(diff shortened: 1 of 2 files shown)


cwisniew:
PR broken

Issue #3293, opened 2021-12-20
[Bug]: Pathfinding can send tokens through unexplored areas.


Phergus:
### Describe the Bug

The AI pathfinding algorithm can choose routes through unexplored, i.e. unexposed, areas of a map potentially allowing players to suss out details of a map they should not now.

### To Reproduce

1. Either create a map with multiple paths between two different locations or use example provided.
2. Start server and connect with player client.
3. Move player token along the longest route between the two locations exposing FoW as it goes.
4. Now drag token back to an earlier location or the original location and see the AI choose a path through the unexplored areas.

See screenshots for examples.

### Expected Behaviour

AI pathfinding should reject paths through unexposed areas (hard fog only) for players.

### Screenshots

GM view of map with player having moved token to 4th room.
![image](https://user-images.githubusercontent.com/34379254/146707196-143ee612-9bb3-4b64-b1e1-05814c79c52d.png)

Player view where player has dragged token back along the exposed area but the pathfinding chose a different (equal length) route.
![image](https://user-images.githubusercontent.com/34379254/146707368-5f0dc094-db8c-4ff7-840d-d8b8374a2362.png)

Player continues dragging token back to the second room visited and another unrevealed path is shown.
![image](https://user-images.githubusercontent.com/34379254/146707533-be4cdea6-bbd4-4c19-bb8a-fb3ca63fb6d8.png)

### MapTool Info

Version: all MT versions with AI pathfinding through 1.11.4

### Desktop

OS: tested under Windows 10 but should be same on all.

### Additional Context

Obviously this happens because all the paths are completely open with no VBL representing doors along any of them.  This is also most likely to happen when no movement limits are placed on players moving token.

Phergus:
Tested.  Working as expected now.

PR #3325, opened 2022-01-13
Restrict A* to exposed areas when FoW is enabled.

kwvanderlinde:

Addresses #3293

### Description of the Change

This change prevents pathfinding from showing paths through hard FoW, so as to avoid revealing details that would otherwise be unknown to the player. The check is entirely analogous to the existing MBL check: for each step, the line from one cell center to the next is required to lie completely within the exposed area or else it will be prohibited.

The area considered to be FoW relies on server settings as well as the effective player role:
- If FoW is disabled, FoW will be ignored for pathfinding.
- If the **Use Individual FoW** server setting is enabled, each client will pathfind according to their own token's exposed area.
- If the GM has **Show As Player** enabled, FoW will be enforced as though a player were moving the token.
- If the GM does not have **Show As Player**, then FoW will be ignored.

Since pathfinding is computed for each client separately, this means each client may potentially see different paths. However, the path seen by the client doing the move will be the authoratative path. This is the path that will be passed to `onTokenMove`, and is the path that will be returned by `getLastPath()`, regardless of which client calls the function.

### Possible Drawbacks

- The fact that different clients can see different paths may cause confusion, particularly if one path triggers certain effects while the other doesn't.

### Documentation Notes

N/A

### Release Notes

- Pathfinding no longer navigates through hard FoW

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3325)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/walker/astar/AbstractAStarWalker.java
+++ b/src/main/java/net/rptools/maptool/client/walker/astar/AbstractAStarWalker.java
@@ -27,6 +27,7 @@ import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.PriorityQueue;
 import java.util.Queue;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3325#-:-MtJjKNk8dM5g8sc2qkB:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

Issue #3101, opened 2021-10-22
Grid color resets after Server start.


Phergus:
**Describe the bug**
While you can set the grid color per map, as soon as you start a server the grid color on maps is set back to black when switching maps.

**To Reproduce**
1. Open MapTool
2. Set the grid color (Ctrl-Shift-A) to something other than black.
3. Observe that grid color has been set to desired color.
4. Start server.
5. Observe that grid color is now back to black.

**Expected behavior**
Grid color should not be reset by server starting.

**MapTool Info**
- Version: 1.11.0 (and all previous releases AFAIK).
- Install: Any

**Desktop (please complete the following information):**
 - OS: Windows but probably all OSes.
 - Version: 10

**Additional context**
Reported on Discord by Nightshade5251212:
https://discord.com/channels/296230822262865920/296657960720007169/901102432891854919

Phergus:
Tested.  Grid color is preserved over server start.

PR #3341, opened 2022-01-17
Copy grid color when copying zones

kwvanderlinde:

Addresses #3101

### Description of the Change

When a server is started, the zones are copied rather than being used as-is. The grid color was not being copied in this, unlike the other `Zone` properties, and this change fixes that.

The same thing affects copies of maps. Now copying a map will create a new map with the same grid color as the original.

### Possible Drawbacks

Shouldn't be any.

### Documentation Notes

N/A

### Release Notes

- Fix to grid color so they are preserved when copying maps and starting servers.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3341)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/Zone.java
+++ b/src/main/java/net/rptools/maptool/model/Zone.java
@@ -442,6 +442,7 @@ public class Zone extends BaseModel {
     } catch (CloneNotSupportedException cnse) {
       MapTool.showError("Trying to copy the zone's grid; no grid assigned", cnse);
     }
+    gridColor = zone.gridColor;
     unitsPerCell = zone.unitsPerCell;
     tokenVisionDistance = zone.tokenVisionDistance;
     imageScaleX = zone.imageScaleX;


FullBleed:
Az recently mentioned that when you "edit" a map it rebuilds/copies it witht he changed settings... out of curiosity, does this issue affect that?

kwvanderlinde:
This change only adds "grid color" to the list of properties that are copied over from one map (`Zone`) to another. It doesn't affect anything more fundamental than that.

Also, although I haven't chased down the necessary code paths yet, I don't think that description of editing a map is accurate. If it were, than after we use the Edit Map dialog we would find the grid color reset just like when starting a server (because zone copies are broken in this regard :smile:). Since that _doesn't_ happen, I'm inclined to think that map edits are written directly to an existing map rather than copying it.

Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3341#-:-MthYpBZCztiz563Q2lT:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

Azhrei:
> Also, although I haven't chased down the necessary code paths yet, I don't think that description of editing a map is accurate.

The Zone itself isn't deleted and recreated, but it is removed from the campaign and re-added.  The code is at **AppActions.java:3209** (comments removed):
```java
                Zone zone = MapTool.getFrame().getCurrentZoneRenderer().getZone();
                MapPropertiesDialog newMapDialog =
                    MapPropertiesDialog.createMapPropertiesDialog(MapTool.getFrame());
                newMapDialog.setZone(zone);
                newMapDialog.setVisible(true);
                MapTool.serverCommand().removeZone(zone.getId());
                MapTool.serverCommand().putZone(zone);
```

Issue #3342, opened 2022-01-19
[Performance]: Light area calculation is slow for many lights


kwvanderlinde:
### Describe the Bug

When loading a map for the first time, the visible area is calculated. This is very slow when many lights are present, with almost of the time being spent combining light areas according lumens.

There seem to be two large factors that cause this the slowdown:
1. Despite being a multithreaded operation, a lock surrounds the most expensive part. This means the operation is effectively single-threaded, plus some lock overhead.
2. The code under the lock does a lot of converting `Area`s to `Path2D` and back again. This is expensive and we can avoid a lot of this work by only converting back to `Area` once all lights have been combined.

### To Reproduce

1. Open the performance logs (Tool menu > Collect Performance Data).
2. Open the attached campaign file. It has a single map with many 144 light-bearing tokens.
3. Wait a while for the map to load. I have to wait over half a minute.
4. Observe that the first `Timer ZoneRenderer.renderZone` time since opening the campaign is high. I observed ~40s.
5. Select any token on the map.
6. Deselect the token and observe the same long delay.

### Expected Behaviour

The provided map should be able to load in under 10 seconds in the same conditions.

### MapTool Info

1.11.4

### Desktop

Linux Mint 20.2

### Additional Context

Here is the test map: [manyLightsTest.zip](https://github.com/RPTools/maptool/files/7894666/manyLightsTest.zip). Just rename it to `.cmpgn` instead of `.zip` once downloaded.

Other actions can trigger the a full recomputation of the visible area. E.g., selecting a token and then deselecting it will result in the same ~40s delay.

Phergus:
Testing.  Big performance gain.

1.11.5:
    2.    9678 ms  calcs-1
    3.    9677 ms  ZoneRenderer-getVisibleArea

Dev Code:
    2.     483 ms  calcs-1
    3.     483 ms  ZoneRenderer-getVisibleArea

PR #3343, opened 2022-01-19
Optimize combining light in `ZoneView.getVisibleArea()`.

kwvanderlinde:

Addresses #3342.

### Description of the Change

Avoid as much contention as possible when combining light areas.

- Remove the use of a granular lock when populating `allLightPathMap`.
- Avoid converting `Path2D` back to `Area` until after all lights have been combined.
- Create `allLightAreaMap` in the swing worker and return it from there. This avoids any future unintended concurrent
  interactions with `allLightAreaMap`, and makes it clear that it is only used in `ZoneView.getVisibleArea()`.

### Possible Drawbacks

Assuming I've done the concurrency correctly, shouldn't be any drawbacks. If I messed up, all sorts of weird lighting glitches could result.

### Documentation Notes

N/A

### Release Notes

- Optimized calculation of visible areas when many lights are present on a map.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3343)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/ZoneView.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/ZoneView.java
@@ -53,8 +53,6 @@ public class ZoneView implements ModelChangeListener {
   private final Map<GUID, Map<String, Set<Area>>> brightLightCache = new Hashtable<>();
   /** Map the PlayerView to its visible area. */
   private final Map<PlayerView, VisibleAreaMeta> visibleAreaMap = new HashMap<>();
-  /** Hold all of our lights combined by lumens. Used for hard FoW reveal. */
-  private final SortedMap<Double, Area> allLightAreaMap = new ConcurrentSkipListMap<>();
   /** Map each token to their personal bright light source area. */
   private final Map<GUID, Set<Area>> personalBrightLightCache = new HashMap<>();
   /** Map each token to their personal drawable lights. */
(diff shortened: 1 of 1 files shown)


kwvanderlinde:
On my system, this PR reduces contention and improves the multithreaded performance by about >6x in the case described in #3342 (> 40s to < 6.5s for map load). As far as that goes, it could be merged as-is.

But the reason I made this a draft is that I've discovered the multithreading just doesn't scale well for lights. It makes sense when the number of lights is spread across many different lumens values, but in practice this won't be the case as the number of lights grows. That's because it's trivial to attach a light source to a token, but adding new lumens requires defining a new light type. So in cases that "matter", i.e., when there are many light sources, there will be relatively few lumens to split them up by, meaning there will be a lot of contention on `allLightPathMap` entries. And after some experimentation, that contention can get really bad and grind everything to a halt.

I'll be toying with a different change that just removes multithreading in this case and makes it totally synchronous. By avoiding multithreading overhead and taking advantage of being able to mutate entries in `allLightPathMap`, this will certainly give benefits over both the current implementation and this PR for the case of many lights with the same lumens value. Depending on how well that experiment goes, I'll open a new PR and close this one off. But early numbers are looking promising...

kwvanderlinde:
I did some more testing. On my desktop with with a Intel Core i5-9400, this change gives the improvement mentioned in this PR. I've now also tried it on my laptop that has a AMD Ryzen 7 5700U. It does much better already with the existing implementation (~16 seconds) and only saves ~3 seconds. So the benefits may be pretty slim on some hardware, but it should still be a net win.

But it's all seems pretty moot anyways. I've implemented the single-threaded alternative and I'm finding the same case only takes ~1 second on my desktop (~40x improvement) and just shy of 1.5 seconds on my laptop (~12x improvement). So I'm closing this PR in favour of the alternative which is faster, needs less code, and is better for maintenance we don't have to reason about concurrency.

PR #3344, opened 2022-01-19
Optimize combining light in `ZoneView.getVisibleArea()`.

kwvanderlinde:

Addresses #3342.

### Description of the Change

The existing multithreaded approach for combining light areas is effectively synchronous as the most expensive part is guarded by a lock. This PR embraces that fact and removes the multithreading altogether in favour of a simpler implementation that can take advantage of mutating state in a way that the multithreaded code cannot do safely.

Additionally, we avoid unnecessarily conversions between `Path2D` and `Area` by working solely with `Path2D` and converting to `Area` only once all light areas have been combined.

### Possible Drawbacks

Shouldn't be any.

### Release Notes

- Optimized calculation of visible areas when many lights are present on a map.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3344)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/ZoneView.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/ZoneView.java
@@ -22,8 +22,6 @@ import java.awt.geom.Rectangle2D;
 import java.util.*;
 import java.util.List;
 import java.util.Map.Entry;
-import java.util.concurrent.*;
-import javax.swing.*;
 import net.rptools.maptool.client.AppUtil;
 import net.rptools.maptool.client.MapTool;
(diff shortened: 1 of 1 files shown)


kwvanderlinde:
Just some numbers to summarize the improvement. The timings I'm referencing are the `ZoneRenderer-getVisibleArea` timer when loading the map provided in #3342.

On my desktop with a Core i5-9400, I was finding this number was consistently just over 40s in 1.11.4. After this change, that value went down to ~1 second.

On my laptop that has a Ryzen 7 5700U, I was seeing ~16s on 1.11.4 and ~1.3 seconds after the change.

Issue #3346, opened 2022-01-19
[Feature]: Allow darkness to affect Day vision


kwvanderlinde:
### Feature Request

I'd like to be able to use areas of darkness (light with negative lumens) on maps even which could be used with either Day or Night vision. Currently this is not possible since Day vision does not account for light sources and thus cannot remove areas of darkness from a token's vision.

### The Solution you'd like

I would like to have areas of darkness with a lumens value of -100 or less to remain hidden from players in the Day, unless they have sight with a suitably high lumens value. This should make the default area of vision in Day behave like sight with the default lumens value.

### Alternatives that you've considered.

Using VBL to occlude a given area could be done. But this is all-or-nothing and doesn't allow tokens with special vision to see into the area the way we can do with negative lumens.

### Additional Context

This is what is looks like for a token with darkvision at Night with a -100 lumen source of darkness nearby: 
![image](https://user-images.githubusercontent.com/7492219/150231980-4053f74c-7d78-412e-958b-25555c7762fe.png)

At day the darkness hides nothing, even though it may not be intended for the players to see into it: 
![image](https://user-images.githubusercontent.com/7492219/150232049-e572e2b3-7f78-4a6b-93a7-5d3810613ed4.png)

The request is to make Day obscure the area of darkness like this: 
![image](https://user-images.githubusercontent.com/7492219/150232334-fd5ad692-2656-44ac-b503-8d8c6df28ef0.png)

adyoungblood:
Actually, I think making these changes would be immensely beneficial towards resolving #1908, as currently the Day and Night modes interact in a hard-coded way with regards to light and vision. The changes I'm envisioning would have fog be the default over the entire map, with Day mode adding a light (at 99 lumens, perhaps) over the entire map. Lights and sights would overwrite the fog aesthetically.

FullBleed:
This all sounds interesting.  I can see the intersection of your ideas working out quite well. 

> Actually, I think making these changes would be immensely beneficial towards resolving #1908, as currently the Day and Night modes interact in a hard-coded way with regards to light and vision. The changes I'm envisioning would have fog be the default over the entire map, with Day mode adding a light (at 99 lumens, perhaps) over the entire map. Lights and sights would overwrite the fog aesthetically.

Out of curiosity, why 99 lumens?  Why not 100?  If Darkness is -100 it seems that "day light" used this way should be 100.

adyoungblood:
> Out of curiosity, why 99 lumens?
[...]

PR #3349, opened 2022-01-22
Calculate lights when map vision is set to Day or Off

kwvanderlinde:

Implemented #3346

### Description of the Change

Lights are now calculated and rendered when the map vision setting is *Day* or *Off* (auras were already rendered in all modes and still are). Additionaly, in *Day* and *Off* modes, the visible area will be treated as a source of light with a lumens value of 1. This means any darkness (light with negative lumens) will restrict vision just as in *Night* mode, while still permitting PCs to counter the darkness if they have a light or sight with a sufficient lumens value.

A refactoring has also been included. It's important to sort the lumens values so that darkness of a certain level can be removed after light of that level or less has been calculated. This used to be done by carefully choosing lumens values to insert into the light area map (integral values for light, fractional values for darkness). Now this is done by adding a comparator so that we can be more free in adding lights to the map without worrying about breaking the sorting order. This has also had a tangible, though slim, performance improvement for a moderate number of lights, and has comparable performance when the number of lights is very large.

### Possible Drawbacks

*Day* and *Off* modes used to be feasible as a sort of escape hatch for performance with regards to rendering light and darkness. With this change, that escape hatch will no longer be available. The only real alternative is to reduce the number of lights on maps that are struggling.

### Documentation Notes

What follows is an adaptation from the Wiki ([Introduction to Lights and Sights](https://wiki.rptools.info/index.php/Introduction_to_Lights_and_Sights#Vision_Modes)), updated to explain the behaviour introduced in this change. If this change goes in, I intend to update the wiki to something like what I've put below:

> Lights (including darkness) and sights will always be calculated regardless of the vision type. The Map Vision only controls what areas of the map are considered visible, and therefore how Fog of War and VBL hide things from the players.
>
> Areas of darkness, unless revealed somehow, will natural keep Fog of War in the area and thus obscure the area from players. If Fog of War is removed in the area (either manually, by dispelling the darkness or by overcoming the darkness with light) that area of the map will become visible like any other revealed area. If the darkness then returns, the map itself will remain visible but things inside that area may become obscured again, much like what happens with VBL.

**Off**

When Map Vision is set to "Off", vision blocking is not taken into account when displaying information to players. All things are visible at all times, unless hidden beneath Fog of War.

**Day**

In Day mode, no part of the map is hidden from players (that is, they will see the entire map - not necessarily all the tokens on a map, but they will see the layout of the entire dungeon, building, or area). Effectively, the Day mode assumes that a bright sun is shining down on everything, illuminating it all and making everything visible unless blocked by VBL, encased in darkness, or covered by Fog of War. This "sun" acts like a light source with minimal lumens, meaning any area of darkness will block the light in that area.

**Night**

This mode assumes that, without a light source or special sight, the tokens are in a pitch-black environment and can see nothing. If Fog of War is used with Night Mode, light sources will reveal areas covered by Fog of War, assuming a token can see the light source and the area is not blocked by VBL.

### Release Notes

- Altered _Day_ and _Off_ vision modes to render and account for lights and darkness as is done in _Night_ mode.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3349)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3349#-:-Mu8BnXdBsSXTKxu3F8J:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #3361, opened 2022-02-04
Upgrade handshake cipher to AES CBC

### LINKED ISSUE #3360: [Feature]: Upgrade handshake cipher to AES CBC
### Feature Request

The handshake process requires decrypting, stripping a prefix from and reversing the challenge, then re-encrypting it.
The cipher used is "AES" which means "AES/ECB/PKCS5Padding".
"ECB" ciphers are disrecommended in-general because they are vulnerable to replay attacks, chosen plaintext attacks and are easier to build a code-book for.
This means such ciphers aren't always available in cipher suites, which makes interoperability more difficult, and while I know of no vulnerability upgrading may be a useful precaution.

### The Solution you'd like

Change the cipher to "AES/CBC/PKCS5Padding", which involves changing the handshake to include exchanging IV parameters.

De-scope password support from my hobby client, or find another library that will accept ECB upstream.

### Additional Context

_No response_

fishface60:

resolves #3360 

### Description of the Change

Changes the CryptoUtil cipher to "AES/CBC/PKCS5Padding", adds an IV field to handshake messages, and generates a new IV per message.

In the process, found a couple of bugs, and removed CipherUtil objects, which were only used to immediately retrieve the key out of.

### Possible Drawbacks

Slightly more CPU cost involved in generating and transmitting the IV, and chaining blocks together.

It's more changes to cryptography that could result in a vulnerability.

### Release Notes

- Fixed a bug which could rarely cause a false handshake failure
- Fixed error message reporting public key authentication failure while using password authentication.
- Upgraded handshake cipher to CBC for better interoperability and precaution.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3361)

### DIFF
--- a/src/main/proto/handshake.proto
+++ b/src/main/proto/handshake.proto
@@ -40,11 +40,13 @@ message ClientInitMsg {
 message UseAuthTypeMsg {
   AuthTypeEnum auth_type = 1;
   bytes salt = 2;
+  bytes iv = 4;
   repeated bytes challenge = 3;
 }
 
 message ClientAuthMsg {
   bytes challenge_response = 1;
+  bytes iv = 2;
 }
 
 message ConnectionSuccessfulMsg {
--- a/src/main/java/net/rptools/maptool/client/AppActions.java
+++ b/src/main/java/net/rptools/maptool/client/AppActions.java
@@ -31,6 +31,7 @@ import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.UnknownHostException;
+import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
 import java.security.spec.InvalidKeySpecException;
@@ -2342,6 +2343,7 @@ public class AppActions {
                   MapTool.showError("msg.error.failedConnect", ioe);
                   failed = true;
                 } catch (NoSuchAlgorithmException
+                    | InvalidAlgorithmParameterException
                     | InvalidKeySpecException
                     | NoSuchPaddingException
                     | InvalidKeyException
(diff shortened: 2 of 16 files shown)


src/main/proto/handshake.proto
```diff
@@ -40,11 +40,13 @@ message ClientInitMsg {
 message UseAuthTypeMsg {
   AuthTypeEnum auth_type = 1;
   bytes salt = 2;
+  bytes iv = 4;
```
thelsing, line 43: any reason what 4 is before 3?

thelsing (approved):
I'm no crypto expert, but the changes look fine to me. I also think it's better to change from ECB to CBC.

src/main/proto/handshake.proto
fishface60: Just that conceptually the salt and IV belong together as encryption parameters and it makes sense for the challenge to come after, but my understanding of the reason why protobufs lets you specify the field order is that they're theoretically backwards compatible.
MapTool doesn't need to support that so I can't think of any problems with changing the order.

thelsing:
Successfully tested.

Phergus (approved):
:lgtm:

> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3361#-:-N-97VddA8UhZUeh0KPy:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @thelsing)

Phergus:
@thelsing  anytime you comment on code you must resolve the discussion for the PR to be accepted.

thelsing (commented):
> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3361)* status: all files reviewed, 1 unresolved discussion (waiting on @thelsing)

PR #3339, opened 2022-01-17
Overhaul how colored lights blend with other lights

adyoungblood:
### Identify the Bug or Feature request

fixes #1550 
fixes #2094 
resolves #1895 

### Description of the Change

This change touches on several issues that ended up all being related to one another. The largest change is to make the blending of colored lights additive rather than subtractive with respect to RGB values. Additionally, the default light is now "clear", i.e. it does not override the colors of other light sources. All FOW vision calculations remain the same. These changes do not extend to auras, though it would be simple to do so. 

### Possible Drawbacks

- Odd edge cases that look unintuitive or odd to users
- Performance issues with many lights, as I removed an optimization that was causing lights with the same color to not blend (though I also took care to make the color composition as performant as possible)
- [Odd workflows](https://xkcd.com/1172/) breaking

### Documentation Notes

**Colored Light Overlap**
Original effect (from #1550):
image
New effect:
image

**Default Light Interaction with Darkness**
Old effect (from #2094):
image
New effect:
image

**RGB Color Blending**
Old effect (from #1895):
image
New effect:
image

### Release Notes

- Made colored lights prettier when combined

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3339)

Issue #1550, opened 2020-04-14
Strange lighting when coloured light sources overlap


kwvanderlinde:
**Describe the bug**
Overlapping light source colours are blended only if they are different, whereas when they are the same colour the result is always that same colour. This creates an undesirable "Venn Diagram" effect when lights with colours for multiple ranges overlap.

**To Reproduce**
1. Modify the `Torch - 20` light source to use colours (under `Edit > Campaign Properties...`, `Light` tab):
   ```
   Torch - 20: circle 20#aaaaaa 40#666666
   ```
2. Add two PC tokens to the map, and attach a `Torch - 20` light source to each.
3. Disable Fog of War and set the map background to pure black. (Not necessary, but avoids distractions).
4. Place the tokens 80' apart, then 60', 50' and finally to 35' apart, observing the overlapping light behaviour. Notice how the brighter colour that is present only when the 20' colour and the 40' colour overlap, but no other blending occurs.

**Expected behavior**

Any one of these or some other reasonable behaviour:
1. Light source colours are always blended.
2. The colour for the smallest range is rendered, with no blending.

**Screenshots**

I've set up a map with a black background, and tokens placed at each of the ranges listed under "To Reproduce" (80' on the left, 35' on the right). Since the blending of the 20' colour and 40' colour is brighter than the constituent colours, we get this "Venn Diagram" effect when the tokens are less than 40' apart from each other.
![image](https://user-images.githubusercontent.com/7492219/79273982-ae9d7d00-7e58-11ea-9d7f-210082dc0358.png)

**MapTool Info**
- Version: 1.6.1
- Install: New

**Desktop (please complete the following information):**
 - OS: Linux Mint
 - Version 19.4

**Additional context**

```
==== MapTool Information ====
MapTool Version: 1.6.1
MapTool Home...: /home/kenneth/.maptool-rptools
Max mem avail..: 3 GB
Max mem used...: 250 MB
Custom Property: -DMAPTOOL_DATADIR=.maptool-rptools

==== Java Information ====
Java Vendor.: Oracle Corporation
Java Home...: /opt/MapTool/runtime
Java Version: 10.0.2

==== OS Information ====
OS Name........: Linux
OS Version.....: 5.3.0-46-generic
OS Architecture: amd64
[...]

PR #3423, opened 2022-05-25
Blending composite screen blend

kwvanderlinde:

Improves on PR #3339.
Relates to #1550 particularly for auras. 

### Description of the Change

There are two improvements in this PR.

#### Screen blend
The first change replaces the additive blending in `BlendingContext` with a ["screen" blend](https://en.wikipedia.org/wiki/Blend_modes#Screen). The screen blend is much like the additive blending: a black input doesn't affect the result; a white input forces a white result; and the result is otherwise always brighter than the inputs. Unlike additive blending, the result is only maxed out at white if one of the inputs is white, which allows for more subtlety when blending lights that aren't completely saturated. For a visual reference, here is what additive blending looks like for inputs ranging from [0, 0] (top left) to [255, 255] (bottom right):
add-blend-mode
Note that the bottom right half of the image is maxed out at white. And this is what screen blending looks like:
screen-blend-mode

#### Auras blend as lights
The second change unifies light rendering with aura rendering. Auras are now also rendered to an image and are blended the same way that lights are. I.e., auras of the same color are no longer merged, and overlapped auras get brigher rather than darker. The auras are then rendered as a transparent overlay on top of the map. Auras have their own overlay separate from the light overlay, and this overlay is drawn on top of the light/sights overlay.

As part of the change, both lights and auras are no longer grouped by common paint but are instead rendered in whatever order they are returned. The rendering order doesn't make a difference for either screen or additive blending, so it  so the fact that a potentially different rendering order is used does not matter. If we ever change to something where order matters, the lights list / auras list can be sorted in a meaningful way. 

### Possible Drawbacks

- Performance is slightly worse since `BlendingComposite` is less performant than `AlphaComposite`.
- Auras and lights not blending with each other could be confusing. Should they use the same overlay?

### Documentation Notes

N/A

### Release Notes

- Auras now add together like lights do.

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/BlendingComposite.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/BlendingComposite.java
@@ -23,6 +23,7 @@ import java.awt.image.DirectColorModel;
 import java.awt.image.Raster;
 import java.awt.image.RasterFormatException;
 import java.awt.image.WritableRaster;
+import org.checkerframework.common.value.qual.IntRange;
 
 /**
  * A custom Composite class to replace AlphaComposite for the purposes of mixing lights, auras, and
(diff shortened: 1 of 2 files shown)


adyoungblood:
I'm glad you could get the performance to a reasonable state. I investigated AlphaComposite to see if it would be possible to take some design hints from it, but best I can tell it's outdated Sun code, and there be dragons.

As for overlays and layering, it might be interesting/useful to refactor the rendering code into many methods that create and return overlays (for lights, auras, etc.), then combine those overlays with varying Composites in a larger method. That would make it easier to change how each layer interacts with the others. For instance, one could change the Composite that blends the lights/auras with the map/background image in a single line.

cwisniew (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3423)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

thelsing:
Looks ok to me.

Issue #3425, opened 2022-05-27
[Bug]: Maptool hangs at splash-screen after macro import with unkown asset-link


Xun71:
### Describe the Bug

Importing a macro featuring an ICON (asset-link) into maptool's Global macro window leads to the effect that maptool does not start anymore after closing. 
Starting MT again keeps it frozen at the splash-screen, last entry in log is: 'INFO  net.rptools.maptool.client.AppUpdate - Cleaned version: 1.11.5'. Memory-usage displayed in TM: ~232MB, no CPU activity.

Only de-install, clearing of all reg-entries referring 'maptool', and re-start of the computer allows a functionig new-install.
The issue can be narrowed down to the case, where the asset-link is not known ('red X as defaouzlt ICON)

### To Reproduce

1. Start fresh installation of MT.
2. import macro with an (unknown) asset referred in the details tab (e.g. 
![grafik](https://user-images.githubusercontent.com/106387368/170706002-4dd082e2-5a0a-48ce-bfad-e1af08be2688.png)
3. apply the change to the macro
4. close MT
5. start MT
6. *hangs at splash-screen'

### Expected Behaviour

I would expect MT to ignore the unknown asset-link and keep the default 'red X' Icon

### MapTool Info

version 1.11.5

### Desktop

Windows 11, 64bit

kwvanderlinde:
A quick test shows that this also impacts Linux, and a check of the code suggests all platforms will be impacted similarly. As far as I can tell from drilling into the code, this is what causes the issue:

At a high level, the problem stems from a weakness in MapTool's initialization code: there is no guarantee that all the "plumbing" exists (e.g., the server command, or the connection beneath that) prior to all the UI-centric code that might rely on it. In this case, the UI is trying to load assets and expects to be able to transfer them from the server, but since the server plumbing isn't set up yet that just fails.

At a very low level, this particular issue is caused by a weakness in realized in `AssetLoader.ImageRetrievalRequest::run()`, which runs deep in the call stack and through a tangle of asynchrony. The fallback logic is particularly at fault:
```
MapTool.serverCommand().getAsset(id);
```
Right now the problems with it are:
1. `ImageRetrievalRequest` relies on this fallback logic to actually load the asset and therefore notify listeners.
2. There is no `ServerCommand` created yet, so this line in fact throws a NPE and does not notify listeners.
3. Even if we have a `ServerCommand` (e.g., by creating it only _one line_ earlier), it won't have a connection and so doesn't do anything.
[...]

PR #3428, opened 2022-05-30
Do not assume a server connection exists when retrieving images

kwvanderlinde:
Address #3425 

### Description of the Change

It can happen that - early in the initialization process - assets are requested to populate the UI
even before any personal server has been started or `ServerCommand` created. In this case, we cannot rely on a server
call to resolve and image and call pending listeners. To work around this problem, we fallback to using a broken image
if the server is not available.

### Possible Drawbacks

Yet one more bit of magic has been added to the codebase.

### Documentation Notes

N/A

### Release Notes

Fixed a problem where missing assets in Global macro labels would cause MT to hang on start.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3428)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/AssetLoader.java
+++ b/src/main/java/net/rptools/maptool/model/AssetLoader.java
@@ -324,11 +324,18 @@ public class AssetLoader {
         }
       }
 
-      // Last resort, ask the MT server
-      // We can drop off the end of this runnable because it'll background load the
-      // image from the server
       // System.out.println("Got " + id + " from MT");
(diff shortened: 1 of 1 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3428#-:-N3jg9SX4f91honxupE1:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

Issue #3052, opened 2021-10-14
Allow tokens to carry Hill and Pit VBL, just as they can carry Wall VBL.


kwvanderlinde:
**Is your feature request related to a problem? Please describe.**
Tokens are able to carry Wall VBL, but Wall VBL hides what is underneath it. Hill VBL and Pit VBL on tokens would allow (parts of) a token to not be hidden by the vision blocking.

**Describe the solution you'd like**
I would like tokens to be able to store Hill and Pit VBL, in the same way they store Wall VBL. The "Edit Token" dialog would be extended to allow manipulation of Hill and Pit VBL, either automatically or by transferring to/from the map.

The following changes would be made to the **Edit Token** dialog:
1. The **Generate Token VBL** button is capable of generating Wall, Hill or Pit VBL as desired.
2. The **Move VBL To Map** button is capable of copying any of the token's Wall, Hill or Pit VBL to the map as desired, respecting the **Erase Source VBL on Move** option.
3. The **Move VBL From Map** button is capable of copying any of the map's Wall, Hill or Pit VBL to the token as desired, respecting the **Erase Source VBL on Move** option.
4. Any use of the word "VBL" can either apply to all types of VBL or is qualified as either "Wall VBL", "Hill VBL" or "Pit VBL".

The following macro functions would also be added so that macros can manipulate token Hill and Pit VBL analogously to how they can manipulate token Wall VBL:
- `getTokenHillVBL(id)`
- `setTokenHillVBL(vbl, id)`
- `transferHillVBL(value), transferHillVBL(value, delete), transferHillVBL(value, id), transferHillVBL(value, delete, id)`
- `getTokenPitVBL(id)`
- `setTokenPitVBL(vbl, id)`
- `transferPitVBL(value), transferPitVBL(value, delete), transferPitVBL(value, id), transferPitVBL(value, delete, id)`

**Describe alternatives you've considered**
It is already possible to add Wall VBL to a token and then mark it as **Is Visible over FoW**. The effect is quite good:
![image](https://user-images.githubusercontent.com/7492219/137372040-f142452d-2b2b-4be4-b688-f093458cd5dd.png)
But this doesn't work for some other situations. For example, sometimes it is handy for the map to actually be a background stamp carrying VBL. In that case, we can't selectively have certain parts of the map be visible over FoW. Also, **Is Visible over Fow** only approximates the effect of Hill-with-Pit VBL, and can't approximate only Hill VBL or only Pit VBL.

**Additional context**
[...]

Issue #2730, opened 2021-06-11
Allow tokens to carry MBL, like they can with VBL.


kwvanderlinde:
**Is your feature request related to a problem? Please describe.**
Tokens are able to block vision, and they are also able to block pathfinding, but only if the AI is set to pathfind around VBL. This means a token cannot block movement without requiring everything else to pathfind around VBL as well.

**Describe the solution you'd like**
I would like tokens to be able to store MBL, in the same way they store VBL. The "Edit Token" dialog would be extended to allow adding MBL, either separately from VBL or in concert (much like MBL/VBL can be drawn on the map separately or together). 

The following macro functions would also be added so that macros can manipulate token MBL analogously to how they can manipulate token VBL:
- `getTokenMBL(id)`
- `setTokenMBL(vbl, id)`
- `transferMBL(value)`, `transferMBL(value, delete)`, `transferMBL(value, id)`, `transferMBL(value, delete, id)`

**Describe alternatives you've considered**
Tokens can already block movement by setting their terrain modifier to "Block". This is quite a blunt option, though, as the token then blocks entire cells. This makes it unsuitable for representing a glass door on a grid line, for example. It also doesn't allow freely placing irregularly shaped terrain, since the token always blocks its entire footprint. Also, Free Size / Native Size tokens cannot block movement this way.

Regarding the macro functions, another option would be to define generic topology functions that can work equally well with VBL and MBL. Given that we already have VBL-specific functions and some MBL-specific functions, it makes the most sense to at least expand the MBL functions to support token MBL. Generic topology functions would then be redundant, and could even be built as UDFs if desired.

**Additional context**
This is somewhat of a follow up to #2234 that added functions for drawing MBL on the map. This FR seeks to expand on that and bring parity between MBL and VBL.

Not all VBL functionality requires MBL equivalents. For example, the "Is Visible over FoW" option and the related `setAlwaysVisible()` macro function are more about a token's relationship with Fog of War than they are about the VBL _per se_, and there isn't an analogous use for MBL.

Phergus:
Tested various iterations of `get`, `set` and `transferMBL`.  Worked as expected.

PR #3430, opened 2022-06-03
Add Hill VBL, Pit VBL and MBL to tokens

kwvanderlinde:

Addresses most of #3052 and #2730, leaving out only the **Edit Token** dialog changes.

#### New features
- Macro functions for manipulating token topology
  - `getTokenHillVBL`
  - `getTokenPitVBL`
  - `getTokenMBL`
  - `setTokenHillVBL`
  - `setTokenPitVBL`
  - `setTokenMBL`
  - `transferHillVBL`
  - `transferPitVBL`
  - `transferMBL`
- Token topology is rendered in distinct colours for each topology type.
- Color sensitivity for auto-generated topology is remembered per topology type.

#### Internal changes / refactorings
- Token model now has fields for Hill VBL, Pit VBL and MBL in addition to the existing Wall VBL.
  - Token model update command `setVBL` has been replaced by `setTopology` which requires a topology type to be sent as well as the area.
- Protobuf Token DTO has been extended to support the new fields.
- The macro function code under `Topology_Functions` was refactored so that each kind of function is handled by a separate method rather than keeping all that code in a single method under an if/else chain.
- `Zone` and `ZoneView` only expose topology via `TopologyType` rather than having distinct methods for each kind of topology.
- Topology caching in `ZoneView` has been revamped.
  - `ZoneView` no longer caches the token topology `Area`s and instead just caches the total topology of the map plus tokens.
  - `ZoneView` now caches the topology `Area`s for each topology type instead of just the `AreaTree`s. This benefits the A* walker because it needs the `Area`s but previously had to rely on the expensive `AreaTree`s.
  - The cached `AreaTree`s are now only needed internally to `ZoneView`, for passing to the FoW calculations.
- Most methods with the term "VBL" in the name now use "topology" to make clear that they are more general.

#### Bug fixes
- Subtle issue in A* algorithm to only subtract token topology from total topology of the same type. Previously a
  token's Wall VBL would be subtracted from all topologies.

### Possible Drawbacks

Performance may be worse in some circumstances as there are now four loops over tokens to collect the various topologies instead of one (e.g., for vision calculations and A*).

### Documentation Notes

Tokens can now carry all kinds of topology (Wall VBL, Hill VBL, Pit VBL and MBL). The topology for a token can be manipulated using these macro functions:
- `getTokenVBL(id)`/`getTokenHillVBL(id)`/`getTokenPitVBL(id)`/`getTokenMBL(id)` are used to get the Wall VBL/Hill VBL/Pit VBL/MBL of a token.
- `setTokenVBL(json, id)`/`setTokenHillVBL(json, id)`/`setTokenPitVBL(json, id)`/`setTokenMBL(json, id)` are used to set the Wall VBL/Hill VBL/Pit VBL/MBL of a token.
- `transferVBL(direction, delete, id)`/`transferHillVBL(direction, delete, id)`/`transferPitVBL(direction, delete, id)`/`transferMBL(direction, delete, id)` are used to copy or move topology from a map to a token or vice-versa.

When a token has a given type of topology, the effect is as if the map has that topology at the token's location. The difference is that a token's topology moves with the token, unlike the map topology. An additional difference is that a token's topology is ignored when that token is pathfinding, so that a token cannot block its own movement. The same does not hold true for vision: for example, a token carrying Pit VBL cannot see outside of its own topology.

### Release Notes

- Added Hill VBL, Pit VBL and MBL for tokens, and macro functions for interacting with them.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3430)


kwvanderlinde:
@thelsing I'd especially love to get your feedback on the protobuf changes if possible as I've hardly worked with protobuf before. So I just made changes that _work_ (based on testing), but I have no idea if they are _good_.

Phergus (approved):
:lgtm:

> Reviewed 17 of 17 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3430#-:-N3jlSplEYlpol5vQuFu:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @thelsing)

Issue #3436, opened 2022-06-07
[Bug]: Unable to join servers with certain tokens present.


kwvanderlinde:
### Describe the Bug

After starting a server with certain tokens present, players are unable to join the server and are met with a black screen saying "Loading Campaign".

### To Reproduce

1. Start a first MapTool instance
2. Import this UVTT file: [10x10 Four Rooms with Lights.dd2vtt.zip](https://github.com/RPTools/maptool/files/8855625/10x10.Four.Rooms.with.Lights.dd2vtt.zip)
3. Start a server
4. Start a second MapTool instance
5. Attempt to join the server from the second instance
6. At this point the second instance should be stuck at "Loading Campaign"
 

### Expected Behaviour

Players can join the server regardless of the tokens present.

### MapTool Info

develop (85417f5)

### Desktop

Linux Mint 20.3

### Additional Context

The bug results from an NPE being thrown on server start thanks to the token's shape/type being null while protobuf expects non-null values.

thelsing:
Is the bug token having no shape or protobuf expecting a shape?...

Phergus:
When the light tokens are created/added in the DungeondraftImporter no shape is being set so that's probably one issue.

I also notice that if I just immediately import that particular UVTT map and start a server when the server started message is shown the map display goes black.  I can then select the map from the map selector and get the map display but get an NPE:
![image](https://user-images.githubusercontent.com/34379254/172463297-476a96a0-8448-47a0-b409-e77e81b59ccf.png)

So possibly something else isn't initialized and needs to be.  Don't see this issue with maps that don't have the light tokens created.

thelsing:
I think this is because the server wasn't started properly. (black screen) You probably shouldn't be able to change the map in this case.

Phergus:
PR #3437 appears to have corrected the issue in my testing.

PR #3437, opened 2022-06-08
Explicitly initialize some more Token fields to avoid nulls

kwvanderlinde:
Addresses #3436 

### Description of the Change

The `Token` class has many different means of initialization (default constructors and similar, construction from transfer data, initialization in `fromDto()`, XStream/`readResolve()`, etc). This means it is very easy for each of these to give subtle differences in the exact state of a token. E.g., a default constructed `Token` will have its `tokenShape`, `tokenType` and `layer` all set to `null`, but a deserialized version of the same token will have those fields initialized to a real value. The `DungeonDraftImporter` was using one of `Token`s simple constructors which - similar to the default constructor - left these fields null, whereas `Token.toDto()` expects these fields to be non-`null`.

This PR gives initializers to each of these fields so that they can reliably be non-`null` when creating a `Token` via any of its constructors. `readResolve()` and the accessor methods also set these particular values if the values are missing, so we can now rely on the fields being non-`null`. Since these fields are now guaranteed to be non-`null`, the accessors have been simplified to no longer check that case.

In a similar vein, we can infer that the fields `name` and `exposedAreaGUID` should not be `null` either, so those are also now initialized to non-`null` and `TokenDto` has been updated accordingly. As far as I can tell, all other fields without initializers are either primitives or can reasonably be `null`.

### Possible Drawbacks
Should be none if I did my job right :smile:

### Documentation Notes
N/A

### Release Notes

N/A (#3436 only affects development versions)

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3437)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/Token.java
+++ b/src/main/java/net/rptools/maptool/model/Token.java
@@ -92,7 +92,7 @@ public class Token extends BaseModel implements Cloneable {
   public static final String LIB_TOKEN_PREFIX = "lib:";
 
   private boolean beingImpersonated = false;
-  private GUID exposedAreaGUID;
+  private GUID exposedAreaGUID = new GUID();
 
   /** the only way to make Gson apply strict evaluation to JsonObjects, apparently. see #2396 */
   private static final TypeAdapter<JsonObject> strictGsonObjectAdapter =
@@ -267,7 +267,7 @@ public class Token extends BaseModel implements Cloneable {
   private Area pitVbl;
   private Area mbl;
 
-  private String name;
+  private String name = "";
   private Set<String> ownerList = new HashSet<>();
 
   private int ownerType;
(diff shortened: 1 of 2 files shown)


thelsing (approved):
Everything looks fine to me.

thelsing (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3437#-:-N40m6pA2kjCuBLbXF_A:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Phergus)

Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3437#-:-N480ebm0qH0wRy1yg6d:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #3449, opened 2022-06-30
Fix for Using addons wont allow players to connect #3359

### LINKED ISSUE #3359: [Bug]: Using addons prevents players from connecting to server
### Describe the Bug

Using an addon in a campaign file prevents players from joining if that addon is not already "loaded in" to their current campaign file. They receive an "Incorrect Password during handshake" error when attempting to join. Adding the same addon to an empty grasslands campaign allows the player to connect.

### To Reproduce

1. Create an addon and add it to the campaign
2. Start a server
3. Have someone try to join the server - they will receive an error.
4. Send the .mtlib file to the player you want to join and have them add it to an empty campaign
5. Have that player try joining again - they will be able to join.

### Expected Behaviour

The connecting player should not need to pre-install any addons to join.

### Screenshots

_No response_

### MapTool Info

1.11.4

### Desktop

Windows 10

### Additional Context

_No response_

cwisniew:
resolves #3359

### Description of the Change

Fixes the problems where the add on wasn't always fetched from the server correctly, and the issue that was causing the asset to be created empty.

### Possible Drawbacks

None foreseeable.

### Documentation Notes
None required as this should allow addons to be used as already documented.

### Release Notes
- Fixed an issue where using add-ons would cause problems for players trying to connect if they did not already have the add on.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3449)


kwvanderlinde:
A consequence of recognizing `application/zip` as `Asset.Type.MTLIB` is that I can now drag and drop other zip files into a campaign where previously I would be prevented from doing so. E.g., before this change, if I tried to drag a `.mttable` file onto a map, I would get this dialog:
image

But after this change I would just get a blank token with a broken image:
image

If this is deemed acceptable, then I am okay with this change. Otherwise a better approach may be to force the type of the asset that is passed to `AddOnLibraryImporter.importFromAsset()` rather than having the MTLIB type be auto-detected for all zip files.

cwisniew:
I think the ability to drag those types of files onto the map but only get a red X and nothing further is done with them is going to be a source of confusion. I will have to update the fix as a suggested.

cwisniew:
Fixed, sadly it wasn't an issue with AddonImport, (that already does the correct thing) the issue is with asset transfer at which point the client doesn't know what was sent it just knows something was sent so it tries to create an asset before saving it to the asset cache. 

I have added the ability for the getMediaType() to check the contents of a zip file to see if it is an addon and only return MTLIB in that case. Since the file is not passed for the version used in the drag and drop code it will not allow you to drop a mtlib (or any zip file).

kwvanderlinde (approved):
> Reviewed 1 of 2 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3449)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cwisniew)

PR #3451, opened 2022-06-30
Fix for UDFs receiving JSON arguments as empty objects

### LINKED ISSUE #3231: [bug] javascript UDFs receive json arguments as empty objects
### Describe the Bug

if you have a js UDF which takes a json it will receive a empty object instead of whatever argument you give it

### To Reproduce

1. create a token name lib:test and set allow uri acess
2. create test macro with 
```javascript
function test(arrayOfObject){	
MapTool.chat.broadcast(JSON.stringify(arrayOfObject));
return arrayOfObject;
}

MTScript.registerMacro("test",test);
```
3. create define test macro with `[r:js.evalURI("test","lib://test/macro/test")]` command
4. create use test with 
``` 
[h:objectWithArray = json.append("[]","array1","dsdas,dsd","array2",4)]
[h:js.test(objectWithArray)]
```
5. see that ouput to chat is empty object instead of array with 4 values

### Expected Behaviour

arguments that the js UDF receives should be the ones i use, and not always an empty object

### Screenshots

_No response_

### MapTool Info

1.11.0

### Desktop

arch linux

### Additional Context

_No response_

dark-ether:

resolves #3231 

### Description of the Change

the discussion of the issue gave me enough context to solve this issue, i searched for where code calls javascript evaluation and added conversions, to ensure that the objects behave exactly as their javascript counterpart new classes that implement the proxy interfaces were added as conversion targets.

### Possible Drawbacks

whenever new forms of evaluating javascript are added it will be necessary to add code to convert the types.

I am converting the arguments for MTScript.getMTScriptCallingArgs before inserting in the stack, which shouldn't be a problem as long as we don't need them as their json counterparts before reconverting them back to MTscript types. 

Besides that when receiving JSON objects and arrays not modified by javascript we get a JSArray or JSObject type which calls the HostObjectToMTscriptType method instead of the expected ValueToMTScriptType but this currently isn't a problem as the 2 classes are instances of  list and  AbstractMap respectively so they are converted correctly to MTScript values however if we were to change that part it could potentially be a problem.

when testing i found that BigDecimal can't be read by javascript and there isn't any way to add the number behaviour, so i am using double, if it is decided that precision is critical and double isn't enough it should be considered to export big decimal to the javascript context

### Documentation Notes

i've attached a zip containing a test campaign showing that it is working now
[testeJsUdf.zip](https://github.com/RPTools/maptool/files/9021184/testeJsUdf.zip)

### Release Notes

- Fixed issue where JSON arrays and objects were received as a empty object to javascript code

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3451)


src/main/java/net/rptools/maptool/client/script/javascript/JSArray.java
```diff
@@ -0,0 +1,86 @@
...
+  public void set(long index, Value value) {
+    checkLong(index);
+    array.set((int) index, value);
+  }
+
+  private void checkLong(long index) {
+    if (index < Integer.MAX_VALUE) {
+      return;
+    } else {
+      String message = "tried to acess index " + index + " which deosn't fit in int";
```
thelsing, line 82: typo

src/main/java/net/rptools/maptool/client/functions/MacroJavaScriptBridge.java
```diff
@@ -147,6 +149,43 @@ public Object childEvaluate(
     }
   }
 
+  public Object HostObjectToJavaScriptType(Object obj) {
+    if (obj instanceof JsonElement) {
```
thelsing, line 153: I would prefer to handle the negation first. It reduces indent for this method and improved readability IMO.

thelsing (approved):
I didn't actual test the changes but the changes look good to me. I only added minor comments.

src/main/java/net/rptools/maptool/client/functions/MacroJavaScriptBridge.java
dark-ether: i could do that however then in the future if it were necessary to add more conversions  it could get messy,as it would need a extra "and" comparison per class that needs conversion. what are the plans for the future? if we want to make javascript a first class scripting language then it is almost certain that adding more conversions would be necessary.

cwisniew: I think leave it as is but since we are using newer version of java you can use pattern matching
if (obj instanceof JsonElement jObj) {
...
}

this removes the requirement for the explicit cast in the next line

cwisniew (changes requested):

Phergus:
*[`src/main/java/net/rptools/maptool/client/script/javascript/JSArray.java` line 82 at r1](https://reviewable.io/reviews/rptools/maptool/3451#-N5pE24ycslXjPk5BxWt-r1-82:-N5zwzae6ZLR7oBsJeqE:b4hp01g) ([raw file](https://github.com/rptools/maptool/blob/b4e70007329a00662c719aefac12a6c79b50aaed/src/main/java/net/rptools/maptool/client/script/javascript/JSArray.java#L82)):*
<details><summary><i>Previously, thelsing wrote…</i></summary><blockquote>

typo
</blockquote></details>

Still a typo:  "deosn't"

Phergus (approved):
:lgtm:

> Reviewed 2 of 4 files at r1, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3451#-:-N6-epUaC5d1YzaJlXDV:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @thelsing)

Phergus:
@cwisniew need you to approve the change you requested.

cwisniew (approved):

Phergus (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3451)* status: all files reviewed, 2 unresolved discussions (waiting on @thelsing)

Phergus (approved):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3451)* status: all files reviewed, 2 unresolved discussions (waiting on @thelsing)

Phergus:
@thelsing you haven't resolved the changes you requested.

thelsing (commented):
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3451)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dark-ether)

Issue #3459, opened 2022-07-05
[Refactoring]: Improvements to visibility calculation for large and complex geometry


kwvanderlinde:
### Describe the problem

The algorithm in `FogUtil::calculateVisibility()` has many inefficiencies that become noticeable when map topology gets complex.
- All faces (with the correct direction) are considered candidates for blocking vision. These need to be sorted and each potentially contributes a shadow. So the visibility calculation suffers when there are many of these faces.
- The algorithm relies heavily on the creation of shadow geometries, and unions them all together to produce a final complete shadow. Each shadow is fairly cheap to create, but when there are many the union operation contributes a fair amount of running time. Although the algorithm limits the number of these shadows by checking if the faces are already contained in a shadow, this is an imprecise check and fairly expensive (though still a net win).
- The algorithm manipulates our AWT-based types (such as `Point2D` and `AreaFace`) only to produce JTS geometry in the end. JTS geometry is generally more efficient, and any conversion from one to the other is also a point of inefficiency.

### The improvement you'd like to see

There are a few main ways I see to improve the visibility calculation:
1. Limit the `VisibleAreaSegment`s to only those within the token's range of vision. Also, cut out individual faces within a segment that lie outside of vision.
2. Use a different algorithm that can build the visible area without creating lots of geometry. E.g., the ["Wall Tracking" algorithm](https://www.redblobgames.com/articles/visibility/#wall-tracking) can produce the visible area as a single polygon without having to union many other geometries or subtract the result from the total vision.
3. Rely more heavily on JTS. E.g., we can avoid `VisibleAreaSegment`/`AreaFace` in the algorithm and instead work on `LineString`s. Then we can rely on that JTS geometry to perform the intersection checks needed for (1).

### Expected Benefits

There are some significant performance gains to be had in the presence of large and complicated topology, and even with not-so-complicated topology.

E.g., an experimental implementation of improvement (1) brings the test case time down from ~21 seconds to ~13 seconds. Improvements (2) and (3) look like they may bring that down further to ~10 seconds. That sounds like a lot, but most of the remaining time is spent elsewhere, specifically in the `Area` unions performed to reveal all the FoW along the path.

### Additional Context

Here is a test map which demonstrates a cave system (complicated topology) with a token with darkvision (limited range): [80x80 Cave.zip](https://github.com/RPTools/maptool/files/9049151/80x80.Cave.zip)

A simple test case is as follows:
1. Import the test map.
   - Reset all fog when prompted.
[...]

PR #3461, opened 2022-07-07
Implement a sweeping algorithm for calculating visibility

kwvanderlinde:

Implements #3459

### Description of the Change

The algorithm for calculating visibility in `FogUtill::calculateVisibility()` has been changed. The previous algorithm was subtractive in that it processed "walls" (`VisibleAreaSegments`) in order to calculate shadows that would be removed from vision. The new algorithm is additive in that it positively identifies the area that remains visible. It does this a 360 degree sweep over all the vertices of the walls in order to determine the furthest extent of vision in each direction. This avoids many of the expensive geometry operations such as unions of polygons and repeated polygon intersection tests.

Another improvement is to ignore walls that do not intersect the vision area. This is a big save when vision is limitted and geometry is complex, such as may be the case in underground cave maps. This improvement can be applied to the previous algorithm for similar improvements.

The methods `AreaContainer::getVisionBlockingBoundarySegments()`/`AreaMeta::getFacingSegments()` now rely on JTS geometries to do their job efficiently (vision intersection tests, in particular). In general, the `AreaTree` and related types are now more JTS-centric, and we avoid custom types unless they provide a solid benefit, as this means less conversion between different representations. As a result:
- `VisibleAreaSegment` is no more. The `AreaFace` data it contains is equivalent to a `LineString`, and its other responsibilities were all specific to the shadow-building algorithm. As there was no more need of these things, we just use `LineStrings` where we used to use `VisibleAreaSegment`.
- `AreaFace` is no more as it was only used in the representation of `VisibleAreaSegment` and `AreaMeta` while providing a facing angle for its line segment. `VisibleAreaSegment` is now gone, and `AreaMeta` only cares about the consecutive points and doesn't use the facing angle anymore (it computes an orientation relative to the origin). So `AreaFace` has been removed in favour of a list of `Coordinate`s.
 
The performance improvements I am seeing from the provided test case are:
- Previous algorithm: ~21 seconds
- Previous algorithm with vision bounds check: ~13 seconds
- Sweep algorithm: ~10 seconds

And some miscellaneous changes:
- The `AreaTreeInspector` demo has been removed as it was specific to the old algorithm.
- The `VisionBlockingAccumulator` has been moved into its own file as `FogUtil` was feeling a bit cluttered.
- A new demo - `VisibilityInspector` - has been added to visualize the result of `FogUtil::calculateVisibility()`. E.g.,:
   VisibilityInspector

### Possible Drawbacks

- Because the algorithm is quite different, there could be a subtle difference in behaviour that is not immediately obvious in testing.
- JTS can be picky about geometry so there could be a problematic case waiting to be discovered.
- As I can't test every case out there, there could be cases where the performance will degrade instead.

### Documentation Notes

N/A (refactoring)

### Release Notes

- Improved the performance of vision calculation on large complex maps.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3461)


FullBleed:
Sweet find and improvement!  :)

cwisniew (approved):

PR #3463, opened 2022-07-08
fix for player.getName() null pointer error

### LINKED ISSUE #3462: [Bug]: player.getName() fails for player on the server if player database is not being used.
### Describe the Bug

When the player.getName() macro function is run on the server it will fail with a null pointer exception if the server was started without a player database.

### To Reproduce

Start a server, ensure that the use easy connect and use player database are not checked.
in chat type [player.getName()]

### Expected Behaviour

It should return the player name and not an error.

### Screenshots

_No response_

### MapTool Info

Version 1.11.x

### Desktop

All platforms 

### Additional Context

_No response_

cwisniew:

resolves #3462

### Description of the Change
player.getName() will now return the name when on the server and player database is not used.

### Possible Drawbacks
This will also alter the results for other player.* functions in the case where they are running on the server and its not using a player database. The risk that it will affect peoples macros should be small as they were previously failing and not providing any meaningful result. 

### Documentation Notes
Should now work as per existing documentation

### Release Notes

- Fixed an issue where `player.getName()` was failing on the server when no player database was being used.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3463)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3463#-:-N6Tv5tEDRTN_SKvpDFj:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cwisniew)

PR #3470, opened 2022-07-09
Improve AreaTree construction in the presence of very sharp internal angles.

### LINKED ISSUE #3466: [Bug]: VBL precision problem.
### Describe the Bug

When there are slight cracks in VBL, MapTool sometimes doesn't know how to process it correctly leading to broken vision.

### To Reproduce

1. Download the test campaign file and rename to remove the .zip extension: [Why can't pcs see themselves.cmpgn.zip](https://github.com/RPTools/maptool/files/9074820/Why.can.t.pcs.see.themselves.cmpgn.zip)
2. Open the campaign file. Notice the PC in the "house".
3. Hover the mouse over the PC. Notice the lack of vision lines.
4. Change to player view (**View** > **Show As Player**). Notice the PC disappear.

### Expected Behaviour

The PC should be able to see the interior of the house but not outside of it. In general, vision should work reasonably even in the precense of subtle VBL features.

### Screenshots

Notice the VBL encompasses the entire house, but the interior is wide open:
image

Zooming in by a lot shows the culprit crack in the VBL at the interior corner of the south-east side of the door:
vbl-crack
This crack does not run all the way through the door, it is actually a very short length.

When we mouse over the token, we should see the white vision lines around the interior like this:
expected-vision

### MapTool Info

1.11.5, develop

### Desktop

Linux Mint 20.3

### Additional Context

_No response_

kwvanderlinde:
Resolves #3466

### Description of the Change

The first important thing here is the reading of `double`s instead of `float` from the `PathIterator`. This increased precision is enough for `AreaMeta` to correctly represent more cases.

The second main change is the replacement of the angle-summing algorithm in `AreaMeta::computeHole()` with `Orientation.isCCW()` from JTS. The latter is more robust and also uses double-double precision to help out in more cases. Since `Orienation.isCCW()` requires an array of coordinates, we just pass in the list of vertices that we would otherwise have to build anyways. As such, the linked list of points is no longer necessary and has been removed.

A little bit of cleanup has also been done:
- `AreaMeta` fields have been encapsulated. No one else has any business modifying its state.
- `GeometryUtil` - which provided the `PointNode` previously used in `AreaMeta` - has been cleaned up as most of its code is no longer used.

### Possible Drawbacks

Shouldn't be any.

### Documentation Notes

N/A

### Release Notes

- Fix a bug where imported VBL with small cracks could break vision.

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3470)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3470#-:-N6Yt2yj8b18x1lcvcQ2:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #3492, opened 2022-07-13
Only init campaign props with default values when wanted

thelsing:
fixes 3487

### Description of the Change

Removed all initialisation code from the getters. Default values are now only initialised with Campaign::initDefault(). This is currently only used in CampaignFactory::createBasicCampaign(). I noticed that we can't set a defaultSight if we have not sights defined so I changed this from string to StringValue in protobuf. I also added readResolve() to CampaignProperties. So all campaigns created with older versions of MT will initialise the various lists and hashmaps even if they were null.

### Possible Drawbacks
I don't think there are any.

### Release Notes

- Fixed an issue where empty lights, states etc. in campaign properties get set to default values automatically

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3492)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/MapTool.java
+++ b/src/main/java/net/rptools/maptool/client/MapTool.java
@@ -949,7 +949,7 @@ public class MapTool {
 
   public static Campaign getCampaign() {
     if (campaign == null) {
-      campaign = new Campaign();
+      campaign = CampaignFactory.createBasicCampaign();
     }
     return campaign;
   }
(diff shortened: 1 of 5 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3492#-:-N6tVird6nj8jj_l1LxW:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @thelsing)

PR #3504, opened 2022-07-19
Validate lastPath when loading tokens

Irarara:
#3254

### Description of the Change
This adds a check to make sure `lastPath` has valid data when loading tokens and throws the data out if it doesn't. 

This is necessary because some older tokens have now invalid paths made up of `AStarCellPoint` nodes, and with the switch to protobuf campaigns with such tokens in them will fail to load (and in the current stable release, if a user tries to revert the path of one of these tokens it will result in an exception).

### Possible Drawbacks
This change does result in the loss of the legacy `lastPath` data and while it would probably be technically possible to restore the path, it seems unlikely anyone will care about the loss of the last path on a token that probably hasn't been touched in at least 7+ years.

### Documentation Notes
N/A

### Release Notes
N/A

---
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3504)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/Token.java
+++ b/src/main/java/net/rptools/maptool/model/Token.java
@@ -2476,6 +2476,16 @@ public class Token extends BaseModel implements Cloneable {
       sizeMap = new HashMap<>();
     }
 
+    // Check to make sure lastPath has valid data
+    if (lastPath != null) {
+      // The second condition seems like it shouldn't be possible, but at some point in the past it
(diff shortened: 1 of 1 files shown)


thelsing (approved):

Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3504#-:-N7R8UIN8g86VU36Tp4t:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Irarara)

Issue #3521, opened 2022-08-05
[Bug]: Sight is broken on a map imported from v1.10.4


fw:
### Describe the Bug

On a map containing VBL, imported from v1.10.4, the token's sight produce a strange and complex shape.
Alternatively, On a map imported from v.11.4, a strange sight shape is not produced.

### To Reproduce

1. Open the attached campaign file.
2. Select the map named "a. made at v1.10.4".
2. Sight produced by the "Hero" token will be a strange shape.

Sample campaign file: [sight-is-broken.zip](https://github.com/RPTools/maptool/files/9273901/sight-is-broken.zip)

-- or --

1. Create a map with VBL on MapTool v1.10.4
2. Export the map.
3. Import the map on MapTool v1.12.0-alpha.1

### Expected Behaviour

Show the token's sight correctly.

### Screenshots

![strange-sight-shape](https://user-images.githubusercontent.com/29275/183220945-d3bf1aed-3b6c-4d1e-9bb2-94d2229e963f.jpg)

### MapTool Info

MapTool 1.12.0-alpha.1

### Desktop

Windows 10 Pro 64bit

### Additional Context

The map "b. made at v1.11.4" in the attached campaign file, the sight shape shows correctly.
On the map named "c. dungeon" and "d. shop", when changing the token's face, the sight shape will be transformed.

I think the problem is related to introducing "Topology" features (adding Pit and Hill).

I made these test maps (a and b) with v1.10.4 and v1.11.4 only. Other versions are not tested.
The maps c and d, those are taken from the campaign file that I run, and created version is not clear, but probably any v1.10.x.

All maps contain VBL only and do not have any map images.

kwvanderlinde:
This is certainly due to the change of vision blocking algorithm I introduced in 1.12. Looks like a point is being projected the wrong direction for some reason. I'll investigate.

kwvanderlinde:
This seems to come down to a precision issue. E.g., on the map `d. shop` there are many points that lie extremely close to one another, such as two adjacent points whose y coordinates are identical and whose x coordinates are `1316.0` and `1316.0000000000002`. Such small distances don't make for good calculations, and our vision blocking algorithm is particularly sensitive to such things. Previous versions may have hidden this because they used lesser precision when processing the VBL and also used a different approach to vision blocking. 

[...]

PR #3525, opened 2022-08-07
Snap AreaMeta vertices to a precision grid

kwvanderlinde:

Addresses #3521

### Description of the Change
The precision grid (not to be confused with MapTool's grid) is provided by `GeometryUtil.getPrecisionModel()`. By snapping to this grid, we avoid needing to check points for "closeness" rather than equality while still avoiding problems with points that would otherwise lie close to one another without technically being the same point. 

The hardcoded precision is high enough that any intentional detail is going to be captured. But it is low enough that any points that would cause issues for the vision blocking algorithm are elided. E.g., the points in #3521 that are within about 2e-13 are elided, while most points further than that can be preserved without issue.

### Possible Drawbacks

Not likely to be any. Released versions of MapTool can't operate at these precisions anyways, so no one should be counting on them one way or another.

### Documentation Notes

N/A

### Release Notes

N/A (fixes a new bug)

- - -
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3525)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3525#-:-N8swsNh4ihWVROuBymo:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #3528, opened 2022-08-08
Fix for add on libproperty compat layer treating everything as strings

### LINKED ISSUE #3322: [Bug]: Numbers stored in Add-On Library properties are Strings.
### Describe the Bug

When a number is stored in the properties of an Add-On library, it is instead stored as a string. So, `[h:setLibProperty("Whatever",0,"Lib:AddOnLib")]` becomes `"0"` instead of `0`.

### To Reproduce

1. Create an Add-On Library
2. Use the setLibProperty() function to store a number in any property
3. `[r:getLibProperty()+1000]` should return "x1000" instead of (x+1000)

### Expected Behaviour

That numbers would be numbers.

### Screenshots

_No response_

### MapTool Info

1.11.4

### Desktop

Windows 10

### Additional Context

_No response_

cwisniew:

resolves #3322

### Description of the Change

Fixes Add-On LibProperty compatibility layer so that things are no longer always treated as strings.

### Possible Drawbacks

MTScript is a fickle mistress and there is sure to be edge cases where conversion between types is not exactly what you want.

### Documentation Notes

See https://docs.rptools.info/docs/add-ons/mtscript-lib-functions#add-on-libraries-and-data

One thing to note is the type is saved along with the data so any lib properties created before this patch will remain strings. To fix this you will need to use the functions from https://docs.rptools.info/docs/data/mtscript-data-functions to fix it.

You will need to remove the value as follows 
```
[h: data.removeData("addon:", "com.example.myaddon", "myVar")]
```
You can then set the value using the functions you used before and it will have the correct type

It is probably better to use the data.* functions than the compatibility LibProperty functions for new code as you have better control over types.

### Release Notes
- Add-On lib properties will no longer always be treated as strings.
- Add data MTScript functions

- - -
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3528)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/DataFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/DataFunctions.java
@@ -19,11 +19,8 @@ import com.google.gson.JsonObject;
 import java.math.BigDecimal;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
-import net.rptools.lib.MD5Key;
 import net.rptools.maptool.client.MapTool;
-import net.rptools.maptool.client.functions.json.JSONMacroFunctions;
 import net.rptools.maptool.language.I18N;
-import net.rptools.maptool.model.AssetManager;
 import net.rptools.maptool.model.gamedata.DataStore;
 import net.rptools.maptool.model.gamedata.DataStoreManager;
 import net.rptools.maptool.model.gamedata.MTScriptDataConversion;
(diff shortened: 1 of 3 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3528#-:-N8xyQaU1Dr2omNWJYBQ:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cwisniew)

Issue #3530, opened 2022-08-08
[Bug]: ClassCastException generated by Tokens with AStarCellPoints in LastPath


Phergus:
### Describe the Bug

When loading older campaigns the tokens may have LastPath data using AStarCellPoints.  This produces an exception causing the campaign load to fail.
```
java.lang.ClassCastException: class net.rptools.maptool.client.walker.astar.AStarCellPoint cannot be cast to class net.rptools.maptool.model.AbstractPoint (net.rptools.maptool.client.walker.astar.AStarCellPoint and net.rptools.maptool.model.AbstractPoint are in unnamed module of loader 'app')
	at java.lang.Iterable.forEach(Iterable.java:75) ~[?:?]
	at java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1092) ~[?:?]
	at net.rptools.maptool.model.Path.toDto(Path.java:288) ~[main/:?]
	at net.rptools.maptool.model.Token.toDto(Token.java:2916) ~[main/:?]
	at net.rptools.maptool.model.Zone.lambda$toDto$29(Zone.java:2273) ~[main/:?]
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197) ~[?:?]
...
```

### To Reproduce

1. Load attached campaign file in 1.12 beta or current dev branch.
2. See exception

### Expected Behaviour

Bad path data should be corrected on load

Remove `.zip` from filename.
[AStarCellPointError2.cmpgn.zip](https://github.com/RPTools/maptool/files/9284868/AStarCellPointError2.cmpgn.zip)
.

### MapTool Info

Current Dev and 1.12.0 beta 1

### Desktop

Windows 10

### Additional Context

Partially corrected in PR #3504.

Phergus:
@Irarara you got this?

Irarara:
👍

Since it was brought up on discord, is there any desire to actually restore the data? It still doesn't really seem necessary to me (since any token with this issue literally hasn't been touched in years), but I can make it happen if others think it's worthwhile.

Phergus:
It guess it's the nice thing to do.   In general I think it is handy to have last path preserved between sessions but as time goes on the information is probably less relevant.  @kwvanderlinde @cwisniew @Azhrei   thoughts?

kwvanderlinde:
[...]

PR #3532, opened 2022-08-09
Add a converter to make CellPoint from pre-1.9.0 AStarCellPoint XML

kwvanderlinde:

#3530

### Description of the Change

There are two cases in which an `AStarCellPoint` can end up having being saved:
- In campaigns created before 1.9.0, `AStarCellPoint` could be freely added to `Path` as it was a subclass of
  `CellPoint`/`AbstractPoint`.
- In campaigns loaded in 1.9.0 or later, if it already had an `AStarCellPoint` in a `Path` it would be deserialized to a
  mostly meaningless object (`null` position in particular). If that campaign is then saved again, the serialized
  `AStarCellPoint` will not have any `CellPoint` fields anymore.

In the former case, the converter builds a `CellPoint` so that the position information is not lost. In the latter
case. the information is already lost so we don't build anything. To account for this, `Path` will clear itself if it
encounters any null points, and `Token` will `null` out any empty `Path`.

### Possible Drawbacks

Because a path now clears itself in some circumstances, the length of the last path will be lost for tokens that have an `AStarCellPoint` in them and which were loaded and saved post-1.9.0.

### Documentation Notes

N/A

### Release Notes

- Fixed a bug causing last path information to be lost when loading campaigns created before 1.9.0.

- - -
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3532)

### DIFF
--- a/src/main/java/net/rptools/lib/FileUtil.java
+++ b/src/main/java/net/rptools/lib/FileUtil.java
@@ -42,6 +42,7 @@ import java.util.regex.Pattern;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 import java.util.zip.ZipInputStream;
+import net.rptools.maptool.model.AStarCellPointConverter;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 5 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3532#-:-N933dSK1zyQ_R3tMwAm:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #3550, opened 2022-08-16
Change JSAPIToken.setProperty() to accept Objects instead of Strings

### LINKED ISSUE #3102: JS: Token.setProperty cannot take numerical value
**Describe the bug**
The Token.setProperty method does not take a number as a value; instead it issues an error.

**To Reproduce**
1. Make a new macro on a token: `[h: js.eval('MapTool.tokens.getSelected().setProperty("HP", "200")')]`
2. Run macro. Observe that the token's HP property is now 200.
1. Make a new macro on a token: `[h: js.eval('MapTool.tokens.getSelected().setProperty("HP", 999)')]` (unquoted value)
2. Run macro.
3. See error, HP is not updated and the following message is in chat:
```
com.oracle.truffle.polyglot.PolyglotEngineException error executing expression js.eval('MapTool.tokens.getSelected().setProperty("HP", 999)').
```

When run via a registered JS function, the error is more explicit: 
```
org.graalvm.polyglot.PolyglotException: TypeError: invokeMember (setProperty) on net.rptools.maptool.client.script.javascript.api.JSAPIToken@15a41497 failed due to: Cannot convert '1'(language: Java, type: java.lang.Integer) to Java type 'java.lang.String': Invalid or lossy primitive coercion. error executing expression js.setHP10(currentToken()).
```

**Expected behavior**
Numbers should work fine with Token.setProperty.

**Screenshots**

**MapTool Info**
- 1.10.3 - Uninstalled older versions then installed

 - Windows 10

Phergus:

Fix #3102 

### Description of the Change

Changed method signature on `JSAPIToken.setProperty()` to take an `Object` for `value` instead of `String` and instead use the `toString()` method of `Object` to pass it on to the `Token.setProperty()` method.

### Possible Drawbacks

Should be none.

### Documentation Notes

n/a

### Release Notes

Fixed issue where numbers passed to `JSAPIToken.setProperty()` would cause a PolyglotEngineException error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3550)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/script/javascript/api/JSAPIToken.java
+++ b/src/main/java/net/rptools/maptool/client/script/javascript/api/JSAPIToken.java
@@ -125,11 +125,11 @@ public class JSAPIToken implements MapToolJSAPIInterface {
   }
 
   @HostAccess.Export
-  public void setProperty(String name, String value) {
+  public void setProperty(String name, Object value) {
     boolean trusted = JSScriptEngine.inTrustedContext();
     String playerId = MapTool.getPlayer().getName();
     if (trusted || token.isOwner(playerId)) {
-      this.token.setProperty(name, value);
+      this.token.setProperty(name, value.toString());
     }
   }
 


cwisniew (approved):

cwisniew:
This API will be deprecated, so a small change like this that stops errors from occurring rather than a sweeping change to allow additional functionality is the best way to go!

Issue #3506, opened 2022-07-20
[Feature]: make console.log correctly handle varargs


perkinslr:
### Feature Request

In standard javascript implementations, `console.log`'s signature is `console.log(...objects)`.  The `console.log` implementation used in the html5 netscape context is `console.log(string)`, silently discarding all but the first argument, which is naivly converted to a string.

Additionally, just calling (or implying) `obj.toString()` on each argument to `console.log` does not match the way arguments are converted to strings by standard implementations of `console.log`.  

### The Solution you'd like

Implement proper varargs wrapper around `MapTool.log(string)`, complete with json encoding arrays and maps, while stringifying other objects.

PR #3559, opened 2022-08-23
Improve html5 console.log

perkinslr:

#3506 

### Description of the Change

The previous PR for #3506 allows `console.log` from an html5 context to display varargs, but introduced some unfortunate edge cases.  

1. Multiple matching primitives were replaced with the recursive-object marker (`console.log(1,1,1)` results in `1 ... ...`).
2. Strings got an extra `"` around them, even when used along (`console.log("hello")` logged `"hello"` instead of `hello`).
3. html tags were rendered to the log instead of logged as text, making diagnosing invisible elements difficult (as they are often invisible in the log too).  

### Possible Drawbacks

The third point might interfere with people using `console.log` to dump html to the chat box, intending it to be rendered as html.  I don't believe this use case has ever been explicitly supported, and given the current threading implications of using `console.log` outside a debugging situation, should *not* be supported.  In any case, the benefit of easily seeing the literal contents of what is logged outweigh this edge case.  

Additionally, the raw log function, which does not do any escaping, is still available as `MapTool.log`.  If the threading issues are resolved in the future, we can advertise its existence for anyone wanting to dump unprocessed html to the chat from an html5 context.

### Documentation Notes

`console.log` from an html5 context accepts multiple arguments, and escapes them such that they are properly visible in the chat box.

### Release Notes

- Fixed an issue where `console.log` would replace primitive types with `...`
- Fixed an issue where `console.log` would incorrectly put `"` around its arguments
- Raw html sent to the chat box via `console.log` is now always escaped.

- - -
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3559)


kwvanderlinde:
The tag replacements and top-level string handling look good to me.

The ellipsis replacement still needs some work. The main issue being the use of the `in` operator, which does not check whether a value exists in an array - it actually checks whether a property exists on an object, e.g., whether an index is present in the array. We should instead use something along the lines of `seen.some(v => v === arg)`.

One thing I'm not clear on with the `seen` list is whether its purpose is to handle recursive objects, or whether it is meant to reduce the verbosity of the output. If it's meant to handle recursion, then it would need to be used as a stack or something in order to do that properly. If it's just meant to reduce verbosity, then it works as-is, though I would expect it to be cleared after serializing each argument. I.e., I would expect this code:
```
const x = {a: 1, b: 2};
console.log(x, x);
```
to result in:
```
{ "a": 1, "b": 2 } { "a": 1, "b": 2 }
```
rather than result in:
```
{ "a": 1, "b": 2 } ...
```

perkinslr:
Good catch...  The `...` is *necessary* to avoid recursion issues.  I hadn't considered it could be the same object in different branches of a tree, so a stack based approach is needed.  I'll ammend the PR presently.

src/main/resources/net/rptools/maptool/client/html5/javascript/Console.js
```diff
@@ -1,16 +1,27 @@
 {
     "use strict";
+    let REPLACEMENTS = [
+	[/&/g, '&#38;'],
+	[/\//g, '&#47;'],
+	[/</g, '&#60;'],
+	[/>/g, '&#62;'],
+	[/=/g, '&#62;'],
```
kwvanderlinde, line 8: `=` should be replaced with `#61`.

kwvanderlinde (approved):

PR #3623, opened 2022-09-12
Hide macro created frames and placeholders after restoring layout

### LINKED ISSUE #3215: [Bug]: Restore Layout malfunctioning, showing empty macro frames from past games on right side
### Describe the Bug

In 1.10.3+, when I use 'Restore Layout', I'm getting unpopulated frames with titles from macro-generated frames created during prior sessions in a tabbed panel on the right hand side. I have not tested on older versions.

In 1.11.0a5 I get the Campaign panel tabbed with the Library panel on the top left, I think this is correct in 1.11.x thanks to #3135.

### To Reproduce

1. Open a new campaign on a computer which has had prior sessions which generated macro frames.
2. Select Window →  Restore layout.
3. See error - right-hand side is occupied by a panel with tabs for old frames.

### Expected Behaviour

I expect the restore layout button to behave consistently regardless of prior macro panels opened on the computer.

### Screenshots

image

Here is a partial list of the tabs that come up (I closed about half so you could see the titles:
image

### MapTool Info

1.10.3, 1.11.0a5, each installed in their own directory.

### Desktop

Windows 10

### Additional Context

Still hoping for #2712 and #1510 ... :)

Irarara:
Fixes #3215

### Description of the Change
Calling `resetToDefault()` to restore the default layout exposes all the placeholder frames needed to maintain layout between launches, so they all need to be hidden again after that is done.

Also, a bit of an aside, in the process of fixing this I learned that `DockingManager.getAllFrames()` exposes the actual key set of the internal map of frames and not a copy, which makes it very dangerous to manipulate. So, that's why I changed that little bit in `MapToolFrame`, because even though it _technically_ hasn't broken anything (yet), I don't want to give the impression that was a safe thing to do.

### Possible Drawbacks
None I know of

### Documentation Notes
N/A

### Release Notes
* Restore Layout could open extra empty windows. Fixed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3623)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/AppMenuBar.java
+++ b/src/main/java/net/rptools/maptool/client/ui/AppMenuBar.java
@@ -14,15 +14,12 @@
  */
 package net.rptools.maptool.client.ui;
 
+import com.jidesoft.docking.DockingManager;
 import java.awt.event.ActionEvent;
 import java.awt.event.ItemEvent;
 import java.io.File;
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.swing.*;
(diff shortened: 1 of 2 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3623#-:-NBmIqiy2kiPV-QV_8JK:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Irarara)

Issue #3626, opened 2022-09-13
[Bug]: Specific arrangement of on-token topology + PC vision causes infinite loop of error dialogs


henczati:
### Describe the Bug

I was trying to create vision blocking for a ship's sails for a movable ship, so topology was saved on-token.

Procedure of topology creation, for each sail (on the same image):
1. Long Wall VBL line.
2. Filled oval Hill VBL intersecting 1.
3. Erase oval Wall VBL of exactly the same size as 2, to remove the segment of 1 where it intersects with 2.
The order of 1 and 2 might have been switched.

If I put a token next to this sails token and set it from NPC to PC (turning on token vision), a seemingly constantly-repeating error dialog (infinite loop?) pops up.

The error in the dialog (without backtrace) is:
```
org.locationtech.jts.geom.TopologyException: found non-noded intersection between LINESTRING ( 208.0 -137.0, 208.0 134.66667 ) and LINESTRING ( 208.0 134.66667, 208.0 134.666667 ) [ (208.0, 134.66667, NaN) ]
```
**Note the `NaN` in the last position at the end.**
Full error details text with the 65 line backtrace in the `error.txt` attachment.
Note that this was produced with a "reduced" scene.  One of the early occurrences in a more complicated scene had 815 lines of backtrace.

WARNING:  If I close the error dialog, there is another of the same underneath it.  I was unable to close all error dialogs, I have to kill the MapTool process when the error occurs to escape it.  While an error dialog is present, the MT window is not re-rendered: moving the error dialog "smears" MT window colors.

Then -- if I did not kill MT yet -- after cca. a minute another error dialog pops-up with:
```
While creating snapshot
java.lang.reflect.InvocationTargetException
```
IMHO, MT probably ran out of some resource and cannot create new instances of the first error dialog anymore.

AFAICT, for the error-loop to occur, I need to have at least 2 such (separate) sail topology definitions on the token, and the PC token has to be (or be moved) into somewhere on the "plane-stripe" between 2 such sail topology definitions.  See the 2 sets of instructions for reproducing the error.  The PC token can be on or off the problematic (sails) token, but as long as it's in the mentioned "plane-stripe", the error occurs.

### To Reproduce

Attached images of steps (in `reproduce-A `[...]`.zip` and `reproduce-B `[...]`.zip`) were made with the "reduced" token (having only 2 separate sail topology definitions).
[...]

PR #3628, opened 2022-09-15
Calculate visibility separately for each topology type.

kwvanderlinde:

Addresses #3626

### Description of the Change

This avoids some pathological cases that can arise from unioning sets of line segments. Many of these cases are handled (noded and dissolved) by JTS, but there are some that cause it trouble depending on how precise the set up is (e.g., two diagonal line segments where one is a proper subset of the other). These cases don't arise within a single topology since they are flat areas, so we should be safe to just handle each one separately.

### Possible Drawbacks

Potential for a performance regression due to `Area` intersections.

### Documentation Notes

N/A

### Release Notes

N/A - fixes a new bug

- - -
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3628)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/FogUtil.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/FogUtil.java
@@ -28,6 +28,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
+import java.util.function.Function;
 import javax.annotation.Nullable;
 import net.rptools.lib.CodeTimer;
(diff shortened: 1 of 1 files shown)


kwvanderlinde:
RE performance: I have tested this on a few complicated maps and have not seen a significant change for better or worse.

Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3628#-:-NC2HDl5DpZx7QLNNF2p:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

Issue #3744, opened 2022-11-22
[Refactoring]: Reduce number of times the exposed area is calculated


kwvanderlinde:
### Describe the problem

When using Individual FoW, the exposed area is recalculated every time fog needs to be rendered. This includes every time the zone is panned or zoomed, even though that has no bearing on the exposed area.

Additionally, we calculate the exposed area twice. We should calculate it at most once per frame because it can be really expensive for complicated FoW.

### The improvement you'd like to see

`ZoneRenderer` would cache the exposed area associated with a given `PlayerView`. If that view ever changes, the exposed area would be recalculated. Also, if FoW is ever explicitly flushed, then the exposed area would also be recalculated.

### Expected Benefits

The first frame that renders a given set of exposed areas will still be expensive, but most frames would be very cheap. Panning and zooming the map would be much more responsive.

### Additional Context

The expensive call that calculates the exposed area is `zone.getExposedArea(view)` in `ZoneRenderer.renderFog()`. While we only call that method once, we actually duplicate its logic a little further down when in the calculation of `tempArea`. I.e., `tempArea` and `combined` are always equal as far as I can tell, so there is no need to recalculate `tempArea` or union it with `combined` as we do.

For comparison, I made a prototypical code change (no cache invalidation) to measure the potential impact. Using the sole map in [this test campaign](https://github.com/RPTools/maptool/files/10061775/slow-fow.cmpgn.renamed-to.zip), I saw an improvement from ~270-350 ms to <20ms for renderFog in most frames. Just removing `tempArea` gives a 2x speed up (naturally) and caching provided the remaining benefit.

kwvanderlinde:
There's one more place where we repeatedly recalculate the exposed area: `AbstractAStarWalker.calculatePath()`. Not bad when only moving one token, but can get sluggish when there are several.

kwvanderlinde:
As this is a pure refactoring and we haven't had complaints yet, I'll close this off. Bug reports can be filed if anything comes up.

PR #3745, opened 2022-11-23
Cache the exposed area used for pathfinding and rendering FoW

kwvanderlinde:

Addresses #3744

### Description of the Change

This PR adds caching of the view-specific exposed areas. There are a few separate changes that come together to make this happen, which can be seen more easily in the commit history rather than the full diff.

The first major change is to simplify the algorithm for calculating the exposed area. The behaviour is the same, but three of the four special cases have been merged as they actually do the same thing (the fourth case is fundamentally different). Also the calculation of the exposed area is now separated from the rendering logic as the rendering logic was the same for all four cases.

The second major change is to move the exposed area calculation into `ZoneView` where it can be cached. Both `ZoneRenderer` and `AbstractAStarCellWalker` now make use that cache to improve their responsiveness.

The final major change is to remove the existing cache of the fog buffer contents. Now that the expensive portion of `renderFog()` is cached in its own right, there is much less benefit to cache the buffer contents. And importantly, the buffer contents would be invalidated anyways in some of the most user-noticeable cases, namely panning/zooming the map. Removing this cache means one less thing for `ZoneRenderer` to keep track of.

#### Sample times

Some (highly variable) timings for typical `renderFog` runs using the linked campaign in #3744:
- On some modernish hardware, went from ~300ms to <20ms
- On some older hardware, went from ~350ms-400ms to <50ms
- On my RPi, went from ~900ms to 50-100ms

Of course, when the cache is invalidated, times still aren't especially good, but are typically about 2x better than before the change.

### Possible Drawbacks

This change is meant to be a pure refactoring, so shouldn't be any issues unless I changed some subtle behaviour or didn't get the cache invalidation correct.

### Documentation Notes

N/A

### Release Notes

- Improved FoW rendering and pathfinding performance for complicated FoW with several owned tokens.

- - -
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3745)


Phergus (approved):
:lgtm:

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3745#-:-NHauhZiCTT6PllGBmcl:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

Issue #3750, opened 2022-11-25
[Refactoring]: Fix the hanging PR test runs


kwvanderlinde:
### Describe the problem

For the last several PRs I submitted, I've noticed that the Windows and Mac checks often run stuck and need to be cancelled and rerun. Sometimes this needs to be done multiple times before a successful build completes.

### The improvement you'd like to see

The tests should run reliably so that the automated PR checks can be completed in a reasonable time frame.

### Expected Benefits

We can rely on the automated checks to bring value to the PR process, rather than just holding things up as a formality.

### Additional Context

The checks always seems to hang in gradle's `:test` task.

kwvanderlinde:
I haven't had the chance to directly test on Windows (and I don't have a Mac), so for now I have to settle for testing on Linux. So my findings might not be the same problem as seen in the PRs, but it's stuff that needs fixing anyways.

I've found that `CampaignPropertiesDialogTest` can sometimes deadlock during the dialog construction. Specifically, the main thread hangs deep inside the `pack()` call, waiting on a reader lock that is blocked by the AWT event queue thread's hold on a writer. Meanwhile, the AWT event queue thread is stuck waiting on the global `Component.LOCK`.

I had success changing this test to run on the Swing thread instead of the main test thread (`SwingUtilities.invokeLater()`), which I think is the right way to avoid the deadlock. But I say that as someone who know next to nothing about the details of Swing, so if there are any Swing experts that want to correct me, please do.

There are also a couple other minor things I've found which I'll also include in a PR, but they are much more rare and don't result in deadlocks.

Azhrei:
That’s likely the issue…

Anything related to the GUI or typical event processing should be done on the Swing thread. MT does have some use of other threads for the A* routing, a rerendering throttle (IIRC), and probably a few other things, which will all conspire to trip people up when writing tests.

I’m not familiar with that/those tests, but I have to wonder why the GUI needs to be constructed in the first place. That class should probably delegate to another class/object for implementation details (Strategy pattern, I think?) and that’s the one that should be tested… (Correct me if I’m wrong…!)

kwvanderlinde:
> That class should probably delegate to another class/object for implementation details ...

If only our logic was so well separated :smile:

kwvanderlinde:
Oh, I should mention how I went about diagnosing the issue, because I had no idea this existing until today.

[...]

PR #3751, opened 2022-11-25
Fix test deadlock and other failures

kwvanderlinde:

Fixes (hopefully) #3750

### Description of the Change

The `CampaignPropertiesDialogTest` is now run on the Swing thread in order to avoid deadlocks that would sometimes occur when running on the main test thread.

The `GetInfoFunctionTest` is now repeatable and not dependent on the order of the tests. The problem was that one test was injecting a dummy `SysInfoProvider` that could be consumed by the other test. This is fixed by always restoring the `SysInfoProvider` implementation after each test.

The `MemoryDataStoreTest` was sensitive to the ordering of its various asynchronous calls. Meanwhile, `MemoryDataStore` does not guarantee the ordering of effects as it runs everything on the common fork-join pool. Sprinkling in some more `.get()` fixes that.

Prior to this change, I could only complete a few test suite runs before encountering a deadlock or other failure. With this change, I have been consistently able to complete thousands of test suite runs, each completing promptly.

### Possible Drawbacks

Should be sunshine and rainbows.

### Documentation Notes

N/A

### Release Notes

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3751)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/getInfoFunction.java
+++ b/src/main/java/net/rptools/maptool/client/functions/getInfoFunction.java
@@ -65,10 +65,15 @@ public class getInfoFunction extends AbstractFunction {
     sysInfoProvider = new MapToolSysInfoProvider();
   }
 
-  // the following is here mostly for testing purpose, until we find a better way to inject
+  // region the following is here mostly for testing purpose, until we find a better way to inject
   protected void setSysInfoProvider(SysInfoProvider sysInfoProvider) {
     this.sysInfoProvider = sysInfoProvider;
   }
+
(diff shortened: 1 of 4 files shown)


Phergus (approved):
:lgtm:

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3751#-:-NHjXMjvA1RDLaEUfmFL:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

Issue #3767, opened 2022-12-06
[Feature]: Make UI a bit more customizeable


thelsing:
### Feature Request

I'd like to change the following things via preferences:
The following ones are quite easy:
- remove Toolbar icons I don't use. (So I can use use MT in a smaller window)
- remove Menu Options (Less confusion for players)
- remove context Menu Options (Less confusion for players)

This one is a bit complicated:
- change what happens on click, doubleclick, hover etc on tokens
a) I use double click on markers to open the notes/gmnotes in a popup.
b) When players doubleclick on not owned token I show the handout in a popup.
c) When I hover over a marker I show the Token name in the left bottom of the map.

### The Solution you'd like

The first part can be implemented a bunch of checkboxes.

For the second part Id like to assign actions to events. In the first step these actions would only include the actions that are available right now, but users could switch map dragging to left mouse button for example. But in the future sth. like assigning macros to actions might be nice.

### Alternatives that you've considered.

Just keep my modifications in my local repo.

FullBleed:
Can you add "Hide footer" to your list?

thelsing:
The fist step would be to just be able to remove stuff locally and be able to import/export the setting. Transfering to client is currently out of scope. I only want to give my players a settings file that they can import to get a stripped down ui.

Players can always change their preferences if they want. If we want to prohibit sth. for players this has to go to the ServerPolicy.

Issue #3708, opened 2022-10-11
[Feature]: Allow changing of Tool and Layer modes via a macro


Jmr3366:
### Feature Request

A function like

setToolsMode("Interaction")
setLayerMode("Hidden")

These commands would mimic the clicking a player / GM normally must do in order to move into these other modes.

### The Solution you'd like

setToolsMode("Interaction")
setLayerMode("Hidden")

### Alternatives that you've considered.

I'm not aware of any other way without external systems

cwisniew:
Before adding this we need to add a way for macro functions to check if they are initiated from a html5 dialog/frame/overlay and fail with an information message because these will certainly not be callable from any macro fired from such a context. 

Also, the layer change/tool change would not occur until after the macro has been completed.

Jmr3366:
You can bypass any check you put in via a "link" or defer method. No is always an option.

cwisniew:
A deferred call should be enough to ensure that the problem of calling it from the HTML5 dialog/frame/overlay won't occur. But in those case the change wouldn't take place until some time after the deferred macro

This also means you wouldn't be able to change layer do something then change back, there would be no way to make that work.

Issue #3678, opened 2022-09-26
[Feature]: Keep on-map and "Map Explorer" selections the same or in sync


henczati:
### Feature Request

This would remove the ability to have 2 different selections at the same time, but would make selections more intuitive.

Also, token group hierarchies suggested in <issue-number> would make possible an arbitrary amount of easily-switchable pre-made token groups to be selected.

IMO, requires #3677.

### The Solution you'd like

When I select some tokens on the map, the same tokens would be selected in the Map Explorer, and vice versa.

aeauseth:
How about when you hold down the space bar for the pointer, any selected tokens would be highlighted on all clients.  Perhaps SHIFT-SPACE or other key combos.  You might be able to leverage halos + macros to do this today but might be messy.  The idea of highlighting token(s) while a key is held down does indeed sound useful.

As for selecting groups of tokens, I remember the game StarCraft where Ctrl+0-9 would assign currently selected units to the numbered control group.  Pressing 0-9 would select units from the numbered control group.  You could probably make a macro + hotkey to something similar today.

henczati:
> How about when you hold down the space bar for the pointer, any selected tokens would be highlighted on all clients. Perhaps SHIFT-SPACE or other key combos. You might be able to leverage halos + macros to do this today but might be messy. The idea of highlighting token(s) while a key is held down does indeed sound useful.

Sounds useful, but I would create a separate issue for it, as it is not directly connected.

henczati:
> As for selecting groups of tokens, I remember the game StarCraft where Ctrl+0-9 would assign currently selected units to the numbered control group. Pressing 0-9 would select units from the numbered control group. You could probably make a macro + hotkey to something similar today.

This should go or be linked to the since created #3680.

PR #3776, opened 2022-12-12
synchronize selected layers between tool and add hotkeys

### LINKED ISSUE #1998: Hotkeys for Layer Selection
It is distracting when I have to look away from what I'm currently working on to click on a button to swap my layer.

I would like hotkeys for each of the layers including the VBL. This way I can focus on the map and not on clicking buttons.

A workable alternative would be a single hotkey which cycled through the layers. This is less desirable as I would have to repeat the same hotkey to get to a desired layer. And if I miss the layer I have to cycle around again.

### LINKED ISSUE #3731: [Feature]: Consistent layer selection for interaction, drawing and template tools.
### Feature Request

Because different tools have their own memory of which layer is current, I have to make sure I am on the expected layer after I switch tools or else risk putting stuff on the wrong layer. I find this inefficient in terms of workflow and also mentally burdensome to remember.

### The Solution you'd like

All tools that use the layer selection box would agree about which layer is currently selected. Switching tools would not switch layers but would leave the last selected layer as current.

Also if there are any other things I'm not thinking of that have their own layer selection, they would also be brought in line with the tools.

I haven't come up with anything.

### Additional Context

It's worth pointing out that template tools and (most) drawing tools already share a layer selection. As far as I'm aware it's just the pointer interaction tool and the delete drawing tool that have their own selections.

As a concrete example, let's say I change to the background layer and drag a token onto it;
image

Then I want to also draw a box on the background layer. So I change to the rectangle drawing tool and get surprised about being back on the token layer:
image

thelsing:
fixes #3731
fixes #1998

### Description of the Change

Moved the LayerSelectionDialog to DefaultTool. Tools are still responsible for displaying it.
The dialog now always sets the active layer in ZoneRenderer. Tool that need to notified of a layer change can overide the selectedLayerChanged method.
(Commented out because if discussion.) I also added CTRL + 1-4 as hotkeys to select layers.

### Possible Drawbacks
Nothing.

### Documentation Notes
/* Layers can now be changed via CTRL - 1 - 4 */

### Release Notes
/* - added CTRL+1-4 hotkeys for layer selection */
- synchronize selected layer across different tools

- - -
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3776)


kwvanderlinde:
Appreciate the synchronization change. Broadly looks good, though can't really look closely at the moment.

But I'm not convinced about the hotkeys, and there hasn't been any discussion on them. My reactions:
- Is there other software based on a layering paradigm that we could emulate? Thinking of whether there is already a defacto standard keybindings that we could leverage. Are these binding that?
- There is a [competing idea](https://github.com/RPTools/maptool/issues/3678#issuecomment-1257499458) for using CTRL+number for token grouping/selection. Very common in RTS, might be valuable in MT as well, but would be incompatible with these layer hotkeys.
- How do these hotkeys age if there is a change to layers? E.g., by adding additional layers, either built-in or custom?

Azhrei:
I believe the Cmd+[1-4] shortcuts should be fine on macOS.  (I've checked the docs and I see that Shift+Cmd+[1-5] are used.)

In regards to future layer expansions, I would expect that Meta+1 moves to the Background layer, and if there are multiple, then repeated Meta+1 keystrokes would cycle through them.  If/when additional layers are added, the keystrokes would/could change, and that could be _very_ confusing!  I picture a new "Foliage" layer that is between the GM/Hidden layer and the Object layer, for example.  (Yeah, silly layer.  But that's not the point. ;) )

I would like to see the hockey changes separated from the other functionality of this patch.  That allows for more discussion while getting the basic refactoring ready and in-place.

thelsing:
> Can we get a function to change layers, too?
There is #3708 for this. I'm not changing any macro stuff for now.

I think users really should be able to to change hotkeys. This is part of #3767

I'll disable the hotkeys for now.

kwvanderlinde (approved):

Azhrei:
I saw this was merged.  We'll need the wiki updated, probably with a note that it applies to v1.13+.  Thanks @thelsing for this. :)

aliasmask:
Will players still be able to bind macros with those hot keys, or have they been removed from the possible choices?

Issue #3426, opened 2022-05-29
[Bug]: Auras with multiple ranges appear as a single color


kwvanderlinde:
### Describe the Bug

Although aura definition share the same syntax as for lights, an aura with multiple ranges will always render as a single color despite the same not being true for lights.

### To Reproduce

1. Add the following to the campaign lights:
    ```
    Aura Test
    ----
    light: circle 10 20#000000 30#555555 40#aaaaaa 50#ffffff
    aura: aura circle 10 20#000000 30#555555 40#aaaaaa 50#ffffff
    ````
2. Add two tokens to the map, equiping one with "light" and the other with "aura"
3. Observe the rendering differences between the two.

### Expected Behaviour

Auras should render each range according to the specified color, just as is done with lights. 

### Screenshots

This shows the differences between how a light and an aura are rendered despite having the same properties:
![image](https://user-images.githubusercontent.com/7492219/170891354-30681735-d3c2-4473-a76a-f6844af00868.png)

### MapTool Info

1.11.5

### Desktop

Linux Mint

kwvanderlinde:
It's not obvious from the screenshot or details, but what's actually happening is that all of the aura's colors are being drawn at the largest distance. So the result is the same as what you would get by equipping 5 different auras that are all 50' but with a different colors.

Pmofmalasia:
The copied aura from the example works correctly for me in 1.13.2:

![image](https://github.com/RPTools/maptool/assets/79408871/68223755-3314-47a0-a7a5-6fbfdbedea6a)

PR #3779, opened 2022-12-14
Render each aura range separately, as is done for normal lights

kwvanderlinde:

#3426

### Description of the Change

Auras now calculate a separate area for each range, and associate those ranges with the corresponding colour in a `DrawableLight`. Previous behaviour was to render the full area of the aura repeatedly, once for each range.

Colourless ("bright") auras are not actually rendered, giving the effect of transparency. This matches the behaviour for lights and allows creating auras that start a certain range away from the token. Previous behaviour was to render such auras as white with a hardcoded alpha of 150. White auras can still be had by explicitly setting the colour to white, and this will also respect the users opacity preferences.

`ZoneView.getLights(Type)` has now been specialized to just handle auras, since that is all it was being used for anyways. Normal lights are handled a different way altogether.

### Possible Drawbacks

Anyone using a purely colourless aura intending for a white value will now get nothing. Setting the aura colour to explicitly white will get a similar effect, though the opacity will respect the user's preferences unlike before.

The `DrawableLight`s for auras are not cached in `ZoneView`, so there is the potential for some perf regression if other changes are made in the area. `ZoneRenderer` does cache the results and only calls it at most once per frame, so the current state of things will be fine.

### Documentation Notes

Auras can have colours just like lights. Assigning different colours to different ranges will result in several colours being shown for the aura. E.g., if we have this aura:
```
Colourful Aura: aura circle 10 20#000000 30#555555 40#aaaaaa 50#ffffff
```
and attach it to a token, it will look like this:
image
Note that the innermost range does not have a colour, and so is presented as transparent.

### Release Notes

- Auras render each of their colours only within the corresponding range. Colourless auras no longer render as white, but are transparent. Colour can be explicitly set to white if needed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3779)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
@@ -1520,7 +1520,7 @@ public class ZoneRenderer extends JComponent
     // Setup
     timer.start("renderAuras:getAuras");
     if (drawableAuras == null) {
-      drawableAuras = new ArrayList<>(zoneView.getLights(LightSource.Type.AURA));
+      drawableAuras = new ArrayList<>(zoneView.getDrawableAuras());
     }
     timer.stop("renderAuras:getAuras");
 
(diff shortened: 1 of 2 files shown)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3779#-:-NJLjJn90ywkoVx5cfRc:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #3787, opened 2022-12-22
make playerAlias getter return null if not DisplayName was set. Accou…

thelsing:
…nt for that change. Centralize generation of String from map in ToString of zone.

fixes 3247

### Description of the Change
Make playerAlias getter return null if not DisplayName was set. Account for that change. Centralize generation of String from map in ToString of zone.

### Possible Drawbacks
Maybe some code depends on zone.toString() to only display the name. I didn't find anything like that though.

### Release Notes
- only show display name of maps if explicitly set

- - -
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3787)


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3787#-:-NJzjrpo-6XepAgPXJs0:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @thelsing)

PR #3801, opened 2022-12-30
Feature campaign dir

### LINKED ISSUE #102: Change the current campaign file format to a directory

thelsing:
resolves #102

### Description of the Change
Since #102 is still a wanted feature I brought the campaign-dir branch up to date.

I made sure that one can only save as a directory and load from it if the file filter was selected. 
The name of the filter contains "(Experimental)" so users should be warned.

I only merged an made it work. So lets all look over this and decide if we want this merged,
if we want changes, or if we scrap it and delete the branch and close the issue.

### Possible Drawbacks
Saving and loading doesn't work any more.

### Release Notes
- added experimental feature to store campaigns as a directory instead of a file.

- - -
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3801)

### DIFF
--- a/src/main/java/net/rptools/lib/BackupManager.java
+++ b/src/main/java/net/rptools/lib/BackupManager.java
@@ -19,6 +19,7 @@ import java.io.IOException;
 import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
+import net.rptools.maptool.util.PersistenceUtil;
 
 public class BackupManager {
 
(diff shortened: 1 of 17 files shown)


thelsing:
I just realized that this also changes how the content is saved. And it can't load the current format.
I think if we want to to create a new format, we should use the protobuf definition and dump the objects as json.

If we support campaign dirs at this time the we shouldn't change the format at the same time. So the content of the directory should match with the content of the current cmpg files.

cwisniew:
> I just realized that this also changes how the content is saved. And it can't load the current format. I think if we want to to create a new format, we should use the protobuf definition and dump the objects as json.
> 
> If we support campaign dirs at this time the we shouldn't change the format at the same time. So the content of the directory should match with the content of the current cmpg files.

This was done before the protobuf changes, and we should move to use those instead now. 

I will close this PR as its probably less relevant now after the changes.

Issue #3747, opened 2022-11-23
[Refactoring]: Improve ZoneView caching in the presence of many sights


kwvanderlinde:
### Describe the problem

Maps do not perform very well when there are many tokens with sight, particularly when individual views are used. In particular, even deselecting tokens can grind to halt even if nothing has changed since the tokens were selected.

There seems to be a few things at play:
1. The visible area cache in `ZoneView` is cleared too eagerly. Even though it can map distinct player views to their own areas, the entire cache is being cleared whenever the player view is changed (such as when selecting / deselecting tokens).
2. The calculation of the visible area is pretty heavy. For each token with sight, we establish its visible area, which involves stacking up all the lights sources along with the token's sight to determine which areas are actually visible. When there are many such tokens, this is extremely expensive.

### The improvement you'd like to see

First, more intelligent invalidation of `ZoneView.visibleAreaMap`. Unless something changes that impacts all views, the cache as a whole should not be cleared. In particular, it should be possible to change views without any invalidation of this cache.

Second, improve to the visible area calculations. We can factor out some common results that can be shared across tokens. For example, the heaviest part of the calculation is probably the need to stack up all the light sources and the token's sight. But that doesn't necessarily need to be repeated for all tokens, e.g., if there is no darkenss or sight multipliers in play, than all tokens would agree on the lit area. We could therefore combine that common result with each token's restriction on vision rather than recalculating everything from scratch for each token.

### Expected Benefits

Performance would be much better when making heavy use of sights, which will improve the user experience. The caching improvement would avoid pauses when selecting and deselecting tokens, while the algorithm improvement would reduce the length of the pauses whenever the calculation need to be redone.

kwvanderlinde:
Found a problem that Expose Last Path doesn't do anything. Working on a PR to fix that up.

kwvanderlinde:
As this is a pure refactoring and we haven't had complaints yet, I'll close this off. Bug reports can be filed if anything comes up.

PR #3813, opened 2023-01-10
Restructure zone view lighting

kwvanderlinde:

Implements #3747

### Description of the Change

The main purpose was to improve the performance of personal lighting. The performance itself is mildly improved, though the main benefit is increased responsiveness when switching `PlayerView`s. There is still room for improvement as personal lights still have a noticeable impact above and beyond normal lights.

The secondary purpose of this change was to enable some different approaches to lights. E.g. I am experimenting with splitting the lumens display out from light blending, which is one of the things that is made easier due to the new `Illuminator` providing a more complete picutre of the lighting situation.

At the core of this change is the addition of `Illuminator` and its result type `Illumination`. An `Illuminator` holds all areas lit by normal lights, and associates them with their lumens value (this is actually done per sight multiplier since that can change the light areas). It can then produce an `Illumination` that holds the various lumens levels as unioned areas, from which we can calculate some derivate results such as which areas are lit and which areas are covered in darkness.  The `ZoneView` is then able to use this "global" result and mix in personal lighting information in order to produce a final result.

The caching in `ZoneView` has also been reworked so that more information can be kept as things change. The main lighting results are cached per `IlluminationKey` (which is just a sight multiplier for now) and which can be reused for every token involved in a view. Other results are cached per `PlayerView`. This includes some results that are specific to the current `PlayerView` being rendered, but now are stored in a `Map` so that they can still be reused even when changing back and forth between `PlayerView`s. As a result, view changes no longer require flushing the `ZoneView`, and correspondingly the `zoneView.flush()` call has been removed from `ZoneRenderer.invalidateCurrentViewCache()`.

A couple minor changes for things discovered during this work as well:
- `FogUtil.calculateVisibility()` is no longer responsible for translating areas according to the visibility origin. That responsibility is now left to the callers, and enables some call sites to be a bit more streamlined then they would otherwise be in terms of managing lit areas.
- The `Direction` enum has been removed as only `Direction.CENTER` was actually used. This also mean we don't have to try and account for it in any caching.
- The `CodeTimer` was tweaked so it doesn't repeat ordering numbers when a single timer is encountered multiple times.

### Possible Drawbacks

There is a bit more going on, so there is the possibility of missing something. E.g., not flushing illumination results properly. This would appear as lights not appearing, or appearing in the wrong place.

### Documentation Notes

N/A (internal change)

### Release Notes

- Mild performance improvement on maps that have many personal lights. Selecting / deselecting tokens will not generally results in long wait times in that case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/rptools/maptool/3813)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/ZoneView.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/ZoneView.java
@@ -867,20 +867,12 @@ public class ZoneView {
       return;
     }
     // Cache it
+    final var illumination = getIllumination(view);
+    // We _could_ instead union up all the individual token's areas, but we already have the same
+    // result via the view's illumination.
     VisibleAreaMeta meta = new VisibleAreaMeta();
-    meta.visibleArea = new Area();
-
+    meta.visibleArea = illumination.getVisibleArea();
     visibleAreaMap.put(view, meta);
-
-    // Calculate it
-    getTokensForView(view)
-        .forEach(
-            token -> {
-              Area tokenVision = getVisibleArea(token, view);
-              if (tokenVision != null) {
-                meta.visibleArea.add(tokenVision);
-              }
-            });
   }
 
   @Subscribe


thelsing (approved):
I don't claim that I completely understand ZoneView but the code looks reasonable. I'm a bit surprised that we don't use the direction of a light anywhere. Don't we have light cones?

kwvanderlinde:
> I don't claim that I completely understand ZoneView but the code looks reasonable. I'm a bit surprised that we don't use the direction of a light anywhere. Don't we have light cones?

That `Direction` enum isn't even really a direction so much as a statement of where on the token a light should originate, i.e., top-left corner, middle right side, etc. Our cones on the other hand always originate at the center, and then rely on their configured arc and offset angles to pick a direction relative to the token's facing.

Phergus (approved):
:lgtm:

> Reviewed 15 of 17 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rptools/maptool/3813#-:-NLbHPGBAW3htiGGJqtr:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @kwvanderlinde)

Issue #3817, opened 2023-01-15
[Feature]: Visualizing light mechanics separately from color


kwvanderlinde:
### Feature Request

Lights (actual lights, and personal lights, but I'm not talking about auras here) are being pulled in a few different directions, especially:
1. As a way to represent / communicate game mechanics (e.g., dim light vs bright light)
2. As a part of the environment, aesthetically illuminating a map (so should blend nicely, etc)

The first was supported in 1.11.5 and earlier, as the colour of a light acted like a tag in addition to determining the colour drawn on the map. Lights of the same colour would have their lit areas merged together and wouldn't blend.

In 1.12.2 and later, the colour of a light only determines which colour is drawn on the map. Lights of the same colour would blend together rather than being merged together.

Trying to support both in one overlay, with both behaviours being based mainly on one property (color) is a bit much since they have inherently different characteristics. Additionally, it could be useful to have both environmental lighting that blend while also getting support for game mechanics.

### The Solution you'd like

I would like to split the current light overlay into two separate overlays.
1. The first overlay (the "light overlay") would render light as part of the environment, i.e., by drawing the light's colour in an area and by blending lights together. This way, lights can be added to the map and add some aesthetic quality (especially if we also improve how lights are rendered!).
2. The second overlay (the "lumens overlay") would show a sort of contour map / elevation map that communications the strongest lumens level at each point on the map. The higher the lumens value, the brighter the colour that is drawn to the map, and darkness would be drawn in black. This would look much like how the default 5e lights look in 1.11.5 and earlier, just merging by lumens instead of merging by colour.
 
Each overlay could be enabled/disabled independently. Those who just want the mechanics given by lumens can use the lumens overlay, those who just want the ambience of lighting can use the light overlay. And they can be comboed, e.g., by temporarily enabling the lumens overlay to check the mechanical situation before going back to just lighting.

On top of that, lumens would be defined for each range of a light source, instead of having one lumens value for the whole thing. This would not only allow more direct modeling of game mechanics, it would also serve as a replacement for the ability to set one colour (or no colour) for the "bright range" of a light and another colour for the "dim range".

The default lights provided by MapTool should also be updated so that they can make use of the new features. E.g., the 5e lights would set high and low lumens values for the lights instead of setting colours for that purpose.

### Alternatives that you've considered.

- Reverting to 1.11.5 behaviour to allow users of the old lighting system to upgrade.
[...]

Issue #3755, opened 2022-11-27
[Bug]: Multiple light sources actually darken each other due to their dim light range when the lumens should be, instead, added to overlapping light or at least not darken.


adventuremagic123:
### Describe the Bug

The problem here is that the multiple light sources are actually darkening each other due to their dim light range.  Instead, the lumens in their dim range should be added to that provided by the other light source.

![Screenshot 2022-11-27 143954](https://user-images.githubusercontent.com/50708405/204158588-8429629e-41cc-45e2-b385-37b9f8804824.png)

### To Reproduce

1. Create multiple light sources whose light overlaps with each other.
2. Notice how the dim light radius darkens overlapping light sources when the lumens should be added.

### Expected Behaviour

Light should add lumens to other overlapping light.

### Screenshots

See above.

### MapTool Info

1.12.2

### Desktop

Windows 10

kwvanderlinde:
I'd like to clarify the how lights work in MapTool, as lumens and colours are entirely separate systems.

**Lumens** establish the mechanics of light source. They determine which light sources are truly lights and which are darkness. They also determine which areas are lit, with darkness and light competing with each other when they overlap. Lumens do not add or blend together when lights overlap. This is working as intended (though it could be an interesting feature request to have lumens add together :slightly_smiling_face:). Also, each light source only has one lumens value which applies to the entire light source and not to a particular range. So, for example, a torch's bright range and dim range are mechanically identical (sounds like another feature request in the making :thinking:)

**Color** is the aesthetic aspect of lights. They don't modify the behaviour of lights in any way except to control which color is used to draw them onto the map. Light colours intentionally blend with one another, though as this bug report points out the blending tends to darken lights instead of brighten them as we would expect.

---

We actually had additive light blending for a short time when 1.12 first came out, but it was pulled for performance reasons. We do want to bring it back, but that will depend on us finding a way to do it without tanking performance.

SherbanWisperwald:
I have read the change log and seen some light changes, but in my opinion the lighting is now worse than before. Is there a possibility to bring the 1.11.5 behaviour back somehow? It was quite usable for determining which regions are lit, dim lit and so on. See examples (all using standard "Candle 5"):
The 1.11.5 Version:
![with_1 11 5](https://user-images.githubusercontent.com/15799278/204394217-582ecd5d-454f-4051-8992-5166c70ebff0.png)
The 1.12.2 Version:
[...]

PR #3837, opened 2023-02-14
Add a lumens overlay and make lights environmental

kwvanderlinde:

Implements #3817
Addresses #3755

### Description of the Change
    
The new lumens overlay is a visualization of the light strength at each point of a map. The strength of the light is determined by its lumens and - as always - darkness obscures light of equal or lesser lumens. The lumens overlay treats lumens less than zero as darkness and renders dark areas as black. On the other end, lumens of 100 or greater are treated as bright areas, and are rendered as a nearly transparent white. The final special case is lumens equal to zero, which is a representation of daylight and is not included in the lumens overlay. Any other lumens values are drawn as shades of grey, with low levels being nearly black, and with higher levels being closer to white.
    
The environmental lights are an improvement to the aesthetics of lights. Lights now blend with one another "additively" (technically according to the Screen blend mode) so that they get brighter when they overlap. After blending the lights, the results are blended with the map using a blend operation that brightens the map while preserving details (similar to overlay blend mode and soft lighting). This make lights look more like they illuminate the map rather than sit on top of it.

Both of these features can be toggled in the *View* menu depending on user preferences and game requirements. The lumens overlay can also have its opacity configured in the Preferences dialog, and the older light and darkness opacities have been removed. Lights no longer have an opacity congiguration since they are conceptually part of the map and not a layer rendered on top.

In order to make the lumens overlay as useful as possible, the lumens values are now attached to individual light ranges rather than to entire light sources. Existing campaigns will have their light sources' lumens values copied to each range. The **Campaign Properties** dialog's _Light_ tab represents per-range lumens as by appending the lumens after the optional range color. So these are now valid ranges:
```
r#ff0000+100
r#aa0000-50
r+100
```
These define, respectively, a saturated red light range at 100 lumens, a desaturated red darkness range at 50 lumens, and a colourless range at 100 lumens.
    
Some refactoring has also been done, particularly for `Light` and `LightSource`. In addition to what was necessary to upgrade existing campaigns, we now ensure (in constructors and `readResolve()`) that fields are only `null` if it actually make sense for them to be that way. Together with that, nullability annotations have been added to most methods and fields in those classes. `Light` is also now immutable so that `LightSource` and others can rely on the values not changing unexpectedly.

### Possible Drawbacks

Since the lumens overlay is rendered as shades of gray, and even black, some of its areas could be confused for soft FoW which is rendered as translucent black.

Lighting syntax is changed, which means existing documentation and guides will be out of date.

### Documentation Notes

Lighting syntax has changed to include lumens information for each light range. For example, the following is a torch definition that models the D&D 5e mechanics of bright (lumens=100) and dim (lumens=50), along with a color:
```
Torch: circle 20#bf973f+100 40#bf973f+50
```

The lumens information can now be visualized with the _lumens overlay_, which can be enabled via the _Show Lumens Overlay_ option in the _View_ menu. This is what the above light looks like when equipped to two nearby tokens:
image

The lights themselves can also be visualized as part of the environment. This can also be enabled via the _View_ menu using the _Show Lights_ option. This is what the same situation looks like with the environmental aspect of lights enabled:
image

A more complete example of what lights can look like when used throughout a map is this demo:
image

### Release Notes

- A lumens overlay has been added for visualizing lumens levels on a map. It can be toggled in the *View* menu.
- The light and darkness overlays have been replaced by environmental lighting. Lights now add together and illuminate the map rather than being rendered on top of the map. The new lights can be toggled in the *View* menu.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/3837)


Phergus (approved):
:lgtm:

> Reviewed 19 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RPTools/maptool/3837#-:-NORbM-T8RGwg0eHLZL1:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #3847, opened 2023-02-27
When copying a campaign, keep the campaign and zone IDs

### LINKED ISSUE #3781: [Feature]: Consistent map IDs
### Feature Request

I've made a library for organizing my maps and associating them with different parts of the campaign, e.g., where they fit in a larger atlas. As part of this, I need to be able to store references to maps that are reliable. This is easy if the maps have constant IDs similar to tokens, which I can almost get via the `"zones"` field of `getInfo("campaign")` and the `"id"` field from `getInfo("map")`. But that ID changes whenever a server is started, meaning I can't actually rely on it.

### The Solution you'd like

Maps would be assigned IDs in a way that is consistent over time and across clients. Starting or stopping a server would not result in different IDs, and each client would agree on a single ID to associate with a given map. Exporting and importing a map would not be expected to keep the same ID.

An added perk would be directly supporting map IDs in macro functions, so that map IDs can be retrieved without digging in `getInfo()`, and so that map IDs can be used in places where map names are currently expected.

Instead of using IDs, I could use map names to refer to maps. The various functions expect map names, so it is necessary to some degree to rely on names anyways. But it's not uncommon for me to change either the display name or GM name for some maps, and that would create a dangling reference.

What I actually do right now is add a specially named token to each map. This token just stores a randomly generated ID in it (actually it can do more, but that's all I use it for now). This works, but it has a couple downsides:
1. The token can get in the way. It's not meant to be interacted with (by players or GM), but is just like any other token. An accidental deletion is all too easy to do and messes everything up.
2. I have to maintain the IDs. Assigning them isn't too bad, though it does have some overhead in terms of generating a random string in MTScript and making sure the special token is present when needed. But the biggest thing is when loading the IDs from the maps - it doesn't take many maps before that can get sluggish. It can be mitigated with some caching, but that also adds more complexity to it.

### Additional Context

_No response_

kwvanderlinde:

Fixes #3781

### Description of the Change

Copying `Campaign` objects (e.g., by starting a server) will now copy the campaign ID and each zone's ID. The `Zone` constructor conditionally copies the ID based on the existing `keepIds` parameter.

Explicit copy actions will still result in new maps being given new IDs.

### Possible Drawbacks

Shouldn't be any.

### Documentation Notes

N/A

### Release Notes

- Fixed an issue where map and campaign IDs would change whenever a server is started.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/3847)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/Campaign.java
+++ b/src/main/java/net/rptools/maptool/model/Campaign.java
@@ -167,6 +167,7 @@ public class Campaign {
    * @param campaign The campaign to copy from.
    */
   public Campaign(Campaign campaign) {
+    id = campaign.getId();
     name = campaign.getName();
 
     /*
--- a/src/main/java/net/rptools/maptool/model/Zone.java
+++ b/src/main/java/net/rptools/maptool/model/Zone.java
@@ -444,6 +444,10 @@ public class Zone {
    * @param keepIds Should the token ids stay the same.
    */
   public Zone(Zone zone, boolean keepIds) {
+    if (keepIds) {
+      this.id = zone.getId();
+    }
+
     backgroundPaint = zone.backgroundPaint;
     mapAsset = zone.mapAsset;
     fogPaint = zone.fogPaint;


Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RPTools/maptool/3847#-:-NPIiS3g4ccpw8hqZn9l:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #3855, opened 2023-03-06
Refactor/3854 spotless for verification

### LINKED ISSUE #3854: [Refactoring]: Replace the direct use of google-java-format with a call to spotless
### Describe the problem

The current formatting check for PRs doesn't provide any value since the results aren't enforced, and because the results disagree with spotlessApply/spotlessCheck. Also, when new versions of google-java-format are released, the current formatting check uses the version even though our spotless configuration remains on an older version, increasing the discrepancy between the two.

At the same time, it would be useful to have a relatively quick formatting verification that is separate from platform verification, as formatting is so easily forgotten by devs.

The `formatting` job for PR verification is changed to run spotless rather than directly running google-java-format. The check should fail if spotless finds any issues.

### Expected Benefits

The `formatting` job will become a more reliable indicator for devs to know whether they have done their part in formatting code.

### Additional Context

_No response_

kwvanderlinde:

Resolves #3854

### Description of the Change

The googlejavaformat action is removed from the formatting PR verification step in favour of running `spotlessCheck`. If the check fails, the PR verification will fail.

I've also updated the spotless gradle plugin to the current latest version (6.16.0) while leaving the backend version of google-java-format where it was to avoid distracting formatting changes.

### Possible Drawbacks

Shouldn't be any.

### Documentation Notes

N/A

### Release Notes

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/3855)

### DIFF
--- a/.github/workflows/verify-build.yml
+++ b/.github/workflows/verify-build.yml
@@ -20,13 +20,14 @@ jobs:
   formatting:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-java@v1
+      - name: Git checkout
+        uses: actions/checkout@v2
+      - name: Set up JDK
+        uses: actions/setup-java@v1
         with:
           java-version: "17"
-      - uses: axel-op/googlejavaformat-action@v3
-        with:
-          args: "--dry-run --skip-sorting-imports"
+      - name: spotlessCheck
+        run: ./gradlew spotlessCheck
   build:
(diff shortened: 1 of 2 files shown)


thelsing (approved):

Phergus (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RPTools/maptool/3855#-:-NQGSwgt45VyP14Fy34C:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kwvanderlinde)

PR #3872, opened 2023-03-17
Allows properties of undefined type to change type

### LINKED ISSUE #3819: [Bug]: Add-on property becomes broken if set as "" for the first time
### Describe the Bug

When setting data on add-ons using `setLibProperty` , if the value is initialized as "" the first time, it becomes *unavailable*, as in:

- The variable name shows using `getLibPropertyNames`;
- Any further call of `setLibProperty` will not change the empty value;
- getLibProperty will always return empty.

This does **not** happen if the variable has been initialized with a non-empty value - variable may be set as empty or any other value without issues. 

### To Reproduce

1. load any add-on or create an empty one (try to attach one as a screenshot).
2. try `{setLibProperty("dummyVar", "", "Lib:Test")}` (replace *Lib:Test* with addon);
3. try `{setLibProperty("dummyVar", "This should be printed", "Lib:Test")}`;
4. try `{getLibProperty("dummyVar", "Lib:Test")}`

### Expected Behaviour

Variable should be changed just like Lib:Tokens

### Screenshots

_No response_

### MapTool Info

1.12.2

### Desktop

Linux [Ubuntu]

### Additional Context

[testlib.zip](https://github.com/RPTools/maptool/files/10434710/testlib.zip)

Test library, initializes some variables called `methodA` (full), `methodB` (empty) (rename to .mtlib)

emmebi:
Allows for an add-on property which got an undefined DataType to be redefined with a different DataType.

Fixes #3819

Fixes #3819

### Description of the Change

When we assign a property in an add-on with an empty string, the property gets the `UNDEFINED` data type; once this happens, it's not possible anymore for the property to change value.

This is because there is an explicit check which controls if we are storing in a property of one type a value which can be converted in that type. No value can be converted into the `UNDEFINED` value, hence the error.

This change relaxes the check so that the conversion logic is invoked only when there is no property with the same name or the property has `UNDEFINED` as the DataType of the value stored.

### Possible Drawbacks

As this is a change in an existing behavior, it's possible that some add-on could break as a result of this. While I fail to see a reasonable use-case for this behavior, it doesn't mean that such an use-case doesn't exist.

### Documentation Notes

Relevant lines
* https://github.com/RPTools/maptool/blob/develop/src/main/java/net/rptools/maptool/model/gamedata/MemoryDataStore.java#L193 - where we check if the data types of the new and old values are compatible
* https://github.com/RPTools/maptool/blob/develop/src/main/java/net/rptools/maptool/model/gamedata/data/DataType.java#L38 - no value can be converted to UNDEFINED
* https://github.com/RPTools/maptool/blob/develop/src/main/java/net/rptools/maptool/model/gamedata/MTScriptDataConversion.java#L80 - converting a string into a value; an empty string is here converted to an UNDEFINED data type

### Release Notes

- Fixed add-on properties becoming unmutable after being assigned the empty string

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/3872)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/gamedata/MemoryDataStore.java
+++ b/src/main/java/net/rptools/maptool/model/gamedata/MemoryDataStore.java
@@ -17,17 +17,6 @@ package net.rptools.maptool.model.gamedata;
 import com.google.gson.Gson;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Collectors;
 import net.rptools.lib.MD5Key;
 import net.rptools.maptool.model.Asset;
 import net.rptools.maptool.model.gamedata.data.DataType;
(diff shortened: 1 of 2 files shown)


emmebi:
Cancelling this PR as it is becoming confusing.

PR #3879, opened 2023-03-22
Add token note types

### LINKED ISSUE #3791: [Feature]: Add proper html support for token notes.
### Feature Request

Currently one can use html tags in token notes, but there is no proper html editor for them. Also the hover display inserts <br> for every linebreak in the notes. One can only work around it by having all html code in one line. 

### The Solution you'd like

I would also like to be able to just copy & paste content from websites into those fields and keep formating.
Best would be a WYSIWYG editor for html with the ability to edit the source code if necessary.

Only using one line of html text. But I still can't paste HTML this way.

### Additional Context

_No response_

thelsing:
resolves #3791

### Description of the Change
This change add a textType to notes and gmNotes of tokesn. The type can be changed in the token editor. 
Currently text/plain, text/html and text/markdown are supported.

Markdown notes also get rendered on the hoverpanel now.

The generated Java code and the form xml was out of sync for two files. I also includes those changes to resync them.

Token without an explicit note type default to text/plain.

### Possible Drawbacks
Nothing if I didn't mess something up.

### Release Notes
Can be folded in the other PR of this FR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/3879)


kwvanderlinde (approved):

thelsing:
I forgot to add the new token properties to the protobuf DTO. Will du so later.

thelsing:
I need to look into some layouting issues that happen when the text is very long.

thelsing:
Couldn't reproduce the issue. The ok button of the EditTokenDialog was pushed outsite the viewable area because the Textarea grew without adding a scrollbar. But it didn't happen again.

kwvanderlinde:
@thelsing Spotless has reared its ugly head once again.

PR #3894, opened 2023-04-01
Show player loading status

### LINKED ISSUE #3890: [Feature]: Let the DM know when a player finishes loading assets
### Feature Request

Info for the DM know when the player is ready to play.

### The Solution you'd like

A simple message in GMs chat would be enough. Maybe some time down the road actual `ClientMessages` with handlers and events to show in a nice separate panel what zone each player is in and their loading progress. 

Asking several times every player if they are in yet every time the map changes 🤢

### Additional Context

_No response_

sauliuskarmanovas:

resolves #3890 

### Description of the Change
In Connections window show what map / zone players are in and if they have loaded assets. Small indicator in the bottom right that shows how many players have finished loading their zone.

### Possible Drawbacks
None idk?

### Release Notes
In Connections window show what map / zone players are in and if they have loaded assets. Small indicator in the bottom right that shows how many players have finished loading their zone.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/3894)


cwisniew:
The problem I see with this is it could get quite spammy if there are a lot of players. Especially if the players jump around between maps, or macros like the teleportation pads are in use. Might be better to add an indicator to the connection window instead

Pmofmalasia:
A connection window indicator would be nice. I think ideal states would be yes/no is any player still loading, and could be clicked/moused over for more info. Since most of the time you just want to know if everyone is loaded in or not anyway, not specific people.

FullBleed:
I have the custom map switcher I use in my framework tell me when players use it to switch maps and where they are.  I usually only have around 4-5 players though and don't find it spammy at all.  But I suppose if I had a lot more players switching maps a lot it could be an issue... so I guess a preference setting for it would be necessary if it was a feature of the built in map switcher.

I do, obviously, think knowing where players are and when they are loaded-in to a new map is a good feature to have though.  ;)

sauliuskarmanovas:
Rewrote everything to show current zone and loading status in connections panel. Not the prettiest code but seems to work.

FullBleed:
Any chance we can get functions to get at this info?  While it makes sense that this would show in the connections panel, I don't really want to have to have the Connections panel open and would rather be able to choose where I want the info displayed.

sauliuskarmanovas:
If you don't want to keep the connections panel open I can add a small indicator at the bottom right of the client next to memory usage. Something like: "[loading icon] 3/5" or "[checkmark] 5/5" with tooltip "{0} out of {1} players have loaded their zone".

If you want something custom then `src/main/java/net/rptools/maptool/client/ui/connections/PlayerListCellRenderer.java` has all the code you need of how the status text is created from a Player object and if you want to subscribe to the status changes then the event is `PlayerStatusChanged`.

FullBleed:
> If you don't want to keep the connections panel open I can add a small indicator at the bottom right of the client next to memory usage. Something like: "[loading icon] 3/5" or "[checkmark] 5/5" with tooltip "{0} out of {1} players have loaded their zone".

Nice.

Would it be a problem to add a tooltip with player names and locations?

sauliuskarmanovas:
Ok added more info to the tooltip. After testing it a bit it seems that there's a small issue with tooltips that it doesn't update when it is on screen. So you got to move the mouse a bit for it to reappear and update

src/main/java/net/rptools/maptool/client/ui/MapToolFrame.java
```diff
@@ -1519,6 +1545,7 @@ private void stopTokenDrag() {
   public void setCurrentZoneRenderer(ZoneRenderer renderer) {
     // Flush first so that the new zone renderer can inject the newly needed images
     if (renderer != null) {
+      updatePlayerStatus(renderer.getZone());
```
cwisniew, line 1548: There is a ZoneDeactivated put onto the event bus when a zone becomes deactivated, 
See HTMLFrameFactory.java for how to subscribe and react to this event. This is the preferred method for reacting to zone changes. You might need to add a new event ZoneLoading for this though. This will also allow others to subscribe to the event in future rather than having to find the right line to add the code.

src/main/java/net/rptools/maptool/client/ui/zone/ZoneRenderer.java
```diff
@@ -1965,10 +1966,21 @@ public boolean isLoading() {
     if (isLoaded) {
       // Notify the token tree that it should update
       MapTool.getFrame().updateTokenTree();
+      updatePlayerStatus();
```
cwisniew, line 1969: Could you also put a new ZoneLoaded event on the event bus :)

cwisniew (changes requested):
Added a few comments, we want to try adding new events to the event bus rather than adding direct calls from zone renderer etc, as its starting (starting? who am I kidding) to get messy.

cwisniew:
It looks like spotless will also need to be run to fix formatting.

./graldew spotlessApply (or .\graldew spotlessApply depending on OS)

src/main/java/net/rptools/maptool/client/ui/MapToolFrame.java
sauliuskarmanovas: Where should I put the listeners to these new events that would make changes to the LocalPlayer?

cwisniew: > Where should I put the listeners to these new events that would make changes to the LocalPlayer?

Create a new class in the player package called `PlayerZoneListener` (or something more appropriate if it listens to other events)

Then in `MapTool.initialize()`  register the listener between the create of the player (line 670) and the creation of the basic campaign (line 671).
https://github.com/RPTools/maptool/blob/2586c63de633919c275945a6d87d556c598b2ab8/src/main/java/net/rptools/maptool/client/MapTool.java#L669-L674

sauliuskarmanovas: Looks like ZoneDeactivated isn't always sent i.e. I saw two ZoneActivated in a row so decide to use the new zoneLoading zoneLoaded events

cwisniew (approved):

PR #3923, opened 2023-04-13
added setZoomLock function git #3438

### LINKED ISSUE #3438: [Feature]: lockZoom()
### Feature Request

The ability to "Lock Zoom" is in the menu bar under **Tool>Lock Zoom**.  I would like to be able to trigger it on clients via a function: lockZoom().

Some use cases:

- Improve the ability to create a more fully featured custom UI overlay.
- Lock Zoom on specific maps where that might be desired when clients switch to them (using custom map selectors).
- Allows us to temporarily Lock Zoom when scaled overlay animations are being used.

### The Solution you'd like

lockZoom(lockState)

**Parameters**

    lockState - The state of Lock Zoom, true(1) or false(0).

_No response_

### Additional Context

It would be nice to be able to do it on a "per map" basis, but that would be added functionality and I don't want to make the feature more complicated than it needs to be.

Also of note is that Lock Zoom is currently a "per client" setting, so clients will be able to "unlock" the zoom if it creates a problem for them.  I could, however, make the argument that there may be situations where a GM might want to set up something that is at a particular zoom level and location and being able to lock the zoom without allowing override might be desirable.... but I don't want the basic feature request to be derailed so, for now, just being able to trigger the lock with a function will provide some new functionality.

Jmr3366:
resolves #3438

### Description of the Change
New function `setZoomLock()` 
requires 1 parameter 0 or 1
Returns nothing
### Possible Drawbacks
none that come to mind
### Documentation Notes
na
### Release Notes
na
Examples:
`setZoomLock(1)`
`setZoomLock(0)`

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/3923)


cwisniew (approved):

cwisniew:
@Jmr3366 The build failed with formatting errors, please run spotlessApply to fix
`./gradlew spotlessApply` or `.\gradlew spotlessApply` on your local branch, commit and push back up to your repo

PR #3964, opened 2023-04-18
Preserve integer nature of execFunction parameters

### LINKED ISSUE #3895: [Bug]:  playStream not working with execFunction
### Describe the Bug

When i try to use playStream with execFunction to play audio it does not work.

### To Reproduce

my code:
[playStream(macro.args)]
[H: songUrl = json.append("[]","https://dl.dropboxusercontent.com/s/ir0f5tr3fkdefxs/rainforest_ambience3.mp3")]
[H: execFunction("playStream", songUrl, 0, "all")]

### Expected Behaviour

Audio should play.

### Screenshots

bug

### MapTool Info

I am using version 1.12.2

### Desktop

Windows 10 Pro 22H2

### Additional Context

_No response_

kwvanderlinde:

Fixes #3895

### Description of the Change

The root of the problem is that when we send `execFunction` parameters to clients, each parameter is represented as a `ScriptTypeDto` which supports doubles strings and JSON values. Numbers (represented as `BigDecimal`) are represented as doubles regardless of whether they are integral or not. When the client receives the parameters, the doubles are converted back to `BigDecimal`, but integral values still keep their fractional part and thus the string representation can't be parsed as integers.

This PR changes the client side conversion to remove zero fractional parts that might exist in the `BigDecimal` on the client side. This guarantees the string representation can be parsed by functions like `Integer.valueOf()`. Note that this change only applies to parameters of `execFunction()`.

I've also added logging for when the callee of `execFunction` fails, so that similar issues in the future can be more easily diagnosed. 

### Possible Drawbacks

None

### Documentation Notes

N/A

### Release Notes

- Fixed a bug where `execFunction()` could not run `playStream()` on clients if two or more parameters were passed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/3964)


kwvanderlinde:
Whoops, I really mess this one up. I'll file another that's actually done against 1.13.

PR #3981, opened 2023-04-19
Fix for lights not updating after token move

### LINKED ISSUE #3979: [Bug]: Lights don't update immediately when token is moved.
### Describe the Bug

After moving a token with a light source attached, the light source will be rendered at the token's previous position until the map is updated (e.g., by moving the cursor).

### To Reproduce

1. Start a new campaign.
2. Make sure that lights are enabled or that the lumens overlay is enabled (_View_ menu,  check one or both of _Show Lights_ / _Show Lumens Overlay_).
3. Add a token to the map and equip a Torch light. Make sure the light is being rendered around the token.
4. Drag the token to a new location, being very careful not to move the cursor when dropping the token.
5. Observe the token being separated from its light.

### Expected Behaviour

Lights should be updated as soon as the token is dropped. No further action should be required to have the light move with the token.

### Screenshots

Here is a token that has been separated from its light
image

### MapTool Info

develop, release-1.13 branches

### Desktop

Linut Mint 21.1

### Additional Context

This is likely related to some recent changes I made in `ZoneView` to run the flush methods on the Swing thread.

kwvanderlinde:

Fixes #3979

### Description of the Change

In PR #3952, I modifed the `ZoneView.flush()` to run on the Swing thread via `SwingUtilites.invokeLater()`. This kept the `ZoneView` consistent with itself, but other components assumed they could update the `ZoneRenderer` immediately after the call to `flush()` was done. By delaying the call using `invokeLater()`, the `ZoneRenderer` would sometime update prior to the `ZoneView` flush being run, causing old state to be rendered until another update.

This PR goes a step further towards guaranteeing that `ZoneView.flush()` is only called on the Swing thread. `ClientMessageHandler` and `ServerMessageHandler` have been updated to run their message handlers on the Swing thread if they are responsible for updating the model or the `ZoneRenderer`. This removes the common cases of the ZoneView being accessed off-thread. With that, we no longer use `SwingUtilities.invokeLater()` inside of the `ZoneView.flush()` methods.

There might still be other places where off-thread access could happen (e.g., app actions run on a background thread). but a complete solution to threading is beyond this PR.

### Possible Drawbacks

Other timing assumptions might be violated somewhere.

### Documentation Notes

N/A

### Release Notes

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/3981)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ClientMessageHandler.java
+++ b/src/main/java/net/rptools/maptool/client/ClientMessageHandler.java
@@ -760,9 +760,12 @@ public class ClientMessageHandler implements MessageHandler {
   }
 
   private void handle(PutAssetMsg msg) {
-    AssetManager.putAsset(Asset.fromDto(msg.getAsset()));
-    MapTool.getFrame().getCurrentZoneRenderer().flushDrawableRenderer();
-    MapTool.getFrame().refresh();
+    EventQueue.invokeLater(
+        () -> {
+          AssetManager.putAsset(Asset.fromDto(msg.getAsset()));
+          MapTool.getFrame().getCurrentZoneRenderer().flushDrawableRenderer();
+          MapTool.getFrame().refresh();
+        });
   }
 
   private void handle(PlayerDisconnectedMsg msg) {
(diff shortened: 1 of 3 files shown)


cwisniew (approved):

kwvanderlinde:
I reran the checks. Maven decided to work this time.

cwisniew:
> I reran the checks. Maven decided to work this time.

The maven repo had a DNS issue which was causing havoc (not just for us, not just for GitHub either) and that's why it was failing.

PR #4045, opened 2023-05-09
Fill the lumens overlay properly

kwvanderlinde:

Fixes a missed case in PR #4003.

### Description of the Change

We removed the logic to clear images returned from `BufferedImagePool` because all the uses of it already either cleared the image or rendered to it completely.

While the lumens overlay also filled the buffer before rendering, it mistakenly did that with `SrcOver`, meaning the existing contents of the buffer were being mixed with the fill color. This has now been changed so that it instead uses `Src` to overwrite the buffer contents.

### Possible Drawbacks

Should be none.

### Documentation Notes

N/A

### Release Notes

N/A (only exists in develop)

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4045)


cwisniew (approved):

cwisniew:
@kwvanderlinde, just to make sure before I merge, is this intended for 1.13 or 1.14

kwvanderlinde:
1.14

PR #4084, opened 2023-05-25
Feature user definable HTML pop stat sheet 

### LINKED ISSUE #1424: Configurable HTML5 Replacement for mouse over stat sheets
Replace the stat sheet for mouse over of tokens with a configurable one (using HTML5). 
The default should be similar to current stat sheet.

### LINKED ISSUE #3149: [UI Update]: Campaign settings - Token properties
### Feature Request

as discussed here: https://discord.com/channels/296230822262865920/902998164951400510/902999222314467380

https://cdn.discordapp.com/attachments/902998164951400510/902999220691288084/unknown.png
the property tab could use an overhaul to make it more user friendly

### The Solution you'd like

A mockup (update):
image

- Clicking New/Add should create a new entry with a unique incremental name e.g. 'property12'
- Two headers: Show Statsheet and Show PropertySheet under those two headers each 3 checkboxes:
 +GM
 +Owner
 +All or Others or Rest (with 'others' you could arguably consider to explicitly NOT show this to gm and owner(s) if those are not checked; with 'all' it would not matter whether gm and owner are checked).
 +optionally: EBP with tooltip: 'editable by player', however if this does not show up on the property sheet this would be covered as well. 

- Basic dropdown: list of available property types where this property fall under. Changing the dropdown will change the list of properties below
- Delete will delete selected property
- 'some preset' is the 'import DD/PF/Cthulhu/Whatever'
- 'replace' replace current set with import
- 'append' is ok or 'merge' 'add' or such
- the text fields are known
- the 'type' is to set it to json object so it shows properly on the property sheet, at the core this dropdown would be:
  + json
  + free
 Optionally replace free for
 + text or string
 + number integer
 + number decimal or real
 + ??

_No response_

### Additional Context

_No response_

cwisniew:
resolves #1424
resolves #3149

I know it shouldn't be a two for one but they are quite connected.

### Description of the Change
This change adds the ability for users to create pop up stat sheets that can be defined using HTML, CSS, JavaScript.
As part of this change the Campaign Dialog Token Properties has also been revamped an no longer requires several sanity roll checks when trying to update token properties.

Apologies in advance for anyone reviewing this as there are a lot of changes to make this all work.
- Campaign token property dialog
- Ability to easily rename / delete token property types
- Every Campaign no longer requires the "Basic" token property type
- MapTool "internal" overlays that can not be interacted with via macros
- New Stat Sheets can appear in any corner or centred along any edge
- CSS generated based on current theme
- Support for built in "add-ons" (to serve above CSS via lib:// urls, and eventually built in stat sheet)
- Utility class for processing handlebars templates (for theme reactive CSS and stat sheets)
- Ability to define stat sheet to use and location at property type level
- Ability to override default stat sheet and location for property for each token (changes to token edit dialog)
- Token Property Type has been added to copy/createToken macros
- Reusable table cell renderer that allows user to edit string in text field or push the "..." button to edit in macro/html editor
- Add Token Property Type selection to the new token dialog.
- Two new events added to event bus for when mouse starts and stops hovering over a token
- GMs can still use the legacy stat sheet should they want to

### Possible Drawbacks
I dread to think what people may not like about this :)

### Documentation Notes

Stat sheets can be defined in an add-on.
See https://github.com/cwisniew/MT-test-stat-sheet for an example really simple stat sheet add on and more details on how to define stat sheets.

Update Campaign Property : Token Properties screen
CleanShot 2023-05-25 at 20 11 24@2x
Its now easier to add property types, delete property types and rename them (in the name text field when selected). There is also the ability to set any token property type as the default (needed as you can now delete or rename "Basic"). Note its not possible to delete the default token property type, you would have to make another property type the default before deleting it. There are also check boxes for all of the different options for stats appearing on the stat sheet, so it no longer a text input looks like you are swearing at your token properties.

C![CleanShot 2023-05-25 at 20 12 17@2x
Default values can be edited in the table if they are simple values, or if more complicated then the '...' button can be used to open an editor. 
CleanShot 2023-05-25 at 20 12 59@2x

Look Ma... No Basic...
CleanShot 2023-05-25 at 20 15 05@2x

Setting at the token property type level
CleanShot 2023-05-25 at 20 17 25@2x

An example of the basic no frills stat sheet from the above example repo
CleanShot 2023-05-25 at 20 29 47@2x  

Using a different theme
CleanShot 2023-05-25 at 20 32 36@2x

The new token dialog has been modified to include property type and stat sheet details
CleanShot 2023-05-25 at 20 18 00@2x

Edit Token Dialog
CleanShot 2023-05-25 at 20 37 19@2x

Stat Sheet pop up at top left
CleanShot 2023-05-25 at 20 38 16@2x

### Release Notes

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4084)


thelsing:
This PR is too big for me to review. Maybe you can split it up into some more manageable chunks?

cwisniew:
> This PR is too big for me to review. Maybe you can split it up into some more manageable chunks?

I will try but you will have to keep in mind when doing the review that if something is added but not used yet it is probably because it's in the next PR. :)

cwisniew:
> > This PR is too big for me to review. Maybe you can split it up into some more manageable chunks?
> 
> I will try but you will have to keep in mind when doing the review that if something is added but not used yet it is probably because it's in the next PR. :)

actually, I take that back, since develop had to be merged into this branch along the way because of conflicts being created, that would create a bunch of PRs harder to review :(

src/main/java/net/rptools/maptool/model/sheet/stats/StatSheetContext.java
```diff
@@ -0,0 +1,338 @@
...
+    return statSheetLocation;
+  }
+
+  /**
+   * Returns the notes of the token.
+   *
+   * @return The notes of the token.
+   */
+  public String getNotes() {
+    return notes;
```
thelsing, line 300: Shouldn't we StringUtil.htmlize the noted to account for the different possible note types possible?

src/main/java/net/rptools/maptool/client/events/TokenHoverExit.java
```diff
@@ -0,0 +1,24 @@
...
+package net.rptools.maptool.client.events;
+
+import net.rptools.maptool.model.Zone;
+
+/**
+ * Event fired when the mouse exits a token.
+ *
+ * @param zone the zone for the event.
+ */
+public record TokenHoverExit(Zone zone) {}
```
thelsing, line 24: Maybe we should have the same members as TokenHoverEnter?

src/main/java/net/rptools/maptool/model/Campaign.java
```diff
@@ -98,6 +89,15 @@ public class Campaign {
...
   // private Map<GUID, LightSource> lightSourceMap;
 
+  /**
+   * Record to hold the arguments for rename token type functionality.
+   *
+   * @param from the name to rename from.
+   * @param to the name to rename to.
+   */
+  public record RenamePropertyType(@Nonnull String from, @Nonnull String to) {}
+  ;
```
thelsing, line 99: Not needed?

thelsing (approved):

src/main/java/net/rptools/maptool/model/sheet/stats/StatSheetContext.java
cwisniew: Updated getNotes and getGMNotes to return HTML

src/main/java/net/rptools/maptool/model/Campaign.java
cwisniew: removed stray semi colon

src/main/java/net/rptools/maptool/client/events/TokenHoverExit.java
cwisniew: Added information to match TokenHoverEnter event.

thelsing:
Another thing to consider: Maybe we should run the notes and tokennotes through the template engine? That way we could add Statblocks containing token properties as notes.

cwisniew:
> Another thing to consider: Maybe we should run the notes and tokennotes through the template engine? That way we could add Statblocks containing token properties as notes.

Isn't that what the HTML stat sheet is though the token stat block? Or do you have other ideas outside of just the stat sheet?

thelsing:
> > Another thing to consider: Maybe we should run the notes and tokennotes through the template engine? That way we could add Statblocks containing token properties as notes.
> 
> Isn't that what the HTML stat sheet is though the token stat block? Or do you have other ideas outside of just the stat sheet?

I'm thinking i things like tracking the number of arrows or usages of an item in a property and us this property to fill the values of an inventory that is kept as a text in the token notes.

PR #4092, opened 2023-05-28
Feature sight dialog revamp

### LINKED ISSUE #4087: [Feature]: Update Campaign Properties - Sight Dialog
### Describe the Problem

The Sight dialog for campaign properties can be difficult for people to decipher and leans further to user-hostile than user-friendly. 

### The Solution you'd like

The dialog should present a more familiar user interface, such as a table where the different values defining sight can be edited.

_No response_

### Additional Context

_No response_

cwisniew:
Fixes #4087 

### Description of the Change
Sight definitions have been updated to use a jtable with all the required values
CleanShot 2023-05-28 at 21 26 30@2x

The bottom pane in the dialog contains a description of all the fields. 
This also introduces a way to store/display HTML documents from the resources folder under `net/rptools/maptool/doc/<lang>/html`, this folder will need to be added to crowdin once the PR is merged.

### Possible Drawbacks
I am sure someone will be upset the text input is no longer there

### Release Notes
- Reworked Campaign Properties - Sight to be more user friendly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4092)


FullBleed:
Very nice.

Are there tooltips on the headers?

kwvanderlinde:
Today we can have multiple personal lights for one sight, just like for light sources. E.g.,
```
MultiSight: circle scale r30#0000ff+50 r60#ff0000+100
```
Based on the screenshot, it looks the new dialog doesn't support this?

cwisniew:
> Today we can have multiple personal lights for one sight, just like for light sources. E.g.,
> ```
> MultiSight: circle scale r30#0000ff+50 r60#ff0000+100
> ```
> Based on the screenshot, it looks the new dialog doesn't support this?

Doh this is going to make any dialog complicated, back to the drawing board

Pmofmalasia:
This looks really nice, and I don't think I would throw it out completely yet. I'm currently moving across the country so can't do a mock-up, but this might still work for multi-ranged sights/lights.

What if there was an additional column (or two) with a button to add/remove layers within that sight. Then, another row could be added within that sight heading that contains the data for that range. Any options that apply to the whole sight (e.g. offset) only get put in the top (master) row. Other things that can vary (e.g. color, lumens) get another entry (preferably with default value the same as the master).

The final result would look similar to the current mock-up for sights with one range. For sights with X number of ranges, the leftmost (Name) column would span X rows, with each of those X rows containing info about one of the different ranges.

Hopefully I've described what I'm thinking of well enough without an image. It's possible I might be anchoring on what's shown in the current mock-up, but I can't really think of another style of input that would solve this issue.

cwisniew:
closing as this is old and needs rework

PR #4101, opened 2023-05-30
Fix for error when executing a token macro as a player with the current map not visible

### LINKED ISSUE #4100: [Bug]: Token macros do not execute when the current map is not visible.
### Describe the Bug

When executing a token macro as a player with the current map not visible it will return that error: 

Could not execute the command: "im 777ACDAA38134C1F9F4391599C0FC29F:[r:getCurrentMapName()]", exception = Cannot invoke "net.rptools.maptool.client.ui.zone.ZoneRenderer.getZone()" because the return value of "net.rptools.maptool.client.ui.MapToolFrame.getCurrentZoneRenderer()" is null

Thats because in the execution of Token macros a lot of thing needs the current zone(that is null when the map is not visible) 

### To Reproduce

1:Create a server and add a token with a macro
2:Join the server as a player
2:Select a token with macros
3:As the GM turn the map invisible for players
4:As player execute a macro

### Expected Behaviour

I think when the map is invisible, the tokens should be removed from the selector(to the player), or at least display a more user-friendly error message in the chat. (I also tried disabling the macro buttons, but I think obtaining them would require extensive modifications to the code.)

PS: I have already developed the code to remove the tokens from the selector, and I will pull request it mentioning this issue.

### Screenshots

_No response_

### MapTool Info

1.13

### Desktop

Windows 11

### Additional Context

_No response_

Kkrafy:
Fixes #4100

### Description of the Change

When executing a token macro as a player with the current map not visible it will return that error:
Thats because in the execution of Token macros a lot of thing needs the current zone(that is null when the map is not visible)

### Possible Drawbacks

None expected 

### Documentation Notes
Fix for error when executing a token macro as a player with the current map is not visible

### Release Notes

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4101)


src/main/java/net/rptools/maptool/client/ClientMessageHandler.java
```diff
@@ -341,8 +341,14 @@ private void handle(SetZoneVisibilityMsg msg) {
...
           zone.setVisible(visible);
-
           ZoneRenderer currentRenderer = MapTool.getFrame().getCurrentZoneRenderer();
+          List<MacroButtonProperties> AllMacroButtons = new ArrayList<>();
+          Collection<GUID> AllTokenIDs = new ArrayList<>();
+          for (Token token : renderer.getZone().getAllTokens()) {
+            AllTokenIDs.add(token.getId());
+          }
+          renderer.getSelectionModel().removeTokensFromSelection(AllTokenIDs);
+          MapTool.getFrame().setCurrentZoneRenderer(null);
```
cwisniew, line 351: The above change should be in the below if statement (where it is setting the current zone renderer to null) because, at the moment, it will get executed for players and GMs when any map is made invisible or visible.

cwisniew (commented):
Thanks for the pull request; it will need a little change to work correctly.

Also, it would be appreciated if you could fill out the details in the PR as that is what gets added to the release notes by github, if there is no real information about the fix in the title and the PR body it will be harder for people to find what was fixed.

src/main/java/net/rptools/maptool/client/ClientMessageHandler.java
Kkrafy: Hi, i updated the commit to put the code below the if statement, now its working.

cwisniew (approved):

PR #4105, opened 2023-06-01
Added multiple gridTypes for SquareGrid

### LINKED ISSUE #3840: [Feature]: Additional options for displaying grid lines
### Feature Request

The current implementation for displaying grid lines is (at best) naff.
Zooming out can eventually leave you with more grid lines on screen than map
image

Equally frustrating can be finding a single colour that doesn't disappear on sections of a map
image

The minimum grid line width can be too wide, and when you increase the grid line width the ugly gets multiplied.

### The Solution you'd like

I want to see additional grid rendering options.
These examples are built by varying the rendering stroke, the colour opacity, and the number of times the grid is drawn.
Implementing different blend modes like burn, dodge, darken, difference, exclusion, hard_light, lighten, multiply, overlay, screen, soft_light, etc. would be additional awesome but that seems a bit harder without JavaFX.

High contrast semi-transparent skinnier grid lines are nicer
Width = 1
image
Width = 5
image

Dashed gridlines are more subtle
image

As are dotted grid lines
image

Just marking grid intersections is my favourite
image

_No response_

### Additional Context

_No response_

goblin-3:
Resolves  #3840 

### Description of the Change

I added an extra item to the adjustGridControlPanelView.form to select a grid type. using a getGridtype in the squareGrid.draw the correct grid type should be displayed 

### Possible Drawbacks

given the extra steps the draw function has to do there may be a slight performance hit

### Documentation Notes
added additional grid type support for square grid 

### Release Notes

added additional grid type support for square grid

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4105)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/tool/gridtool/AdjustGridControlPanelView.form
+++ b/src/main/java/net/rptools/maptool/client/tool/gridtool/AdjustGridControlPanelView.form
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <form xmlns="http://www.intellij.com/uidesigner/form/" version="1" bind-to-class="net.rptools.maptool.client.tool.gridtool.AdjustGridControlPanelView">
-  <grid id="241de" binding="mainPanel" layout-manager="GridLayoutManager" row-count="8" column-count="3" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
+  <grid id="241de" binding="mainPanel" layout-manager="GridLayoutManager" row-count="9" column-count="4" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
     <margin top="5" left="5" bottom="5" right="5"/>
     <constraints>
-      <xy x="10" y="10" width="308" height="258"/>
+      <xy x="10" y="10" width="351" height="318"/>
     </constraints>
     <properties/>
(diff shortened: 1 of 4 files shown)


cwisniew:
@goblin-3 the formatting check has failed, could you please run 
.\gradlew spotlessApply                   (for windows, or)
./gradlew spotlessApply                   (for linux, mac)
 
 To fix the formatting locally, then commit and push your changes to same branch which will automatically get picked up by this PR

Thanks

bubblobill:
Does not resolve #3840,
Does contribute towards resolution, but as mentioned in #3840, the solution needs to be one that can be applied to all existing Grid Types, i.e. Square, Hex_H, Hex_V, and Iso (should probably also leave the door open for implementation of future grid types, Hex_Iso, Triangle and Tri_Iso).
Although I would rather a LineStyle exist in its own enum, we do need the feature of being able to define whiche style to apply which this implements. 
As it is, any changes this makes will not propogate to clients. Player clients cannot access the adjustment tool so the view will only change for the GM. 
To fully resolve the issue we will need to build on this;  enabling clients to select the line style  and incorporate any rendering preferences into data transfer.

bubblobill:
If this doesn't get up I will definitely be stealing bits of it for my future work on Grids

cwisniew:
@bubblobill has been steeling bits and this is out of date

Issue #3505, opened 2022-07-20
[Feature]: Map creation


LKS23:
### Feature Request

Hello

I was writing this on the discord the other day. I'm currently working on a System to randomly create dungeon maps, as a part of a mega dungeon project. Maptool currently offers enough functions & features to create a macro that creates the full VBL, when provided with a sufficient image. However, to use something like that, one manually has to create a map first to use it on. If it was possible to implement a function, that can create maps, one would be able to fully create a full dungeon map via macro, which would be immensely helpful for randomized dungeon crawls.

If the feature is too difficult to produce, I have a possibly easier compromise: A function that lets you set a new map image. This way, the same could be achieved with the copyMap function + setting a new map image.

### The Solution you'd like

**New Function -> createMap(1,2,3,4,5)**
1 - Name -> necessary input
2 - DisplayName -> necessary input
3 - Background (color id / asset ID) -> necessary input
4 - Map (asset id) -> optional input
5 - Fog (color id / asset id) -> optional input

**Further options? (these could be simply taken by the default maptool properties or maybe via additional "setMap" functions)**
Cell Type (SQUARE,HEX1,HEX2,ISOMETRIC,NONE) -> default to maptool properties
Distance per cell -> default to maptool properties
Pixels per cell -> default to maptool properties
Vision distance -> default to maptool properties
AI - (NONE,CELLU,INTEGER) -> default to maptool properties
Lights (NONE,DAY,NIGHT) -> default to maptool properties

### Alternatives that you've considered.

**Alternate / Additional expansion of the existing "g/setMap" functions:**
g/setMapBackground(Mapname,colorid/assetid)
g/setMapImage(Mapname,assetid)
g/setMapFog(Mapid,assetid)

FullBleed:
Would be a nice function.

I think everything in the "New Map" dialog definitely needs to be an option... probably using a json object similar to how create token does it.

kwvanderlinde:
I agree that a JSON object would be a good way to do the configurations, especially since most could have default values.

Here's how I picture the function being defined:
```
createMap(mapName)
createMap(mapName, settings)
```

A full `settings` object would look like this, where each key would be optional:
```json
{
  "display name": "New Map Display Name",
  "player visible": true,
  "vision type": "NIGHT",
  "vision distance": 500,
  "grid": {
    "type": "Vertical Hex",
    "size": 50,
    "units per cell": 10,
    "color": "#ffffff",
    "x offset": 25,
    "y offset": 31
[...]

PR #4128, opened 2023-06-12
Add a createMap() macro function

kwvanderlinde:

Implements #3505

### Description of the Change

First up there is some supporting refactoring to bring paint string parsing and asset ID/URL parsing into `FunctionUtil`. `DrawingFunctions` and its children, and `LookupTableFunction` now use these common methods instead of using their own. These will also be used by `createMap()`.

Second is a tweak to `MapTool.addZone()` to fix an edge case I encountered while testing `createMap()`. In the case that we replace the default map, there were two undesirable effects of how we did the replacement:
1. On the client adding the map, the new map would not be set as current, leaving the client on no map. Otherwise things work fine, the client just has to be manually switched to the new map before they can see anything.
2. On a connected client receiving the new map, an NPE could result. Since we delete the default map first, then send the new map, there is a short time where the client has no maps at all.

The behaviour is now changed so that we add the new map first, then delete the default map. We also always force a map change on the originating client if adding the map results in deleting the default map.

Third is a change to `getInfo("map")` / `getInfo("zone")` to include more information via new keys. The new keys hold info that will be useful when calling `createMap()`, so I wanted easy access to these values. The new keys are:
- `"lighting style"`
- `"has fog"`
- `"ai rounding"`
- `"background paint"`
- `"fog paint"`
- `"map asset"`

Finally is the main change: `createMap()`. This trusted macro function builds brand new maps given only a name and some options. This will allow add-ons and other macros to create new maps from nothing, which can then be populated using existing macros, e.g., `createToken()`, etc. The `createMap()` function supports every option in the **Map Properties** dialog, and also every meaningful property returned by `getInfo("map")` (inlcuding the newly added ones). I aimed for alignment between `createMap()` keys and `getInfo("map")` keys to make it easier to build a configuration object for use with `createMap()`. The documentation notes below detail the exact set of keys that are supported in the configuration.

### Possible Drawbacks

Should be none

#### createMap

**Usage**
```
createMap(name)
createMap(name, config)
```
**Parameters**
- `name` The name of the new map.
- `config` A JSON object of options to apply to the new map.

**config parameter**
`config` is a JSON Object that can contain any of the following fields. Field names are case-sensitive, and all fields are optional:
- `display name`: defaults to none
- `player visible`: defaults to user preferences
- `vision type`: defaults to user preferences
- `vision distance`: defaults to user preferences
- `lighting style`: defaults to `"OVERTOP"`
- `background paint`: defaults to built-in Grass
- `fog paint`: defaults to black
- `has fog`: defaults to user preferences (TODO Implement it)
- `map asset`: defaults to none
- `grid`: JSON object, see below

The `grid` value is another JSON object, with these optional keys:
- `type`: defaults to user preferences
- `color`: defaults to user preferences
- `units per cell`: defaults to user preferences
- `size`: defaults to user preferences
- `x offset`: defaults to 0
- `y offset`: defaults to 0

**Examples**

```
[h: vNewMapId = createMap("New Map", json.set("{}",
        "display name", "Display Name",
        "player visible", json.true,
        "vision type", "NIGHT",
        "vision distance", 600,
        "lighting style", "OVERTOP",
        "has fog", json.true,
        "ai rounding", "CELL_UNIT",
        "grid", json.set("{}",
                "type", "Square",
                "color", "#FF0000",
                "units per cell", 20,
                "size", 200,
                "x offset", 7,
                "y offset", 11
        ),
        "background paint", "asset://da1712226dac0f02c4a9a57ab772856b",
        "fog paint", "asset://d3e430bd9f1e380ec749c42e060f015b",
        "map asset", "asset://014aaa10a10b5ddcbbe006f3c22188c2"
))]
```

### Release Notes

- Fixed a NullPointerException that could result when creating a new map in a blank campaign with connected clients.
- Added the following keys to the result of `getInfo("map")`/`getInfo("zone")`: `"lighting style"`, `"has fog"`, `"ai rounding"`, `"background paint"`, `"fog paint"`, `"map asset"`
- Added `createMap(name, [config])` function to create new maps with configurable options.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4128)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/MapTool.java
+++ b/src/main/java/net/rptools/maptool/client/MapTool.java
@@ -1136,15 +1136,24 @@ public class MapTool {
   }
 
   public static void addZone(Zone zone, boolean changeZone) {
+    Zone zoneToRemove = null;
     if (getCampaign().getZones().size() == 1) {
       // Remove the default map
       Zone singleZone = getCampaign().getZones().get(0);
       if (ZoneFactory.DEFAULT_MAP_NAME.equals(singleZone.getName()) && singleZone.isEmpty()) {
-        removeZone(singleZone);
+        zoneToRemove = singleZone;
       }
     }
     getCampaign().putZone(zone);
     serverCommand().putZone(zone);
+
+    // Now that clients know about the new zone, we can delete the single empty zone. Otherwise
+    // clients would not have anything to switch to, and they would get all confused.
+    if (zoneToRemove != null) {
+      removeZone(zoneToRemove);
+      changeZone = true;
+    }
+
     new MapToolEventBus().getMainEventBus().post(new ZoneAdded(zone));
     // Now we have fire off adding the tokens in the zone
     new MapToolEventBus().getMainEventBus().post(new TokensAdded(zone, zone.getTokens()));
--- a/src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
@@ -20,9 +20,7 @@ import java.awt.Point;
 import java.awt.geom.PathIterator;
 import java.math.BigDecimal;
 import java.util.List;
-import net.rptools.lib.MD5Key;
 import net.rptools.maptool.client.MapTool;
-import net.rptools.maptool.client.MapToolUtil;
 import net.rptools.maptool.language.I18N;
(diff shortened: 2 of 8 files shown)


Azhrei:
Thanks for the doc; I'll get it added to the wiki. 👍  Just like your other pending PR, they'll be "proposed" until they make it into a release after testing. Thanks again!

cwisniew (approved):

PR #4157, opened 2023-06-17
Reimplement the add-on dialog so that it follows the selected MT theme.

### LINKED ISSUE #4154: [Feature]: Refactor Add On Dialog so that it uses applied theme
### Describe the Problem

The Add On Dialog uses JavaFX, and as such, the user's theme is not applied to it

image

### The Solution you'd like

Reimplement the dialog in swing so that the user's theme will be used.

_No response_

### Additional Context

_No response_

cwisniew:
resolves #4154 

### Description of the Change
This PR reimplements the add-on dialog in Swing so that it will pick up the theme the user has selected for MapTool.

### Possible Drawbacks
Should be none

### Documentation Notes
Reimplements the add on dialog in Swing so that it will pick up the theme the user has selected for MapTool.

### Release Notes
- The add-on dialog now respects the theme selected by the user.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4157)


kwvanderlinde:
Cool! The code changes look good, though the dialog itself is much harder to read as a user. There's a few things I would change:
1. The labels need to stand out somehow compared to the values. Previously they were bolded, can we do the same here, or is that strictly controlled by the theme? If we can't, perhaps some judicial spacing is in order.
2. The alignment of values within the columns should be more consistent. E.g., the `Namespace` value is flush left, the `Add-On Name` value is indented a bit, the `License` value is indented more, and the  `Short Description` is _very_ indented. I would expect these all to be flush-left, or consistently slightly indented. This would also help with the first point.
3. The `Git URL` label is out of place under the `Version` and `Short Description` values. With everything on a grid now, this really makes it feel like that label is actually a value. Ideally this label would be under the other labels. Or maybe we should break with the original dialog here and put the `Git URL` beneath the `Website URL`, with the `License` and buttons stacked on the right.
4. The dialog is cut off a bit on the right (see the `Vew Read Me File` button and `Git URL` value below). There's also extra padding on the left compared to the buttons below, making it feel like things have been accidentally shifted over.
5. There dialog doesnt have a title anymore.

***

Some screenshots on the Cyan Light theme to show what I mean.

Original:
image

New:
image

cwisniew:
I will have a play around to fix it... The labels should be easy the alignment and cut off at the edge is going to be more fiddly to get something that works for all themes but doesn't leave too much room on the edge in others where it is not cut off. Its probably my lack of knowledge on the tool but I can't seem to find a way to make the intellij forms work great with resizing correctly for multiple themes (I know intellij itself does it but they also wrote their own l&f and large updates to swing, neither of which I wish to do :) ) 

In other news, I might have to add a feature request to start up MT with a random theme to make it easier to spot these things while developing.

cwisniew:
Bit of rejigging... I had to move the git link and buttons down, which made the vertical size of the dialog a bit larger. With a large font and longer URL/Git URL short description will still get truncated, but the buttons etc should be fine.

Below are a few examples using different themes.


Ahh Large Font
image

Ahh Very Large Font
image


Cyan Light
image

Nord
image

FullBleed:
> In other news, I might have to add a feature request to start up MT with a random theme to make it easier to spot these things while developing.

I like it.  You might have been thinking of it being something for you locally, but I think It would be a whimsical addition to core that would give adventurous people more exposure to various themes.

It would be even better if you could pick particular themes to be in the rotation (or at least types, like "dark/light").  ;)

kwvanderlinde (approved):

PR #4167, opened 2023-06-23
Feature addon metadata

### LINKED ISSUE #4166: [Feature]:  Add on Meta Data
### Describe the Problem

Files such as README.md and license.txt make more sense in the top level of a project than they do in the library/ sub folder. Especially if you are going to use GitHub or GitLab to host the source repository for your add on. But at the moment only files in the library/  folder are available to be read. 

### The Solution you'd like

The files in the top level of the add on should be available from within MapTool under the metadata/ directory. This will allow the use of the same read me and license files for sites like GitHub and the add on dialog, as well as giving add ons access to the files if they wish to display them. 

It should not be limited to just the read me and license file as much of the other data would be beneficial as well, with the library.json file available add-ons would be able to query the version dynamically without having to duplicate the information it else where for example.

Logically copying the readme and license files into the root of the library/, but this could break existing add ons.

### Additional Context

_No response_

cwisniew:
resolves #4166

### Description of the Change
Any files in the root of the mtlib zip file also get entries in the metadata/ directory

### Possible Drawbacks

If someone has a metadata/ directory already that contains the same filename it will be overwritten

### Documentation Notes
Files at the top level of the .mtlib zip file will be available lib://<namespace>/metadata/<filename>

### Release Notes
- Files at the top level of the .mtlib zip file will be available lib://<namespace>/metadata/<filename>

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4167)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/library/addon/AddOnLibraryImporter.java
+++ b/src/main/java/net/rptools/maptool/model/library/addon/AddOnLibraryImporter.java
@@ -209,7 +209,7 @@ public class AddOnLibraryImporter {
         Asset asset =
             Type.fromMediaType(mediaType)
                 .getFactory()
-                .apply(namespace + "/" + METADATA_DIR + path, bytes);
+                .apply(namespace + "/" + path, bytes);
         addAsset(asset);
         pathAssetMap.put(path, Pair.with(asset.getMD5Key(), asset.getType()));
       }
--- a/src/main/java/net/rptools/maptool/model/library/addon/AddOnLibraryImporter.java
+++ b/src/main/java/net/rptools/maptool/model/library/addon/AddOnLibraryImporter.java
@@ -59,6 +59,9 @@ public class AddOnLibraryImporter {
   /** The name of the file with the stat sheets. */
   public static final String STATS_SHEET_FILE = "stat_sheets.json";
 
+  /** The directory where metadata from the root dir of the zip directory is copied to. */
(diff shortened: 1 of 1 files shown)


src/main/java/net/rptools/maptool/model/library/addon/AddOnLibraryImporter.java
```diff
@@ -191,6 +197,25 @@ public AddOnLibrary importFromFile(File file) throws IOException {
...
+    var entries = zip.stream().filter(e -> !e.getName().contains("/")).toList();
+    for (var entry : entries) {
+      String path = METADATA_DIR + entry.getName();
+      try (InputStream inputStream = zip.getInputStream(entry)) {
+        byte[] bytes = inputStream.readAllBytes();
+        MediaType mediaType = Asset.getMediaType(entry.getName(), bytes);
+        Asset asset =
+            Type.fromMediaType(mediaType)
+                .getFactory()
+                .apply(namespace + "/" + METADATA_DIR + path, bytes);
```
kwvanderlinde, line 212: Is it intentional to add `METADATA_DIR` to `path` a second time here?

cwisniew: Would you believe me if I said yes? No? yeah I wouldn't either... 

I have updated so that the metadata dir doesn't get appended twice to the name of the asset

kwvanderlinde (approved):

PR #4181, opened 2023-06-29
Fix for MapTool Freeze when using an asset that can't be found on macro button

### LINKED ISSUE #4163: [Bug]: Missing macro image freezes maptool
### Describe the Bug

I have a few left over macros on old tokens that use a missing asset as an image for the macro.
chrome_jSUHSB5QoP

This was fine in 1.12.2 and simply displays a red X to indicate the asset is missing.

In 1.13.1 clicking on the token which contains this macro will freeze Maptool to the point of needing to End Task.
Clicking on the macro without the 'Selected' window, thus never displaying the asset, avoids this behavior.

### To Reproduce

1. Make a new macro and paste `` into the label (this asset does not exist)
2. After clicking apply maptool will lock up if the macro is visible to you

### Expected Behaviour

Return to 1.12 behavior in displaying a missing asset image or text

### Screenshots

_No response_

### MapTool Info

v1.13.1

### Desktop

Windows

### Additional Context

_No response_

cwisniew:
fixes #4163

### Description of the Change
Changed the client handler put asset call so that the adding of the asset to the asset manager occurs on the current thread rather than a deferred Swing thread. As the memory cache is thread safe, this is fine, and it avoids the issue where Swing is blocking waiting for an image which currently locks up MapTool.

### Possible Drawbacks

Should be none

### Documentation Notes
Missing assets used in macro buttons OR  tags in chat or frame/dialog should no longer cause MapTool to lock up.

### Release Notes
- Missing assets used in macro buttons OR  tags in chat or frame/dialog should no longer cause MapTool to lock up.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4181)


cwisniew:
branch stuff up some how

PR #4208, opened 2023-07-14
#4207 #4206 Changed the ConnectionInfoDialog to show active interface…

### LINKED ISSUE #4206: [Bug]: ConnectionInfoDialog does not show the active interface
### Describe the Bug

If there are more then one networkinterface on the local machine, the ConnectionInfoDialog does not show the active interface, but the last one.

### To Reproduce

Goto serverinformation while having more then one ethernet interface while the first interface is active.

### Expected Behaviour

You will get the local address of the last interface.

### Screenshots

_No response_

### MapTool Info

1.13.2

### Desktop

Windows10

### Additional Context

_No response_

### LINKED ISSUE #4207: [Feature]: ConnectionInfoDialog does not show the IPv6 Address if available
### Describe the Problem

ConnectionInfoDialog only shows IPv4 information

### The Solution you'd like

Add the IPv6 information of the local active interface to the ConnectionInfoDialog.

_No response_

### Additional Context

_No response_

PeterPeet:
… and IPv6 Address

fixes #4206
resolves #4207

### Description of the Change
Added a method that differentiates between the active and non active network interfaces and is able to return the IPv6 Address. Then I added a field to the ConnectionInfoDialog and differnetiated between IPv4 and IPv6, so that the dialog is now able to output both IPs at the same time if applicable.

### Possible Drawbacks
Positive side-effect is that if you get a native IPv6 Address and configure that in your router to be passed through the firewall the IPv6 Address shown in the ConnectionInfoDialog should be the one that could by used by others to connect to your game via IPv6 when using the direct connection tab. There shouldn't be negative impacts. 

### Documentation Notes
The InfoDialog is self-explanatory in my opinion. 

### Release Notes
- ConnectionInfoDialog shows active interface information for IPv4 and IPv6 if available.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4208)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/connectioninfodialog/ConnectionInfoDialog.java
+++ b/src/main/java/net/rptools/maptool/client/ui/connectioninfodialog/ConnectionInfoDialog.java
@@ -16,8 +16,12 @@ package net.rptools.maptool.client.ui.connectioninfodialog;
 
 import java.awt.GridLayout;
 import java.io.IOException;
+import java.net.Inet6Address;
 import java.net.InetAddress;
-import java.net.UnknownHostException;
+import java.net.NetworkInterface;
+import java.net.SocketException;
+import java.util.Collections;
+import java.util.Enumeration;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


Azhrei:
On my phone so I can’t check this easily…When you take out the resolving attempt of www.RPTools.net, how do you know you’re going to get the proper interface later, when a connection is actually attempted?For example, a system with two NICs (such as wired and wireless, or two wired) might have one as the default and yet the other one might be used when connecting to the outside world due to routing table rules.  The simplest way to detect this is to attempt to reach “the outside world”.This patch would always retrieve the first address reported for the default NIC, but not the first address for the NIC actually used later.I also saw that the call to super() was removed.  Why was that?--Frank EdwardsEdwards & Edwards Consulting, LLCHours:  10A-6P ETPhone:  (813) 406-0604Sent from my iPhone.  No electrons were harmed in the creation or transmission of this message.On Jul 14, 2023, at 2:00 PM, PeterPeet ***@***.***> wrote:﻿… and IPv6 Address
Identify the Bug or Feature request

Description of the Change

Possible Drawbacks

Documentation Notes

Release Notes




You can view, comment on, or merge this pull request online at:
  https://github.com/RPTools/maptool/pull/4208

Commit Summary

  276cbfc  #4207 #4206 Changed the ConnectionInfoDialog to show active interface and IPv6 Address


File Changes  (2 files)

  
    M
    src/main/java/net/rptools/maptool/client/ui/connectioninfodialog/ConnectionInfoDialog.java
    (61)
  
  
    M
    src/main/java/net/rptools/maptool/client/ui/connectioninfodialog/ConnectionInfoDialogView.form
    (42)
  


Patch Links:

  https://github.com/RPTools/maptool/pull/4208.patch
  https://github.com/RPTools/maptool/pull/4208.diff

PeterPeet:
That with the super was another thing I was testing and I overlooked it in the commit. You are totally right, I had the assumption, that only one interface at a time is active, but as you stated correctly that is wrong I will have to check like before. I will add that to my change.

PeterPeet:
Sorry for the bugs and edits, it is my first pull request of this type. I think I got it right now, but could not test it on Mac or Linux.

cwisniew (approved):

PR #4217, opened 2023-07-26
Save and load server preferences

### LINKED ISSUE #4216: [Feature]: load and save "StartServerDialog" settings from the dialog
### Describe the Problem

If you play with multiple groups you have to fill in the ServerDialog by hand every time you start a game with different settings. In the discord we discussed that it would be great if it would be possible to save and load different setups from the dialog.

### The Solution you'd like

Save and Load different setups from the dialog in XML.

Refactor the dialog, so that it uses protobuf to generate DTO objects from json and then export and import that into the dialog, to get it transitioned to the new method. It was also a wish to create a dropdown and populate it with the last exports, so that you can load a setup directly without the open dialog, but in my opinion it is easy and quick enough to just open different files.

### Additional Context

_No response_

PeterPeet:
resolves #4216

### Description of the Change
I added two buttons on the StartServerDialog to export and import the settings as XML. I also added the translation for them in the i18n-bundle. 

### Possible Drawbacks
* I do not check the XML-File for the correct well formed structure and if it is valid to import.
* I override the selected XML-File without requesting a confirmation from the user, if a file of that name already exists.
* If the export or import fails half way through the process the data model and the view can get out of sync temporarily until the dialog is reopened or one of the operations: export, import or start of the server is reexecuted, but this also happens, if you edit the dialog so it is not a really bad thing.
* To export and import the setup, I programmed a get-Method for the prefs object. I don't know if this is a good way to program it, but it works that way. 
* I did not refractor the dialog in the progress to not depend on XML as Craig suggested.

### Documentation Notes
Export to XML: You can save your actual server configuration into a XML-File. The selected file will be overwritten.
Import from XML: Load a prior exported XML-File with a server configuration into the StartServerDialog.  

### Release Notes
- Export and import server preferences added to StartServerDialog.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4217)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/startserverdialog/StartServerDialog.java
+++ b/src/main/java/net/rptools/maptool/client/ui/startserverdialog/StartServerDialog.java
@@ -280,7 +280,7 @@ public class StartServerDialog extends AbeillePanel<StartServerDialogPreferences
                 prefex.exportSubtree(outStream);
                 outStream.close();
               } catch (IOException | BackingStoreException | IllegalStateException prefserr) {
-                MapTool.showError("Error while Exporting, please try again.");
+                MapTool.showError("ServerDialog.error.xmlExport");
               }
             });
   }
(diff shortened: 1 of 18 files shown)


cwisniew (changes requested):
I am hesitant to merge this as is, as mentioned we want to move to JSON rather than XML, and if I include this, then when it's fixed later on, we would need to support two formats. for ever and a day. 

It should be easy to change this to protobuf to read/write JSON format.

You can take a look at src/main/proto/library.proto  to get an idea of how to create the .proto file, src/main/java/net/rptools/maptool/model/library/token/LibTokenConverter.java to see how to write out the protubuf as a JSON and src/main/java/net/rptools/maptool/model/library/addon/AddOnLibraryImporter.java on how to read a JSON back to protobuf.

I can help with questions if you have them

PeterPeet:
Yes, I am your opinion. This should not be merged, but redesigned.

PR #4262, opened 2023-08-25
Block loading/saving campaign from/to installation dir

### LINKED ISSUE #4209: [Feature]: Safer Installation/Updating.
### Describe the Problem

1)  When installing MT it defaults to installing over the previous version/location.  If the previous version is not removed first, this causes problems with the install.  The previous version at the default location needs to be Uninstalled first.

2)  When Uninstalling MT, everything in the install directory is destroyed.  This causes data loss for users who think the installation directory is safe.

The existing guardrails are not sufficient (as noted by the number of issues people have with failed "Updates" and/or destroying their data.)

I know that some of these issues have been discussed and are the result of limitations of the packaging software, but I hope that more can be done to address the problems.  These issues are fairly common pain-points for new users and do not reflect how most software updates and uninstalls and done.  Please suggest alternatives or more viable solutions if you have them.

### The Solution you'd like

A)  The default installation directory should be based on the name of the MT Version.  This would incentivize new installs to be done in a safer location.

B)  If possible, disallow installation over an existing MT install location.  If that's not possible, perhaps provide a Warning at installation that if an existing version of MT exists at the installation location that it needs to be Uninstalled first (and that the Uninstall will destroy data in the directory.)  People often click through installation instructions believing most of the text to be boilerplate, but warnings at the point of install should be more effective than expecting users to read the readme or release notes. 

C)  If possible, disallow saving any MT specific file (campaign, table, rptok, macroset, etc.) to anywhere in the install directory and inform people during the attempt that they are attempting to save to an Unsafe Location that will be erased when MT is uninstalled.  If the save can't be disallowed, a Warning would be better than nothing.

D)  If possible, have a warning at Uninstall to inform users that all data/assets saved in the install location will be deleted.

If there is a packaging option that provides either non-destructive uninstalls or more seamless updates over existing installs, then that would be preferable.

If none of the better guardrails mentioned above can be done, perhaps a notice on first run explaining these two issues could be added.  (i.e." "Do not save data to the Maptool Installation Directory."  "Before Updating, Uninstall the previous version of MT.  Please note that this will destroy all data saved in the Maptool Installation Directory.  Do not save data to the Maptool Installation Directory.")

A warning about the perils of Updating at the point of the notice of a new available version could also be displayed:  "There is a new version of MT available.  Before Updating:  1) Make sure that you have not installed any data to the Installation Directory.  If you have--Move It Now.  2) You will need to Uninstall previous versions of MT if you are going to install the update to the same location.  This will result in the destruction of all data you may have saved in that directory.  If you have--Move It Now."

### Additional Context

_No response_

cwisniew:
Partially addresses #4209 

### Description of the Change
This change will stop users from saving campaign files into the installation directory.
If the user tries to do so a dialog will appear telling them that they are unable to do so and after clicking Ok they will be returned to the save dialog where they can choose another location.

As well as this if the user attempts to load a campaign file from the installation directory a dialog will appear informing them its not supported and asking them to copy the file elsewhere and try again. In this case they will not be returned to the load dialog.

### Possible Drawbacks
I was not able to address the other requirements like default install to use version number as all attempts I tried just broke the msi builder.

### Release Notes
- MapTool will now block users from saving or loading campaigns from the installation directory.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4262)


kwvanderlinde:
When I put myself in the shoes of a user that have fallen into this trap, I see two problems on the loading side (saving side seem fine to me):
1. Not all users may readily know where there installation directory is, and no direction given to them on how to find it.
2. Even if a user does know where it is, it's a lot of traversal require to remedy the situation. First navigating to the install directory in the file explorer, then navigating somewhere else to put the file, then going back to MapTool and navigating again to the new location via the Load Campaign dialog.

We can help the user out a bit more by:
1. Changing the message to read something like: "MapTool does not support loading the campaign from the installation directory. Please select a new location for the file."
2. When they close the message, we immediately open a save dialog and let them pick the new location right there.

Azhrei:
I do like forcing them to **Save As** immediately.

However, if they Cancel out of that dialog (or just close it), the **File > Save** option should be disabled, forcing them to use **File > Save As...** and picking a new location.

cwisniew:
Changed it so file is still loaded but user gets this message
CleanShot 2023-08-27 at 22 17 08@2x
This occurs from both load and recent campaigns



Then when trying to save 
CleanShot 2023-08-27 at 22 17 23@2x


I decided against opening the save dialog right away as some is going to just click ok on the dialog when its first presented and ben prompted with a save dialog and be confused.

FullBleed:
Does this extend to when people try to save other MT assets to the install directory like tokens, tables, campaign props, macro sets, etc?

cwisniew:
> Does this extend to when people try to save other MT assets to the install directory like tokens, tables, campaign props, macro sets, etc?

The updated change blocks Campaign, Tokens, Tables, Props, Macros, Maps from being saved into the install directory. It does not yet stop screenshots (they are handled differently. so I will get to them in a later change, they are also less likely to be a problem.)

kwvanderlinde (approved):

PR #4267, opened 2023-08-28
Add macro function getIllumination()

### LINKED ISSUE #3565: [Feature]: getIllumination()
### Feature Request

In my framework, you can assign multiple types of light/dark-sources to a token. When attacking a token, i want to calculate how bright the target is illuminated and determine a modifier for the attack roll according to this value. Since there are multiple light sources in my campaign with different modifiers i would also like to get the lumens of the illumination or the brightest light source, that is illuminating my target.

### The Solution you'd like

I would like to call getIllumination(token) to get the lumens of the brightest light-source or the brightest light-source that is illuminating the token.

Since my framework does not support magical darkness yet, i used to determine the illumination manually with 2 scripts. All my light-sources have 4 Parts in 4 different Radii (e.g. "Torch 1", "Torch 2", "Torch 3", "Torch 4"). The lights are turned on/off by a custom UI so when using a Torch all the technically lights "Torch 1-4" are turned on/off at once. My scripts look like this:

getIllumination():
```[h: target = arg(0)]

[h: info = json.indent(getInfo("map"), 2)]
[h: visiontype = json.get(info, "vision type")]

[h,if(visiontype == "NIGHT"),Code:
{
	[h: newTarget = copyToken(target)]
	[h: tokens = getTokens("json", json.set("",
		"layer", json.append("[]", "token", "object", "background"),
		"range", json.set("", "token", newTarget, "from", 0, "upto", 24),
		"light", json.set("", "value", 1, "category", "DSA")
	))]
	[h: brightness = 4]
	[h,foreach(tok, tokens),Code:{
		[h: newTok = copyToken(tok)]
		[h: setHasSight(true, newTok)]
		[h: setSightType("Normal", newTok)]
		[h: isTargetVisible = json.length(canSeeToken(target, newTok))]
		[h,if(isTargetVisible > 0): brightness = min(brightness, getBrightnessFrom(target, tok))]
		[h: removeToken(newTok)]
	}]
	[h: removeToken(newTarget)]
	[h: macro.return = brightness]
};
{
	[h: macro.return = 0]
}]
```

getBrightnessFrom():
```[h: target = arg(0)]
[h: illuminator = arg(1)]
[h: distance = getDistance(illuminator, 1, target, "NO_GRID")]
[h: lights = getLights("DSA", "json", illuminator)]
[h: radius = 0]
[h: currentLight = ""]

[h,foreach(light, lights),Code:{
	[h,if(endsWith(light, "4") == 1 && currentLight == ""): currentLight = light]
}]

[h,if(currentLight != ""),Code:{
	[h: cmp = json.indent(getInfo("campaign"), 2)]
	[h: dsalights = json.get(json.get(cmp, "light sources"), "DSA")]
	[h,foreach(light, dsalights),Code:{
		[h: lname = json.get(light, "name")]
		[h,if(lname == currentLight): radius = json.get(light, "max range")]
	}]
};{}]

[h,if(radius == 0),Code:{
	[h: macro.return = 0]
};{
	[h: width = scale(getSize(target)) / 2.0]
	[h: steps = (distance - width) / (radius / 4.0)]
	[h: brightness = max(0, min(4, floor(steps)))]
	[h: macro.return = brightness]
}]
```

This works almost fine. One problem occurs when a light-source is not hitting the center or corner of the target. In this case canSeeToken() does not work but thats fine. Whats more problematic is the magical darkness that i wanted to add support for in my framework. I would like to add negative lumen lights for this case. This works well for the visuals in the map. But the calculation of the illumination modifier when attacking a token should get very complicated at this point.

Since maptool should internally know more details about the illumination of a token i think its easier for framework-devs to get a function for all this.

### Additional Context

_No response_

kwvanderlinde:

Implements #3565

### Description of the Change

This adds a new macro function `getIllumination()` that returns the lumens level at a given point on a given map. Whether personal lights and sight magnifiers are included can be controlled via parameters, allowing the function to be used for at least these cases:
- Querying what the user currently sees on the current map.
- Querying for lumens levels on any map independent of token vision
- Querying for lumens levels according to the vision of specific tokens.

### Possible Drawbacks

Should be none

#### getIllumination() function

Get the lumens level at a given point on a map.

**Usage**
```
getIllumination(x, y)
getIllumination(x, y, mapNameOrId)
getIllumination(x, y, mapNameOrId, tokenIds)
getIllumination(x, y, mapNameOrId, json.null)
```

**Parameters**
- `x` - The X coordinate to check, measured in pixels.
- `y` - The Y coordinate to check, measured in pixels.
- `mapNameOrId` - The map to check the illumination of. Defaults to the current map.
- `tokenIds` - The tokens IDs whose personal lights and sight magnifiers should contribute to the result. Defaults to those in the current player view. Passing `json.null` indicates that token-specific views should not be used (i.e., as if no tokens were selected on the map).

**Examples**
To check the illumination of the point (100, 200) as seen in the player's current view:
```
[r: getIllumination(100, 200)]
```

To check the illumination of the point (100, 200), without including personal lights or magnifiers:
```
[r: getIllumination(100, 200, getCurrentMapId(), "[]")]
```

To check the illumination of the point (100, 200) as if no tokens were currently selected:
```
[r: getIllumination(100, 200, getCurrentMapId(), json.null)]
```

To check the illumination of the point (100, 200) including the sight of a token named "Elf" (and no other tokens):
```
[r: getIllumination(100, 200, getCurrentMapId(), json.append("[]", findToken("Elf"))]
```

### Release Notes

- Added a new macro function `getIllumination(x, y, [mapNameOrId, [tokenIds]])`

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4267)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/MapToolExpressionParser.java
+++ b/src/main/java/net/rptools/maptool/client/MapToolExpressionParser.java
@@ -43,6 +43,7 @@ public class MapToolExpressionParser extends ExpressionParser {
               ExecFunction.getInstance(),
               FindTokenFunctions.getInstance(),
               HasImpersonated.getInstance(),
+              IlluminationFunctions.getInstance(),
               InitiativeRoundFunction.getInstance(),
               InputFunction.getInstance(),
               IsTrustedFunction.getInstance(),
--- a/src/main/java/net/rptools/maptool/client/functions/IlluminationFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/IlluminationFunctions.java
@@ -1,0 +1,78 @@
+/*
+ * This software Copyright by the RPTools.net development team, and
+ * licensed under the Affero GPL Version 3 or, at your option, any later
+ * version.
+ *
+ * MapTool Source Code is distributed in the hope that it will be
+ * useful, but WITHOUT ANY WARRANTY; without even the implied warranty
+ * of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * You should have received a copy of the GNU Affero General Public
+ * License * along with this source Code.  If not, please visit
+ * <http://www.gnu.org/licenses/> and specifically the Affero license
+ * text at <http://www.gnu.org/licenses/agpl.html>.
+ */
+package net.rptools.maptool.client.functions;
+
+import java.awt.geom.Point2D;
+import java.math.BigDecimal;
+import java.util.List;
+import net.rptools.maptool.client.MapTool;
+import net.rptools.maptool.client.ui.zone.PlayerView;
+import net.rptools.maptool.client.ui.zone.ZoneRenderer;
+import net.rptools.maptool.language.I18N;
+import net.rptools.maptool.util.FunctionUtil;
+import net.rptools.parser.Parser;
+import net.rptools.parser.ParserException;
+import net.rptools.parser.VariableResolver;
+import net.rptools.parser.function.AbstractFunction;
+import net.rptools.parser.function.Function;
+
+public class IlluminationFunctions extends AbstractFunction {
+  private static final IlluminationFunctions instance = new IlluminationFunctions();
+
+  private IlluminationFunctions() {
+    super(0, Function.UNLIMITED_PARAMETERS, "getIllumination");
+  }
+
+  public static IlluminationFunctions getInstance() {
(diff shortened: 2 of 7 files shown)


cwisniew (approved):

Lector:
I tested this on v1.14.2 and it works fine. Thanks for adding this :)

PR #4284, opened 2023-09-11
onChangeMap event

### LINKED ISSUE #1574: Add an onMapChange Event
When switching maps it can be good to have an event triggered on Lib:Tokens to set moods or sounds.

An `onMapChange` event called on implementing Lib:Tokens could provide an API for triggering such moods and sounds.

I don't see that there any alternatives, other than creating a macro for each sound or mood and manually clicking on them. However, this will put much burden on the GM to get everything right instead of interacting with the players.

Jmr3366:
Resolves #1574  

### Description of the Change
When changing maps, any library token with a macro named `onChangeMap` will execute.

### Possible Drawbacks
For any pre-existing library tokens with a macro named `onChangeMap`, this could cause unintended macro execution and/or unexpected results.

### Documentation Notes
Any time the map is changed and this includes first loading a campaign, the `onChangeMap` macro will execute if found on a library token.

### Release Notes
Examples:

Create a macro on an existing or new library token, named `onChangeMap`.
Within this macro place:
`[h: broadcast(getPlayerName() +" " + getCurrentMapName(),"self")]`

When you change the map, your `Player Name` and the current map name will output to chat.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4284)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppActions.java
+++ b/src/main/java/net/rptools/maptool/client/AppActions.java
@@ -14,6 +14,8 @@
  */
 package net.rptools.maptool.client;
 
+import static net.rptools.maptool.client.functions.MapFunctions.ON_CHANGE_MAP_CALLBACK;
+
 import com.jidesoft.docking.DockableFrame;
 import java.awt.*;
 import java.awt.event.ActionEvent;
@@ -498,6 +500,12 @@ public class AppActions {
           }
 
           MapTool.serverCommand().enforceZone(renderer.getZone().getId());
+          List<Token> libTokens = EventMacroUtil.getEventMacroTokens(ON_CHANGE_MAP_CALLBACK);
+          String prefix = ON_CHANGE_MAP_CALLBACK + "@";
+          for (Token handler : libTokens) {
+            EventMacroUtil.callEventHandlerOld(
+                prefix + handler.getName(), "", handler, Collections.emptyMap(), true);
+          }
         }
       };
 
--- a/src/main/java/net/rptools/maptool/client/AppActions.java
+++ b/src/main/java/net/rptools/maptool/client/AppActions.java
@@ -14,8 +14,6 @@
  */
 package net.rptools.maptool.client;
 
-import static net.rptools.maptool.client.functions.MapFunctions.ON_CHANGE_MAP_CALLBACK;
-
 import com.jidesoft.docking.DockableFrame;
 import java.awt.*;
 import java.awt.event.ActionEvent;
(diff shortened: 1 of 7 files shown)


src/main/java/net/rptools/maptool/model/player/PlayerZoneListener.java
```diff
@@ -54,6 +59,8 @@ public void OnZoneLoading(ZoneLoading event) {
 
   @Subscribe
   public void OnZoneLoaded(ZoneLoaded event) {
```
cwisniew: You shouldn't add your code to this class.
Instead you create a new class that listens for the event and register it with the event bus the same way that this class has been registered.

src/main/java/net/rptools/maptool/client/AppActions.java
```diff
@@ -498,6 +500,12 @@ protected void executeAction() {
           }
 
           MapTool.serverCommand().enforceZone(renderer.getZone().getId());
+          List<Token> libTokens = EventMacroUtil.getEventMacroTokens(ON_CHANGE_MAP_CALLBACK);
```
cwisniew: I don't think this would be needed if you are going to use the onZoneLoaded event

cwisniew (changes requested):

src/main/java/net/rptools/maptool/model/player/PlayerZoneListener.java
Jmr3366: I will attempt to do so.

src/main/java/net/rptools/maptool/client/AppActions.java
Jmr3366: In my testing, when I did not have that code in AppActions, the event wouldn't trigger on a Tools -> Enforce Players to Current Map.  So a GM can change maps and the event fires.  Then if a GM pulls all the players to the map they're on, the event fires for the players but without this, it wouldn't fire _again_ for the GM (as they're on the map already)

Jmr3366: This has been removed (2b is after spotless).

src/main/java/net/rptools/maptool/model/player/PlayerZoneListener.java
Jmr3366: This should be correct (and moved the `onChangeMap` to the legacy switch case)

cwisniew (approved):

PR #4296, opened 2023-09-21
Extended getTokenByID to search other Maps

### LINKED ISSUE #3069: JS: MapTool.tokens.getTokenByID should be able to check other maps and return null instead of an error when ID isn't found.
**Describe the bug**
MapTool.tokens.getTokenByID has two issues:

1. It cannot find tokens on maps that aren't the current map.
2. When the token ID is invalid or exists on a different map, it returns an error which breaks the macro, rather than 'null':
```
java.lang.NullPointerException: Cannot invoke "net.rptools.maptool.model.Token.getId()" because "this.token" is null error executing expression js.evalNS("test", ' MapTool.tokens.getTokenByID("C03957B1EDB44293BF8B471B5867C99B"); ').
```

I think a null result would be more useful. There may be a way to handle that error that I don't know about. If there is, it isn't obvious to me 

**To Reproduce**

1. Create a new campaign and add a 'Hero' token. Impersonate the token and run `[r: currentToken()] in chat to obtain the ID.
2. Run the following macro, replacing the ID with your own:  
```
[r: js.evalNS("test", 'MapTool.tokens.getTokenByID("A80B7011AF66452CBD425448BF153434")')]
```
3. Macro runs, returning the ID of the token that has been fetched.
4. Make a new map. Run the macro again.
5. See error.

**Expected behavior**
1. I expect to be able to choose the scope of my search for the ID to either the current map, a specific map (or list of maps), or all maps.
2. I think getting a null result or an empty string when the token isn't found would make more sense.

**MapTool Info**
- Version: 1.10.2
- Install: Uninstalled all other versions and freshly installed.

 - OS: Windows 10.

**Additional context**
Identified by @perkinslr in [this Discord discussion](https://discord.com/channels/296230822262865920/890318378043068456/899177380684378152).

### LINKED ISSUE #3609: [Bug]: New Add-On added to Campaign in 1.12 betas are not saved correctly.
### Describe the Bug

When adding an Add-On in 1.12 betas, the Add-Ons are saved as empty assets in the campaign file. 
The next time you try to open the campaign, you will get an error.

### To Reproduce

Add an Add-On to a new campaign, save the campaign, and then try to load the campaign at which point you will get an error.

### Expected Behaviour

The add-on should be saved correctly to the campaign file.

### Screenshots

_No response_

### MapTool Info

1.12 Betas

### Desktop

All

### Additional Context

_No response_

ColdAnkles:
… for tokenID if not found on current Map.

Closes #3069

### Description of the Change
Extends the functionality of the JS MapTool.Tokens.getTokenByID(id) function to attempt a search of other maps if token with id not found on the current map.

### Possible Drawbacks
Not great if token on the last map and there's a large quantity of maps.

### Documentation Notes
Attempts to get token from current map, if not found, searches all maps for token, returning null if not found at all.

### Release Notes
 - Fixed an issue where MapTool.Tokens.getTokenByID(id) did not return token with id.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4296)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/script/javascript/api/JSAPIToken.java
+++ b/src/main/java/net/rptools/maptool/client/script/javascript/api/JSAPIToken.java
@@ -177,4 +177,11 @@ public class JSAPIToken implements MapToolJSAPIInterface {
   public boolean isOwner(String playerID) {
     return this.token.isOwner(playerID);
   }
+
+  @HostAccess.Export
+  public boolean isOnCurrentMap() {
+    Token findToken =
+        MapTool.getFrame().getCurrentZoneRenderer().getZone().getToken(new GUID(this.getId()));
+    return this.token == findToken;
+  }
 }
--- a/src/main/java/net/rptools/maptool/client/script/javascript/api/JSAPITokens.java
+++ b/src/main/java/net/rptools/maptool/client/script/javascript/api/JSAPITokens.java
@@ -101,7 +101,7 @@ public class JSAPITokens implements MapToolJSAPIInterface {
       List<ZoneRenderer> zrenderers = MapTool.getFrame().getZoneRenderers();
       for (ZoneRenderer zr : zrenderers) {
         findToken = zr.getZone().resolveToken(uuid);
-        if(findToken!=null){
+        if (findToken != null) {
           token = new JSAPIToken(findToken);
           break;
         }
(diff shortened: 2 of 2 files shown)


cwisniew (changes requested):
We should have a function that checks only the current map as well, maybe getMapTokenByID() that only searches the current map and returns null if the token is not on the current map.

cwisniew:
@ColdAnkles 
I assume this is for issue #3069 and not 3609,

ColdAnkles:
Yeah, issue number was a typo.

token.isOnCurrentMap() as another option?

cwisniew:
> Yeah, issue number was a typo.
> 
> token.isOnCurrentMap() as another option?

Might be worth adding that as well.
Sorry for asking now :)

cwisniew (approved):

PR #4299, opened 2023-09-23
Cover VBL Addition

### LINKED ISSUE #3732: [Feature]: New VBL - Cover
### Feature Request

The existing suite of VBL tools are excellent, but one use case I've come across and not found a satisfactory solution to is cover: an area of a map that hides anything within, but does not prevent those within from seeing out.

### The Solution you'd like

A new VBL definition alongside the existing three: Cover.

Tokens outside the cover area cannot see into the cover area - identical to wall VBL in this case.
Tokens inside the cover area are unaffected by the cover area - identical to both hill and no VBL in this case.

Relevant `drawCoverVBL()` `setTokenCoverVBL()` etc. also added.

I am currently using Wall VBL on a "terrain" token, and removing the VBL when it is the turn of a token inside of the terrain, and replacing it at the end of their turn.

It works, but is not very good, and requires fiddly, easy to mistake VBL removal/reapplication.

### Additional Context

The Hero Cannot See into Cover
image

The Troll in Cover, is unaffected, and can see the Hero outside.
image

ColdAnkles:
Adds Cover VBL
closes #3732 

### Description of the Change
Adds cover VBL to available vision options.

Tokens outside cover cannot see into it - effectively identical to wall.
Tokens inside cover see out of it - effectively nonexistent.

### Possible Drawbacks
Limited use? Unwanted? Complicates colours of VBL view.

### Documentation Notes

The Hero Cannot See into Cover
image

The Troll in Cover, is unaffected, and can see the Hero outside.
image

### Release Notes
- Added Cover VBL

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4299)


ColdAnkles:
This is a more complicated contribution than my first, so while I think its acceptable, I've got it in draft to allow for some commentary/feedback first.

thelsing:
So this is basically a one may mirror? If so it works like hill-vbl but also supports lines. Maybe it could be foldet into it?
The question is how do you see wich side you can see through from. 
I think draft-PRs are only necessary if you don't want the PR merged yet. If you think it could be merged just use a normal PR. IMO people will be more likely to look at "real" PRs.

ColdAnkles:
More like pit-opposite. Pits can be seen into from outside, but can't be seen out of.
This you can't see into from outside, but can see out of.

src/main/java/net/rptools/maptool/client/ui/zone/vbl/VisionBlockingAccumulator.java
```diff
@@ -212,4 +212,48 @@ public boolean addPitBlocking(AreaTree topology) {
...
+    /*
+     * There are two cases for Cover VBL:
+     * 1. A token inside Cover VBL can see everything, unobstructed by the cover.
+     * 2. A token outside Cover VBL can see nothing, as if it were wall.
+     */
+
+    if (container instanceof final AreaIsland island) {
+      /*
+       * Since we're in an island, vision is entirely unblocked, and we do nothing.
+       */
```
kwvanderlinde: This case does require some vision blocking from other islands. Any sibilng islands should block vision, as should any parent islands (technically the boundary of the parent ocean). E.g., in this example, the trolls should be covered so the Hero should not be able to see them:
image

src/main/java/net/rptools/maptool/model/Token.java
```diff
@@ -270,6 +270,7 @@ public enum Update {
   private Area vbl;
   private Area hillVbl;
   private Area pitVbl;
+  private Area coverVbl;
```
kwvanderlinde, line 273: Doesn't look like this PR exposes Token Cover VBL in any way (no macro functions, no UI in **Edit Token** dialog). So this can be removed for now and we'll just have Map Cover VBL. Or if you're feeling motivated you can build out the functions / UI for Token Cover VBL, up to you :slightly_smiling_face:.

src/main/java/net/rptools/maptool/client/ui/zone/vbl/VisionBlockingAccumulator.java
kwvanderlinde: And one more case for child islands:
image

src/main/resources/net/rptools/maptool/language/i18n_uk.properties
```diff
@@ -2446,6 +2446,7 @@ tools.ignore_vbl_on_move.tooltip = Tokens will navigate around VBL if selected,
 tools.topology_mode_selection.vbl.tooltip = Draw Wall VBL.
 tools.topology_mode_selection.hill_vbl.tooltip = Draw Hill VBL.
 tools.topology_mode_selection.pit_vbl.tooltip = Draw Pit VBL.
+tools.topology_mode_selection.cover_vbl.tooltip = Draw Cover VBL.
```
kwvanderlinde: This should be placed in `i18n.properties` instead.

kwvanderlinde (changes requested):

src/main/java/net/rptools/maptool/model/Token.java
kwvanderlinde: And I can lend a hand with the UI & functions if that's the route you want to go.

src/main/resources/net/rptools/maptool/language/i18n_uk.properties
ColdAnkles: You'd think I would've noticed that file was there.

src/main/java/net/rptools/maptool/model/Token.java
ColdAnkles: Can't say for sure I got it all, but some of the necessary stuff done.

src/main/java/net/rptools/maptool/client/ui/zone/vbl/VisionBlockingAccumulator.java
ColdAnkles: I think this is all resolved now. Appreciate the examples though.

src/main/java/net/rptools/maptool/client/ui/zone/vbl/VisionBlockingAccumulator.java
```diff
@@ -212,4 +212,61 @@ public boolean addPitBlocking(AreaTree topology) {
...
+     * There are two cases for Cover VBL:
+     * 1. A token inside Cover VBL can see everything, unobstructed by the cover.
+     * 2. A token outside Cover VBL can see nothing, as if it were wall.
+     */
+
+    if (container instanceof final AreaIsland island) {
+
+      final var parentOcean = island.getParentOcean();
+
+      final var grandparentIsland = parentOcean.getParentIsland();
```
kwvanderlinde: No need for the `grandparentIsland` to contribute here (and it won't tend to add segments anyways). The contributions from `parentOcean` already block any vision into the `grandparentIsland`.

ColdAnkles:
Sometimes I really struggle with git and merging upstream changes back into my fork -.-

kwvanderlinde (approved):

kwvanderlinde:
Took this for a spin, there's one edge case that is not working as I would expect: a token in cover will have its vision blocked by the opposite side of an enclosed ocean (hole):
image

Compare with the case where the "hole" is not completely enclosed, then vision is not obstructed at all:
image

I would expect the first case to also not block vision either.

PR #4318, opened 2023-10-03
onChangeMap add-on compliance, output added

### LINKED ISSUE #1574: Add an onMapChange Event

Jmr3366:
Feature #1574 

Initial creation of `onChangeMap` was not add-on compliant.
This also resolves a false/positive event trigger when entering the menu Map - Edit Map dialog - Cancel (the event will still trigger if variables are changed and selecting OK).
### Description of the Change
Compliance, tested with add-ons
Added output `mapID`,`mapName`
### Possible Drawbacks
No additional drawbacks from prior PR

### Release Notes

Any macro named `onChangeMap` located on a library token will execute when the map is changed.
The macro will also receive the `mapID` and `mapName` through `macro.args`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4318)

### DIFF
--- a/src/main/java/net/rptools/maptool/events/ZoneLoadedListener.java
+++ b/src/main/java/net/rptools/maptool/events/ZoneLoadedListener.java
@@ -18,11 +18,19 @@ import static net.rptools.maptool.client.functions.MapFunctions.ON_CHANGE_MAP_CA
 
 import com.google.common.eventbus.Subscribe;
 import java.util.Collections;
+import java.util.concurrent.ExecutionException;
+import net.rptools.maptool.client.MapTool;
 import net.rptools.maptool.client.events.ZoneLoaded;
-import net.rptools.maptool.model.Token;
+import net.rptools.maptool.client.ui.zone.ZoneRenderer;
+import net.rptools.maptool.language.I18N;
+import net.rptools.maptool.model.library.Library;
+import net.rptools.maptool.model.library.LibraryManager;
 import net.rptools.maptool.util.EventMacroUtil;
+import org.apache.log4j.LogManager;
+import org.apache.log4j.Logger;
 
 public class ZoneLoadedListener {
+  private static final Logger LOGGER = LogManager.getLogger(EventMacroUtil.class);
 
   public ZoneLoadedListener() {
     new MapToolEventBus().getMainEventBus().register(this);
@@ -30,12 +38,26 @@ public class ZoneLoadedListener {
 
   @Subscribe
   public void OnChangedMap(ZoneLoaded event) {
-    var libTokens = EventMacroUtil.getEventMacroTokens(ON_CHANGE_MAP_CALLBACK);
-    String prefix = ON_CHANGE_MAP_CALLBACK + "@";
-
-    for (Token handler : libTokens) {
-      EventMacroUtil.callEventHandlerOld(
-          prefix + handler.getName(), "", handler, Collections.emptyMap(), true);
+    ZoneRenderer currentZR = MapTool.getFrame().getCurrentZoneRenderer();
+    try {
+      var libs = new LibraryManager().getLegacyEventTargets(ON_CHANGE_MAP_CALLBACK).get();
+      if (!libs.isEmpty()) {
+        for (Library handler : libs) {
+          try {
+            String libraryNamespace = handler.getNamespace().get();
+            EventMacroUtil.callEventHandler(
+                ON_CHANGE_MAP_CALLBACK,
+                libraryNamespace,
+                currentZR.getZone().getId().toString() + "," + currentZR.getZone().toString(),
+                null,
+                Collections.emptyMap());
+          } catch (InterruptedException | ExecutionException e) {
+            throw new AssertionError("Error retrieving library namespace");
+          }
+        }
+      }
+    } catch (InterruptedException | ExecutionException e) {
+      LOGGER.error(I18N.getText("library.error.retrievingEventHandler"), e);
     }
   }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/net/rptools/maptool/events/ZoneLoadedListener.java
```diff
@@ -14,28 +14,49 @@
...
-    var libTokens = EventMacroUtil.getEventMacroTokens(ON_CHANGE_MAP_CALLBACK);
-    String prefix = ON_CHANGE_MAP_CALLBACK + "@";
-
-    for (Token handler : libTokens) {
-      EventMacroUtil.callEventHandlerOld(
-          prefix + handler.getName(), "", handler, Collections.emptyMap(), true);
+    ZoneRenderer currentZR = MapTool.getFrame().getCurrentZoneRenderer();
+    try {
+      var libs = new LibraryManager().getLegacyEventTargets(ON_CHANGE_MAP_CALLBACK).get();
+      if (!libs.isEmpty()) {
```
thelsing: IMO it's better to exit the method if libs.isEmpty(). Generally it's better to handle the conditions you don't want first. Otherwise they tend to be forgotten. You also get less indent.

src/main/java/net/rptools/maptool/events/ZoneLoadedListener.java
```diff
@@ -14,28 +14,49 @@
...
+          try {
+            String libraryNamespace = handler.getNamespace().get();
+            EventMacroUtil.callEventHandler(
+                ON_CHANGE_MAP_CALLBACK,
+                libraryNamespace,
+                currentZR.getZone().getId().toString() + "," + currentZR.getZone().toString(),
+                null,
+                Collections.emptyMap());
+          } catch (InterruptedException | ExecutionException e) {
+            throw new AssertionError("Error retrieving library namespace");
```
thelsing: You probably want to add the original exception as inner exception or at least log it. Otherwise it will be difficult to determine the reason for the AssertionError if it ever occurs in a log file.

thelsing (changes requested):

thelsing (approved):

src/main/java/net/rptools/maptool/events/ZoneLoadedListener.java
```diff
@@ -14,28 +14,49 @@
...
-    var libTokens = EventMacroUtil.getEventMacroTokens(ON_CHANGE_MAP_CALLBACK);
-    String prefix = ON_CHANGE_MAP_CALLBACK + "@";
-
-    for (Token handler : libTokens) {
-      EventMacroUtil.callEventHandlerOld(
-          prefix + handler.getName(), "", handler, Collections.emptyMap(), true);
+    ZoneRenderer currentZR = MapTool.getFrame().getCurrentZoneRenderer();
+    try {
+      var libs = new LibraryManager().getLegacyEventTargets(ON_CHANGE_MAP_CALLBACK).get();
+      if (libs.isEmpty()) return;
```
kwvanderlinde: Please add braces and indentation to this `if` for readability.

src/main/java/net/rptools/maptool/events/ZoneLoadedListener.java
```diff
@@ -14,28 +14,49 @@
...
+      for (Library handler : libs) {
+        try {
+          String libraryNamespace = handler.getNamespace().get();
+          EventMacroUtil.callEventHandler(
+              ON_CHANGE_MAP_CALLBACK,
+              libraryNamespace,
+              currentZR.getZone().getId().toString() + "," + currentZR.getZone().toString(),
```
kwvanderlinde: I think we should remove the second argument as it's not needed - if the event handler does need it, it can get it based on the ID.

The result of `Zone#toString()` is also not very predictable: depending on whether it's a GM or player client, and whether the name and display name match, there could be several possible values. So if we keep the parameter, we should pick either `getName()` or `getDisplayName()` for it.

kwvanderlinde (changes requested):

kwvanderlinde:
> This also resolves a false/positive event trigger when entering the menu Map - Edit Map dialog - Cancel (the event will still trigger if variables are changed and selecting OK).

@Jmr3366 I don't see anything in this PR that will address that point. Is it just the addition of parameters to `onChangeMap`, or was there meant to be more to this PR?

src/main/java/net/rptools/maptool/events/ZoneLoadedListener.java
Azhrei: I glanced at the code and don't see an answer — why is the zone name being used instead of the id in the first place?  If it's just informational, why not just the zone id and let the listener query other attributes that it wants?

Jmr3366: I assumingly thought there would be added value to include the _textual name_ as it was readily available.
I will remove it.

Jmr3366:
@kwvanderlinde  I'm not certain what was causing the edit map-cancel event trigger issue.  When I found that issue, I was also requested to test and make it add-on compliant.  Once _converted_ the issue I described was gone.

That said, this function is victim to how a rename of a map is really a copy, paste, delete (thus causing this event to fire).  This update does not change that.

kwvanderlinde (approved):

kwvanderlinde (approved):

PR #4319, opened 2023-10-04
onMouseOver event

Jmr3366:
#720 - onMouseOver event request 
### Description of the Change
When the pointer moves over a token on the token layer, any library token with a macro named `onMouseOver` will execute.
### Possible Drawbacks
For any pre-existing library tokens with a macro named `onMouseOver`, this could cause unintended macro execution and/or unexpected results.
### Documentation Notes
Any time the mouse cursor moves over a token, the `onMouseOver` macro will execute if found on a library token.
### Release Notes
Examples:

Create a macro on an existing or new library token, named `onMouseOver`.
Within this macro place:
`[h: broadcast(macro.args,"self")]`

When you move the mouse cursor over a token this will be output to chat 
`tokenid`,`token pixel X`,`token pixel Y`,`shift key state`,`control key state` 
When you move the mouse cursor OFF a token this will output to chat
`tokenid`,`exit`

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4319)


thelsing (changes requested):
The comments for #4318 apply here to.

thelsing (approved):

PR #4325, opened 2023-10-17
Fix for #4324

### LINKED ISSUE #4324: [Bug]: Map-Export Map- can't cancel
### Describe the Bug

If you attempt to export a map, you can't cancel out of the menu and are forced to save the file.
You can choose to overwrite a file and will be prompted cancel after you've selected save

### To Reproduce

Click Map
Select Export Map

Click cancel

### Expected Behaviour

Clicking cancel will close the dialog

### Screenshots

na

### MapTool Info

v1.14.0rc2

### Desktop

Win11, Ubuntu 23.04

### Additional Context

_No response_

Jmr3366:
resolves #4324
### Description of the Change
added a "return"
### Possible Drawbacks
none _(that I found)_

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4325)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppActions.java
+++ b/src/main/java/net/rptools/maptool/client/AppActions.java
@@ -2848,30 +2848,27 @@ public class AppActions {
           chooser.setSelectedFile(new File(zr.getZone().getName()));
           boolean tryAgain = true;
           while (tryAgain) {
-            if (chooser.showSaveDialog(MapTool.getFrame()) == JFileChooser.APPROVE_OPTION) {
-              File mapFile = chooser.getSelectedFile();
-              var installDir = AppUtil.getInstallDirectory().toAbsolutePath();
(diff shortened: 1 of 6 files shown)


src/main/java/net/rptools/maptool/client/AppActions.java
```diff
@@ -2870,6 +2870,8 @@ protected void executeAction() {
                   MapTool.showError("msg.error.failedSaveMap", ioe);
                 }
               }
+            } else {
```
thelsing: I think it would be better to just change "if (chooser.showSaveDialog(MapTool.getFrame()) == JFileChooser.APPROVE_OPTION) { do save }" to "if (chooser.showSaveDialog(MapTool.getFrame()) != JFileChooser.APPROVE_OPTION) break;" instead.

Jmr3366: suggestions completed

Issue #3087, opened 2021-10-20
Add 'Implicit Lights'


wolph42:
**Is your feature request related to a problem? Please describe.**
The problem is that EVERY light source needs to be defined in the campaign properties. This has the following issues:
- hard to transfer to another campaign
- always visible to players (e.g. the source 'glowing dragon eyes' might be something you  don't want to show
- light sources are not 'self contained' (like tokens)

**Describe the solution you'd like**
Add the option to the token editor to set an 'implicit lightsource'. That is a text field where you can enter e.g.:
Dragon Eyes : cone arc=70 30.5 122.5#000000
this makes it easy to create specific lights AND to export import specific lights AND export import maps without worrying about the light settings in the campaign properties.

**Describe alternatives you've considered**
none

**Additional context**
suggested spot for the lightbox: 

<img width="460" alt="light" src="https://user-images.githubusercontent.com/19568714/138168397-7b7f1e6d-aa79-49c6-b0e5-9e0e0be830c8.png">

Phergus:
See #331.

melek:
Would this make it possible for lights imported via the the Universal VTT format to work out of the box, at least in a limited way?

FullBleed:
Totally for this.

BlackwingNL:
All 3 goals listed by @Phergus will need the Light itself to be implemented first. To achieve that the current Light needs to be adapted with (as far as I can see):

1. The type distinction between a regular light and this new type light
2. A location where to store these lights.

As for 2: As I understand it, tokens don't actually contain a light object, but only a reference (GUID) to a light object in the central list of light sources. These light objects only exist in a campaign, not in an exported map or a saved token (!). This significantly limits the portability of lights through maps of tokens but that's a different issue.
So for these implicit lights a light object has to be created and stored somewhere so it can be referred to by the token in the same way and current functions that interact with lights need to be able to find these implicit lights as well.

One option would be to still store them in the campaign with the other lights, just hide them in relevant places (in de campaign properties dialog, and the light menu when the token does not contain that specific light). Outside of the UI this would mean the rest would be pretty much transparent for the light functions.

A second option is to store them on the token that also has the reference to it. This would mean a bit more work for the light functions but makes these lights more modular and able to be saved with a map or token.

Phergus:
Tokens actually have a list of AttachedLightSource objects and each of those have a GUID reference to a campaign-defined LightSource object that contains a list of Lights that make up the LightSource.

[...]

Issue #331, opened 2019-03-17
Enhancement:  Macro functions to dynamically create light source and manage same


Phergus:
Creating new light sources in Campaign Properties for environmental lights that are only used once or can vary in size/color/shape is both tedious and results in a cluttered Light menu.

Being able to create lights dynamically via macro would make it much easier to handle situations like light spells in games where the size and shape are chosen at the time of casting or environmental effects such as when someone sets a large pile of lumber on fire.

Dynamic light sources would _not_ be added to the campaign properties upon creation. (Optional?)

They would be preserved in the saved campaign such that tokens/objects/stamps would carry them through campaign save/load operations.

Suggested functions:
- createDynamicLight("name", "definition string" [, "tokenid"])
- getDynamicLights(["tokenid" [, "delim"]])
- removeDyanamicLight("name"|"all")

Other possible functions
- modifyDynamicLight("existing name", "new definition") - or maybe create on an existing dynamic light just overwrites?

Existing Light functions
- setLight("Dynamic", "name", 0|1) - passed a type of "Dynamic" and a dynamic light name applies that to current token
- getLights() - works as normal with Dynamic lights having type "Dynamic"

Begin discussion.

Azhrei:
Overall, I like this idea. :)

> Being able to create lights dynamically via macro would make it much easier to handle situations like light spells in games where the size and shape are chosen at the time of casting or environmental effects such as when someone sets a large pile of lumber on fire.
> 
> Dynamic light sources would _not_ be added to the campaign properties upon creation. (Optional?)
> 
> They would be preserved in the saved campaign such that tokens/objects/stamps would carry them through campaign save/load operations.

Are they preserved only when there's a token that has them turned on, so that if no tokens are using them, they would be purged from the campaign?

> Suggested functions:
> 
> * createDynamicLight("name", "definition string" [, "tokenid"])
> * getDynamicLights(["tokenid" [, "delim"]])
> * removeDynamicLight("name"|"all")

Are strings the way to go in defining a light?  There's an obvious connection between using a given string in the campaign properties and the same string in a macro, but wouldn't a JSON object of properties make more sense?

> Other possible functions
> 
> * modifyDynamicLight("existing name", "new definition") - or maybe create on an existing dynamic light just overwrites?
> 
> Existing Light functions
> 
> * setLight("Dynamic", "name", 0|1) - passed a type of "Dynamic" and a dynamic light name applies that to current token
[...]

PR #4356, opened 2023-11-01
Add unique light sources on tokens, and macros for manipulating them.

kwvanderlinde:

Implements #331 and so also relates to #3087.

### Description of the Change

This changes adds the concept of "unique light sources", which are lights sources defined on tokens rather than in the campaign as a whole. A token with such a unique light source can equipped or unequip it just like any other light sources in its right-click menu.

Macro functions are provided for defining and manipulating unique light sources. No UI is provided at this time for defining the lights (doing so would round out #3087).

#### Existing macro function changes

The following existing macro functions have been updated:
- `getLights([category, [delim, [token, [map]]]])`
- `setLight(category, name, state, [token, [map]])`
- `hasLightSource(category, name, [token, [map]])`

The `category` parameter of each can now be set to the special `"$unique"` category to identify unique light sources. The wildcard `"*"` will likewise now include the token's unique light sources. I.e., these function treat the token's unique light sources as though they just belong to a specially named category.

#### New macro functions

The following new macro functions have been added:
- `createUniqueLightSource(lightSourceJson, [token, [map]])` to create a new unique light source.
- `updateUniqueLightSource(lightSourceJson, [token, [map]])` to modify a unique light source with the matching name.
- `deleteUniqueLightSource(name, [token, [map]])` to delete a unique light source by name.
- `getUniqueLightSource(name, [token, [map]])` to get a unique light source by name, as a JSON object.
- `getUniqueLightSources([token, [map]])` to get all of a token's unique light sources as a JSON array of JSON objects.
- `getUniqueLightSourceNames([token, [map]])` to get the names of all of a token's light sources, as a JSON array.

A consistent JSON format is used for light sources in the parameters and return values of these functions, and uses terminology from the **Campaign Properties** dialog. It looks like this:
```json
{
	"name": "Torch - 20",
	"type": "NORMAL",
	"scale": false,
	"lights": [
		{"shape": "CIRCLE", "range": 20, "lumens": 100},
		{"shape": "CIRCLE", "range": 40, "color": "#000000", "lumens": 50}
	]
}
```

The exact fields are:
- `"name"`: `string`; required.
- `"type"`: (`"NORMAL"`|`"AURA"`); defaults to `"NORMAL"`.
- `"scale"`: `boolean`; defaults to `false`.
- `"lights"` `array`; defauls to `[]`.

And for each light:
- `"range"`: `number`, required.
- `"shape"`: (`"SQUARE"`|`"CIRCLE"`|`"CONE"`|`"HEX"`|`"GRID"`), defaults to `"CIRCLE"`.
- `"offset"`: `number`, defaults to `0`. Only permitted if the shape is `"CONE"`.
- `"arc"`: `number`, defaults to `0`. Only permitted if the shape is `"CONE"`.
- `"gmOnly"`: `boolean`, defaults to `false`. Only permitted if the light source type is `"AURA"`.
- `"ownerOnly"`: `boolean`, defaults to `false`. Only permitted if the light source type is `"AURA"`.
- `"color"`: `string` of the form `"#rrggbb`, defaults to `null`.
- `"lumens"`: `number` (integer), defaults to 100.

#### Refactoring

There was some refactoring done as groundwork.

`LightSource` is now immutable to avoid worries about cloning them along with tokens. In practice lights weren't been modified anyways - only `CampaignPropertiesDialog` and `LightSourceCreator` used the mutators, and that was only to build the light. Once built, the `LightSource` instances were never being modifed.

The existing token update messages for light sources (e.g., for attaching to a token) were reduced to passing around the light source ID instead of the entire light source definition. The ID was the only thing ever used, so there was no point serializing everything for one GUID. Now the only time a complete light source is sent in a token update is the new case of creating a unique light source.

`AttachedLightSource` has been encapsulated more and made responsible for the common case of looking up light sources by ID. Callers just need to provide the current campaign (and now also the associated token) and the `AttachedLightSource` will look itself up and return a `LightSource`.

### Possible Drawbacks

There were some changes to the model. Should be benign, but need to watch for deserialization failures on existing campaigns.

If someone is already using `$unique` as a light source category, it won't be possible to find them with the macro functions  now.

### Documentation Notes

For the existing functions `getLights()`, `setLight()`, and `hasLightSource()`, the type parameter (which should be called "category") can now be set to `"$unique"` to access unique light sources on the token.

#### createUniqueLightSource() function

_Trusted function_

Creates a new unique light source on a token.

**Usage**
```
createUniqueLightSource(lightSourceJson)
createUniqueLightSource(lightSourceJson, token)
createUniqueLightSource(lightSourceJson, token, map)
```

**Parameters**
- `lightSourceJson` - a JSON object defining the new unique light source. See below for details.
- `token` - the name or ID of the token to add the unique light source to. Defaults to the current token.
- `map` - the name or ID of the map on which to find `token`. Defaults to the current map.

The `lightSourceJson` is a JSON object with these fields:
- `"name"`: The user-visible name of the unique light source. A unique light source with the same name must not already exist on the token. Required field.
- `"type"`: The type of the unique light source. Either `"NORMAL"` or `"AURA"`; defaults to `"NORMAL"`.
- `"scale"`: Whether to add the token footprint (size) to the range, so that the light starts at the token's edge vs the token's center. Either `json.true` or `json.false`; defaults to `json.false`.
- `"lights"`: A JSON array of lights; defauls to `[]`.

Each light has this JSON format:
- `"range"`: The distance that the light extends, in map units. Required field.
- `"shape"`: The shape of the boundary of the light. Must be one of `"SQUARE"`, `"CIRCLE"`, `"CONE"`, `"HEX"`, or `"GRID"`; defaults to `"CIRCLE"`.
- `"offset"`: (_for cones only_) The counterclockwise angle (in degrees) to shift the cone relative to the token's facing. Defaults to `0`.
- `"arc"`: (_for cones only_) The angle of the cone in degrees. Defaults to `0`.
- `"gmOnly"`: (_for auras only_) Whether the aura should only be shown to GMs. Either `json.true` or `json.false`; defaults to `json.false`.
- `"ownerOnly"`: (_for auras only_) Whether the aura should only be shown to the tokens owners. Either `json.true` or `json.false`; defaults to `json.false`.
- `"color"`: The color of the light, in the form `"#rrggbb"`; defaults to no color (transparent).
- `"lumens"`: How bright the light should be treated, with negative values indicating darkness. Defaults to 100.

**Return value**

The name of the new unique light source.

**Examples**

To create a standard D&D torch, but with a blue flame:
```
createUniqueLightSource(json.set("{}",
	"name", "Blue Flame Torch",
	"type", "NORMAL",
	"scale", false,
	"lights", json.append("[]",
		json.set("{}", "shape", "CIRCLE", "range", 20, "color", "#0000ff", "lumens", 100),
		json.set("{}", "shape", "CIRCLE", "range", 40, "color", "#000077", "lumens", 50)
	)
))
```

And the same example, but using defaults where possible:
```
createUniqueLightSource(json.set("{}",
	"name", "Blue Flame Torch",
	"lights", json.append("[]",
		json.set("{}", "range", 20, "color", "#0000ff"),
		json.set("{}", "range", 40, "color", "#000077", "lumens", 50)
	)
))
```

#### updateUniqueLightSource() function

_Trusted function_

Modifies a unique light source on a token.

**Usage**
```
updateUniqueLightSource(lightSourceJson)
updateUniqueLightSource(lightSourceJson, token)
updateUniqueLightSource(lightSourceJson, token, map)
```

**Parameters**

- `lightSourceJson` - a JSON object defining the new unique light source. The format is the same as for `createUniqueLightSource()`.
- `token` - the name or ID of the token to to find the unique light source on. Defaults to the current token.
- `map` - the name or ID of the map on which to find `token`. Defaults to the current map.

**Return value**

The name of the modified unique light source.

#### deleteUniqueLightSource() function

Removes a unique light source from a token.

Note: if you merely want to unequip or turn off the unique light source, but keep it around for later, use `setLight()` instead.

**Usage**
```
deleteUniqueLightSource(name)
deleteUniqueLightSource(name, token)
deleteUniqueLightSource(name, token, map)
```

**Parameters**
- `name` - the name of the unique light source to delete.
- `token` - the name or ID of the token to delete the unique light source from. Defaults to the current token. _This parameter can only be used in a trusted macro._
- `map` - the name or ID of the map on which to find `token`. Defaults to the current map.

#### getUniqueLightSource() function

Looks up a unique light source by name on a token, returning the 

**Usage**
```
getUniqueLightSource(name)
getUniqueLightSource(name, token)
getUniqueLightSource(name, token, map)
```

**Parameters**
- `name` - the name of the unique light source to delete.
- `token` - the name or ID of the token to find the unique light source on. Defaults to the current token. _This parameter can only be used in a trusted macro._
- `map` - the name or ID of the map on which to find `token`. Defaults to the current map.

**Return value**

The matching unique light source, as a JSON object. See `createUniqueLightSource()` for the JSON format of a light source.

If there is no match unique light source, an empty string is returned.

**Examples**

```
[h: createUniqueLightSource(json.set("{}",
	"name", "Simple Torch",
	"lights", json.append("[]", json.set("{}", "range", 40))
))]

[r: getUniqueLightSource("Simple Torch")]
```
This will print `{"name":"Simple Torch","type":"NORMAL","scale":false,"lights":[{"shape":"CIRCLE","range":40.0,"lumens":100}]}`

#### getUniqueLightSources() function

Gets all the unique light sources defined on a token.

**Usage**
```
getUniqueLightSources()
getUniqueLightSources(token)
getUniqueLightSources(token, map)
```

**Parameters**
- `token` - the name or ID of the token to get the unique light sources from. Defaults to the current token.  _This parameter can only be used in a trusted macro._
- `map` - the name or ID of the map on which to find `token`. Defaults to the current map.

** Return value**

A JSON array of all the token's unique light sources. As if `getUniqueLightSource()` was call for each unique light source on the token and the results put in an array.

#### getUniqueLightSourceNames() function

Get the names of all the unique light sources defined on a token.

**Usage**
```
getUniqueLightSourceNames()
getUniqueLightSourceNames(token)
getUniqueLightSourceNames(token, map)
```

**Parameters**
- `token` - the name or ID of the token to get the unique light source from. Defaults to the current token.  _This parameter can only be used in a trusted macro._
- `map` - the name or ID of the map on which to find `token`. Defaults to the current map.

** Return value**

A JSON array of all the names of the token's unique light sources.

### Release Notes

- Added macro functions for creating and manipulating unique light sources on tokens.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4356)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/AbstractTokenPopupMenu.java
+++ b/src/main/java/net/rptools/maptool/client/ui/AbstractTokenPopupMenu.java
@@ -22,7 +22,7 @@ import java.awt.event.ActionEvent;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -143,17 +143,46 @@ public abstract class AbstractTokenPopupMenu extends JPopupMenu {
       }
       menu.addSeparator();
     }
+
+    // Add unique light sources for the token.
+    {
+      JMenu subMenu = new JMenu("Unique");
+
+      List<LightSource> lightSources = new ArrayList<>(tokenUnderMouse.getUniqueLightSources());
+      Collections.sort(lightSources);
+
+      LIGHTSOURCES:
+      for (LightSource lightSource : lightSources) {
+        for (Light light : lightSource.getLightList()) {
+          // TODO Shouldn't we only skip if *all* child lights are GM only. And what about owner
+          //  only?
+          if (light.isGM() && !MapTool.getPlayer().isGM()) {
+            continue LIGHTSOURCES;
+          }
+        }
+        JCheckBoxMenuItem menuItem =
+            new JCheckBoxMenuItem(new ToggleLightSourceAction(lightSource));
+        menuItem.setSelected(tokenUnderMouse.hasLightSource(lightSource));
+        subMenu.add(menuItem);
+      }
+      if (subMenu.getItemCount() != 0) {
+        menu.add(subMenu);
+        menu.addSeparator();
+      }
+    }
+
     for (Entry<String, Map<GUID, LightSource>> entry :
         MapTool.getCampaign().getLightSourcesMap().entrySet()) {
       JMenu subMenu = new JMenu(entry.getKey());
 
-      List<LightSource> lightSources = new ArrayList<LightSource>(entry.getValue().values());
-      LightSource[] lightSourceList = new LightSource[entry.getValue().size()];
-      lightSources.toArray(lightSourceList);
-      Arrays.sort(lightSourceList);
+      List<LightSource> lightSources = new ArrayList<>(entry.getValue().values());
+      Collections.sort(lightSources);
+
       LIGHTSOURCES:
-      for (LightSource lightSource : lightSourceList) {
+      for (LightSource lightSource : lightSources) {
         for (Light light : lightSource.getLightList()) {
+          // TODO Shouldn't we only skip if *all* child lights are GM only. And what about owner
+          //  only?
           if (light.isGM() && !MapTool.getPlayer().isGM()) {
             continue LIGHTSOURCES;
           }
--- a/src/main/java/net/rptools/maptool/client/ui/AbstractTokenPopupMenu.java
+++ b/src/main/java/net/rptools/maptool/client/ui/AbstractTokenPopupMenu.java
@@ -466,9 +466,9 @@ public abstract class AbstractTokenPopupMenu extends JPopupMenu {
           continue;
         }
         if (token.hasLightSource(lightSource)) {
-          token.removeLightSource(lightSource);
+          token.removeLightSource(lightSource.getId());
         } else {
-          token.addLightSource(lightSource);
+          token.addLightSource(lightSource.getId());
         }
         MapTool.serverCommand().putToken(renderer.getZone().getId(), token);
 
--- a/src/main/java/net/rptools/maptool/client/ui/AbstractTokenPopupMenu.java
+++ b/src/main/java/net/rptools/maptool/client/ui/AbstractTokenPopupMenu.java
@@ -22,6 +22,7 @@ import java.awt.event.ActionEvent;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -146,25 +147,7 @@ public abstract class AbstractTokenPopupMenu extends JPopupMenu {
 
     // Add unique light sources for the token.
     {
-      JMenu subMenu = new JMenu("Unique");
-
-      List<LightSource> lightSources = new ArrayList<>(tokenUnderMouse.getUniqueLightSources());
-      Collections.sort(lightSources);
-
-      LIGHTSOURCES:
-      for (LightSource lightSource : lightSources) {
-        for (Light light : lightSource.getLightList()) {
-          // TODO Shouldn't we only skip if *all* child lights are GM only. And what about owner
-          //  only?
-          if (light.isGM() && !MapTool.getPlayer().isGM()) {
-            continue LIGHTSOURCES;
-          }
-        }
-        JCheckBoxMenuItem menuItem =
-            new JCheckBoxMenuItem(new ToggleLightSourceAction(lightSource));
-        menuItem.setSelected(tokenUnderMouse.hasLightSource(lightSource));
-        subMenu.add(menuItem);
-      }
+      JMenu subMenu = createLightCategoryMenu("Unique", tokenUnderMouse.getUniqueLightSources());
       if (subMenu.getItemCount() != 0) {
         menu.add(subMenu);
         menu.addSeparator();
@@ -173,30 +156,34 @@ public abstract class AbstractTokenPopupMenu extends JPopupMenu {
 
     for (Entry<String, Map<GUID, LightSource>> entry :
         MapTool.getCampaign().getLightSourcesMap().entrySet()) {
-      JMenu subMenu = new JMenu(entry.getKey());
-
-      List<LightSource> lightSources = new ArrayList<>(entry.getValue().values());
-      Collections.sort(lightSources);
-
-      LIGHTSOURCES:
-      for (LightSource lightSource : lightSources) {
-        for (Light light : lightSource.getLightList()) {
-          // TODO Shouldn't we only skip if *all* child lights are GM only. And what about owner
-          //  only?
-          if (light.isGM() && !MapTool.getPlayer().isGM()) {
-            continue LIGHTSOURCES;
-          }
-        }
-        JCheckBoxMenuItem menuItem =
-            new JCheckBoxMenuItem(new ToggleLightSourceAction(lightSource));
-        menuItem.setSelected(tokenUnderMouse.hasLightSource(lightSource));
-        subMenu.add(menuItem);
-      }
+      JMenu subMenu = createLightCategoryMenu(entry.getKey(), entry.getValue().values());
       if (subMenu.getItemCount() != 0) {
         menu.add(subMenu);
       }
     }
     return menu;
+  }
+
+  protected JMenu createLightCategoryMenu(String categoryName, Collection<LightSource> sources) {
+    JMenu subMenu = new JMenu(categoryName);
+
+    List<LightSource> lightSources = new ArrayList<>(sources);
+    Collections.sort(lightSources);
+
+    for (LightSource lightSource : lightSources) {
+      // Don't include light sources that don't have lights visible to the player. Note that the
+      // player must be an owner to use the popup, so don't bother checking `::isOwner()`.
+      boolean include =
+          MapTool.getPlayer().isGM() || !lightSource.getLightList().stream().allMatch(Light::isGM);
+      if (include) {
+        JCheckBoxMenuItem menuItem =
+            new JCheckBoxMenuItem(new ToggleLightSourceAction(lightSource));
+        menuItem.setSelected(tokenUnderMouse.hasLightSource(lightSource));
+        subMenu.add(menuItem);
+      }
+    }
+
+    return subMenu;
   }
 
   protected Token getTokenUnderMouse() {
--- a/src/main/java/net/rptools/maptool/model/LightSource.java
+++ b/src/main/java/net/rptools/maptool/model/LightSource.java
@@ -14,6 +14,7 @@
  */
 package net.rptools.maptool.model;
 
+import com.google.common.collect.ImmutableList;
 import com.google.protobuf.StringValue;
 import java.awt.geom.Area;
 import java.io.IOException;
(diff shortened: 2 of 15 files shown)


src/main/proto/data_transfer_objects.proto
```diff
@@ -641,16 +642,18 @@ enum TokenUpdateDto {
...
-    removeLightSource = 47;
-    addLightSource = 48;
-    setHasSight = 49;
-    setSightType = 50;
-    flipX = 51;
-    flipY = 52;
-    flipIso = 53;
-    setSpeechName = 54;
-    removeFacing = 55;
+    createUniqueLightSource = 46;
```
thelsing, line 648: Just for discussion: do we want to keep related stuff together and break compatibility with (probably not existing) other tools, or do we want to keep the protobuff stuff backwards compatible. Currently it does't really matter, but when we use it for saving campaigns we need to keep compatibility. I have not strong opinion about this.

src/main/java/net/rptools/maptool/tool/LightSourceCreator.java
```diff
@@ -57,28 +58,33 @@ public static void main(String[] args) {
...
-            false));
-
-    return source;
+    return LightSource.createRegular(
+        name,
+        new GUID(),
+        LightSource.Type.NORMAL,
+        false,
+        List.of(
+            // new Light(0, 5, arcAngle, new DrawableColorPaint(new Color(255, 255, 0, 50)))
```
thelsing: I think we can remove this comment when updating this.

src/main/java/net/rptools/maptool/model/Token.java
```diff
@@ -916,14 +922,30 @@ public String getImageTableName() {
...
+  public void removeUniqueLightSource(GUID lightSourceId) {
+    uniqueLightSources.remove(lightSourceId);
+  }
+
+  public void addLightSource(GUID lightSourceId) {
+    lightSourceList.add(new AttachedLightSource(lightSourceId));
   }
 
   public void removeLightSourceType(LightSource.Type lightType) {
     for (ListIterator<AttachedLightSource> i = lightSourceList.listIterator(); i.hasNext(); ) {
```
thelsing, line 948: We might es well refactor those loops to use foreach syntax IMO.

src/main/java/net/rptools/maptool/model/LightSource.java
```diff
@@ -156,37 +173,17 @@ public void setId(@Nonnull GUID id) {
...
-    lightList.remove(source);
-  }
-
   /**
-   * @return the lights belonging to this LightSource.
+   * @return A read-only list of lights belonging to this LightSource
    */
   public @Nonnull List<Light> getLightList() {
-    return Collections.unmodifiableList(lightList);
+    return lightList;
```
thelsing, line 194: Can we change the type if lightList to ImmutableList? This way it would be more obvious that the list is read only. I can already see somebody adding a constructor that sets the field to 
ArrayList coming.

src/main/java/net/rptools/maptool/client/ui/AbstractTokenPopupMenu.java
```diff
@@ -143,17 +143,46 @@ protected JMenu createLightSourceMenu() {
...
     }
+
+    // Add unique light sources for the token.
+    {
+      JMenu subMenu = new JMenu("Unique");
+
+      List<LightSource> lightSources = new ArrayList<>(tokenUnderMouse.getUniqueLightSources());
+      Collections.sort(lightSources);
+
+      LIGHTSOURCES:
```
thelsing: This may be just my personal opinion but continue label looks bad for me. (To much like goto.) How about you add an LightSource.IsGM() and use that here?

src/main/java/net/rptools/maptool/client/ui/AbstractTokenPopupMenu.java
```diff
@@ -143,17 +143,46 @@ protected JMenu createLightSourceMenu() {
...
-      LightSource[] lightSourceList = new LightSource[entry.getValue().size()];
-      lightSources.toArray(lightSourceList);
-      Arrays.sort(lightSourceList);
+      List<LightSource> lightSources = new ArrayList<>(entry.getValue().values());
+      Collections.sort(lightSources);
+
       LIGHTSOURCES:
-      for (LightSource lightSource : lightSourceList) {
+      for (LightSource lightSource : lightSources) {
         for (Light light : lightSource.getLightList()) {
```
thelsing: Same as above.

thelsing (commented):
I didn't look much at how macros work so I can't say much about TokenLightFunctions but it looks good. 
For the other stuff I added some comments.

src/main/proto/data_transfer_objects.proto
kwvanderlinde: Because backwards compatibility isn't an issue (client + server always have to match anyways) I'd rather make it read nicely. If/when BC concerns change in the future, we'll of course be more particular.

And if there is some 3rd party tool that also consumes our protobufs, it would be that author's responsibility to keep things up-to-date until we commit to a stable protocol.

src/main/java/net/rptools/maptool/model/Token.java
kwvanderlinde: foreach is more unwieldy here since it doesn't support removal. We would have to build up a list of things to remove, then do the removal separately.

src/main/java/net/rptools/maptool/model/LightSource.java
kwvanderlinde: I started with that actually, but had to backtrack. Serialization is our enemy here.

First, modern campaigns will serialize `lightList` in the plainest XML possible (which is great), but older ones, and also our current default light XML, might specify the implementation (such as `LinkedList`). In this case, it's the difference between `<lightList>` and `<lightList class="linked-list">`. If we changed the field type, anything with the latter wouldn't be able to deserialize due to a type mismatch.

Second, XStream only has [special converters](https://x-stream.github.io/converters.html#java.util) for `ArrayList` and `LinkedList`, not for `List` in general. So if we serialize an `ImmutableList`, it will include the implementation details. That will lock us into supporting this 3rd party `List` implementation, and we would have to hope their internals never change in the wrong way.

Both these should are solvable, e.g., by writing custom converters. But it's not trivial and every time I look at them there turns out to be more subtleties than I expect.

src/main/java/net/rptools/maptool/client/ui/AbstractTokenPopupMenu.java
kwvanderlinde: Yeah, I should rewrite these loops. The menu wasn't a focus for me, so I didn't want to change the behaviour, but these are just not correct as they are.

E.g., consider a partially GM-only aura: `Test: aura circle 20#ff0000 GM 40#00ff00`. This shouldn't be excluded from the menu.

thelsing (approved):

src/main/java/net/rptools/maptool/model/LightSource.java
thelsing: Can you add a comment to lightList for that?

kwvanderlinde:
Resolved the conflicts with the Cover VBL addition.

thelsing (approved):

bubblobill:
Umm.. hello? Can we undo this please? You know, until it works and I can edit tokens and stuff.

PR #4374, opened 2023-11-04
Revert "Add unique light sources on tokens, and macros for manipulating them."

bubblobill:
Reverts RPTools/maptool#4356

Cannot open edit token dialog and a few other things seem a bit squirly.
Please take this back

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4374)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ServerCommandClientImpl.java
+++ b/src/main/java/net/rptools/maptool/client/ServerCommandClientImpl.java
@@ -674,16 +674,18 @@ public class ServerCommandClientImpl implements ServerCommand {
 
   @Override
   public void updateTokenProperty(Token token, Token.Update update, LightSource value) {
-    if (update == Token.Update.createUniqueLightSource) {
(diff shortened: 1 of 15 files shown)


kwvanderlinde:
Which other things?

The problem with the Edit Token dialog is not due to the change you're reverting.

kwvanderlinde:
#4375 fixes the dialog

bubblobill:
hopefully other changes to the token will also work. Had trouble with things like setting halos.
assuming it was something to do with token data
didn't dig deep, just noticed I couldn't play with the bit I wanted

kwvanderlinde:
For something like that it's possible you need to refresh your protobuf codegen.

kwvanderlinde:
@bubblobill Can you confirm whether things work for you now on the latest `develop`?

bubblobill:
Just had a quick test. The bits I was playing with previously are behaving in the current develop

PR #4402, opened 2023-11-16
Generalize the use of layers using new properties and standard enum functionality

### LINKED ISSUE #4378: [Refactoring]: Generalize layers internally
### Describe the problem

There's a lot of places in the codebase where we depend on specific layers to modify functionality. This makes it difficult to add new layers or think about other options for layering.

I would like to generalize layer functionality by defining a set of concrete properties that each layer provides. Functionality that today depends on specific layers would be rewritten to operate on any layers with matching properties. Also, functionality that today enumerates all layers by name would be rewritten to enumerate them another way (e.g., via `Layer.values()`).

I've identified the following properties for layers (all boolean):
1. Is it visible to players? _Currently `layer != GM`_
2. Is it a token layer? _Currently `layer == TOKEN`_
3. Is it a stamp layer _Currently `layer != TOKEN`_
4. Should STG tokens be snapped according to their center? _Currently `layer != BACKGROUND`, otherwise snapping by top-left corner._
5. Should token "drag"" using arrow keys be completed each step? _Currently `layer == BACKGROUND`_

Also a couple defaults that can be set in a common place rather than explicilty mentioned the layer everywhere:
1. `TOKEN` is a default/fallback layer for lots of player-oriented things.
2. `BACKGROUND` is the fallback layer for drawings.

### Expected Benefits

Layers will be more general, making it easier to define new layers without having to make code changes in a million places.

### Additional Context

The mentioned properties only _mostly_ define layer behaviour. There are a couple other bits that I can't see how to generalize right now, at least not in a satisfying way.

1. The `ZoneRenderer` has some very particular needs with layers, especially interleaving layers with each other (Hidden and Token) and other elements (Object and the grid).
2. When changing a token's layer ...
    - to Object or Background: non-figure tokens are changed to Top-down.
    - to Token: we try to guess a good shape, usually leaving Figures alone.
    - to GM: do nothing
    - BUT if we do it via macro functions instead of the UI, we don't treat Figures specially, and the "to GM" case is the same as "to Token".
3. The `getToken()` function defaults to using the Token and Hidden layer which otherwise have almost no properties in common. So how do we - and should we - generalize that?

kwvanderlinde:

Resolves #4378

### Description of the Change

This PR could be summarized by three main changes.

The primary change is the introduction of properties associated with each token that can be queried to decide on application behaviour. Rather than comparing against specific layers, we can instead questions of the layer to tell us what behaviour it should have. These are all the new layer properties that exist:
- `isTokenLayer()`: indicates that tokens on this layer are conceptually tokens (`TOKEN`).
- `isStampLayer()` indicates that tokens on this layer are conceptually stamps (`GM`, `OBJECT`, `BACKGROUND`).
- `supportsWalker()` indicates that pathfinding, including naive movement, is permitted (`TOKEN`).
- `anchorSnapToGridAtCenter()` indicates the resizing operations for snap-to-grid tokens should keep the token's center fixed, otherwise the top-left corner will remain fixed (`TOKEN`, `GM`, `OBJECT`).
- `oneStepKeyDrag()` indicates that movement with the arrow keys should immediate "drop" the token (`BACKGROUND`).
- `supportsGuessingTokenShape()` indicates that tokens added to or moved to the layer can have their shape automatically guessed from the token's image via the new `Token.guessAndSetShape()` (`TOKEN`).
- `isVisibleToPlayers()` indicates that the layer should be rendered in player views (`TOKEN`, `OBJECT`, `BACKGROUND`).
- `isPlayerLayer()` indicates that players can have this layer as their active layer (`TOKEN`).
- `isMarkerLayer()` indicates that tokens on this layer can have their notes and portrait displayed as a marker (`GM`, `OBJECT`, `BACKGROUND`).

None of this is exposed to users, it is purely an internal representation.

A second change is to avoid explicit enumeration of layers in various cases. Some of these are obvious, where lists of all `Layer` constants are being created instead of using `.values()`. Others are sneakier, such as the existence of distinct `Zone` methods for looking up tokens on each layer, and separate fields for drawings on each layer. Such methods and fields have been coalesced in a way that a `Layer` can be provided as a key to look up the approriate values. In a similar vein, the `Token` methods for checking layers have been removed in favour of using the new layer properties, or in the worst case, in favour of equality checks against `Layer` constants.

Finally, some work was done on those GUI elements that depend on layers. The most extreme of these is `TokenPanelTreeModel` that has been largely reworked to be simplified in the presence of the new layer attributes. Also the `DrawPanelTreeModel` will display any and all layers automatically. `ExportDialog` is in an midway state - the class logic can handle any layers, but the `.form` file requires updates to actually make it functional for new layers.

The remaining direct uses of `Layer` constants are in `ZoneRenderer` (where the current rendering order is too nuanced to have an obvious generalization) and the `getToken()` function (which uses `TOKEN` and `GM` as defaults despite them not otherwise sharing any behaviour in common).

### Possible Drawbacks

The "Change To" menu in the right-click menu on the Token-layer now includes the Token layer where it did not before. This matches the behaviour for stamps, but is not ideal.

### Documentation Notes

N/A

### Release Notes

- Refactored layers to make it easier to define new layers, and to maintain existing layers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4402)


cwisniew:
I know the TokenFilter class and subclasses predate this change, but while there are changes in those classes, it would be nice to change the method name to test() rather than accept(), since with the addition of functional interfaces introduced in 8 (yeah this class also predates that) test() implies predicate and accept() implies a consumer.

So while making this change it would be nice to clean that up

cwisniew (approved):

thelsing (approved):

Issue #4354, opened 2023-10-31
[Bug]: JAR build is unable to use GraalJS


kwvanderlinde:
### Describe the Bug

When running MapTool directly from the released JAR file, the `js.createNS()` function fails with this message:
> Event continuing after error running onCampaignLoad@Lib:Elf: java.lang.IllegalArgumentException: Could not find option with name js.ecmascript-version. error executing expression js.evalURI("elf", "lib://Elf/macro/udf.js", "Example").

### To Reproduce

1. Open this campaign file: [4354-graaljs.cmpgn.renamedto.zip](https://github.com/RPTools/maptool/files/13220310/4354-graaljs.cmpgn.renamedto.zip)
    - The `Lib:Elf` token has an `onCampaignLoad` that sets up the JS namespace and runs it sole function.
3. Observe the stated error message.

### Expected Behaviour

For the attached campaign, the message "Test hi" should be printed to chat.

In general, it should be possible to use GraalJS (`js.*` functions) even when running via the JAR release.

### MapTool Info

1.13.2

### Desktop

Linux Mint 21.2

kwvanderlinde:
I've done a bit of sleuthing on this already. The only way this error made sense to me is that GraalJS was somehow not being registered. Some debugging showed that to be correct, instead there was only the "regex" language available but no "js" language.

I found this [GraalJS issue](https://github.com/oracle/graaljs/issues/125) which is also using an uber JAR setup, but with exactly the opposite problem ("js" is available, "regex" is not). In summary, files from different dependencies can overwrite / hide each other when creating an uber JAR, in this case preventing one language from being registered with Graal. _On another note, I've noticed before that 1.13.2 can't be extract via `unzip` due to the presence of conflicting files; instead I have to use `jar xf ...`, and this should have been a red flag to me a long time ago..._

The comments at the end of that issue suggest this can be solved by using the `mergeServiceFiles()` option for [`shadowJar`](https://imperceptiblethoughts.com/shadow/). We don't use `shadowJar`, but I took it for spin anyways on the 1.13.2 tag. And it absolutely worked! Js functions worked perfectly _and my `unzip` issue also disappeared_ :smirk:.

I haven't done much testing beyond that, so I don't know if there are downsides. But at least it looks like there is an avenue to fixing this.

***

To really sum up, our JAR's `/META-INF/services/com.oracle.truffle.api.TruffleLanguage$Provider` file contains:
```
com.oracle.truffle.regex.RegexLanguageProvider
[...]

PR #4421, opened 2023-11-17
Revert inclusion of shadowJar plugin "Replace home-grown uberJar with shadowJar plugin"

kwvanderlinde:

Undoes work for #4354

### Description of the Change

Reverts #4385

Also includes a tweak to #4401 so that duplicate sentry.properties files can be coped with (to reduce annoyance with checking out different commits).

### Possible Drawbacks

Still won't be able to use Graal for now.

### Documentation Notes

N/A

### Release Notes

- Reverted fix for JS UDFs running on the standalone JAR release.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4421)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -58,7 +58,7 @@ jobs:
         pkgname: 'package/archlinux/maptool'
     - name: Build uberJar on Windows
       if: matrix.os == 'windows-latest'
-      run: ./gradlew shadowJar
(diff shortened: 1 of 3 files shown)


kwvanderlinde:
I may have figured out how to make `runtime` and `shadow` play nice. Will leave this PR up for now as testing that might take a while and I don't want to hold other things up.

PR #4449, opened 2023-11-21
Introduce ShapeType.BEAM

### LINKED ISSUE #4448: [Feature]: Line shape type for vision/lights
### Describe the Problem

My group uses a series of narrow cone auras to approximate a line which we use for determining line-of-sight.
It's clumsy and not as pretty as it should be.

### The Solution you'd like

A new Shape Type for Line

_No response_

### Additional Context

_No response_

bubblobill:
resolves #4448

### Description of the Change
Added ShapeType.BEAM for use with vision and lights.
Allows defining a line (really a rectangle) that covers up to 180 degrees of arc of the token's size.

### Possible Drawbacks
none foreseen

### Documentation Notes
New Shape Type for vision and lights, Line.
Refer to i18n light definitions.
image

### Release Notes
Added new shape "Beam" for vision and lighting.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4449)


kwvanderlinde (changes requested):
I'll have a closer look at this later, but a few things that stand out to me right away:

First, there's some new commented-out code that should be removed.

Second, `IsometricGrid.getShapedArea()` also needs an update otherwise lines will be circles there.

Finally, I don't like `arc` as the way to control line width. It's not intuitive to interpret as a user, and it makes the grid size an arbitrary maximum width for the line. That also means a line light is dependent on grid settings which is not good. Instead I would suggest a `width` parameter that allows the width to be written direcltly in map distances. E.g., a very D&D-ish 5'x30' line could be written as `line width=5 30`.

bubblobill:
Good catch, in my quest for impacts I missed the isometric override.

I was envisioning it as an actual line, as it is born of a need for something to assist with determining LoS.
I started out drawing a line but this obviously doesn't work as an area. With no control over stroke/width to improve visibility I needed to use a rectangle.

Navigating dTOs is a nightmare for me and introducing another parameter is more than I am capable of. I used arc= because it was already there, limits width to token size, and makes sense (admittedly, with a picture). 

Your example D&D line isn't really a line; it's actually a proper rectangle and warrants a separate shape (conceptually speaking). I might change my LINE to BEAM to leave room for adding a not-really-a LINE later, as line as an area does occupy a place in the minds of most of our users as an area-of-effect template.

Shall modify my code and go again.

bubblobill:
Revisions done

kwvanderlinde:
I think this is shaping up nicely. I like the term "beam" :slightly_smiling_face: 

If I provide the changes for it, would you be alright with moving away from `arc`?

bubblobill:
Happy to use an alternative to arc if I don't have to do the work : )

kwvanderlinde (approved):
Awesome, looks good, then! I'll take care of changing `arc` -> `width` in a follow-up PR.

bubblobill:
> Awesome, looks good, then! I'll take care of changing `arc` -> `width` in a follow-up PR.

Give me a heads-up when you do and I will update the campaign properties dialogue descriptive text

kwvanderlinde:
> > Awesome, looks good, then! I'll take care of changing `arc` -> `width` in a follow-up PR.
> 
> Give me a heads-up when you do and I will update the campaign properties dialogue descriptive text

Sounds good. I'm hoping to have some time for it this weekend.

PR #4516, opened 2023-12-01
Change BEAM lights and sights to accept width instead of arc

kwvanderlinde:

Improves #4448

### Description of the Change

Instead of an `arc=` parameter, BEAM lights and sights now take a `width=` parameter measured in map units (same as ranges).

Also includes clean up for `SightType` and fixed sight output in `getInfo("campaign")`.

### Possible Drawbacks

It's possible to make beams so wide they don't look like beams anymore.

### Documentation Notes

BEAM lights can take `width=` parameter to set the thickness of the beam.

### Release Notes

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4516)


bubblobill:
The code looks good to me, but I couldn't find a default value for width in the parsing (I just got out of bed so may have missed it). A default width should be thin enough to be useful but thick enough to be visible.
Something like this, fiddlethe value to produce a satisfactory result.
```
this.width = (width == 0) ? 0.2 : width;
this.arcAngle = (arcAngle == 0) ? 90 : arcAngle;
```
Aim for something like this
image

kwvanderlinde:
@bubblobill Any default value or lower bound for width won't necessarily be meaningful for all maps as they can have wildly settings. So instead I changed the meaning of width=0 (or any tiny width) to mean a rectangle that's just about as thin as possible, something we can determine dynamically.

image

thelsing (approved):

Issue #4506, opened 2023-11-29
[Refactoring]: Visibility calculation improvements


kwvanderlinde:
### Describe the problem

There's a lot of things that are suboptimal about our visibility calculations.

On the one hand there's a lot of noise and unnecessary complexity in the code that could be improved. E.g.:
- The structure of `AreaTree` is split up across several types despite being quite straightforward, and its logic is mutually recursive between those types.
- `AreaMeta` can't decide whether to be represented as an AWT `Area` or a JTS `LinearRing`.
- The visibility sweep is not encapsulated at all, but squats in `FogUtil` as a bunch of `static` methods.

There's also a lot of performance being left on the table when topology stops being trivial. E.g.:
- Building `AreaTree` involves an expensive brute force approach to finding parent-child relationships between islands and oceans. This is enough to cause noticeable delay when rebuilding topology trees.
- The use of JTS `Geometry` (including `LineString` and `LinearRing`) when it's not necessary. `Geometry` is expensive to use due to its generallity, but we don't need that generality and so can avoid the overhead.
- The sweep algorithm itself does a poor job of using the right tools (expensive unions to fix already valid geometry, unsorted sets for open walls, hash tables for endpoint deduplication, expensive intersection checks). This really slows things down, and for operations like "Expose last path" it can grind everything to a halt.

### The improvement you'd like to see

`AreaTree` will use a structure that clearly shows it is a basic n-ary tree with values (`AreaMeta`) at each node. A single node type will suffice, and algorithms for `AreaTree` will be defined on `AreaTree` rather than its node type. `AreaMeta` will also be simplified to represent its region solely as a ring of `Coordinate` (avoiding both AWT `Area` and JTS `Geometry`), and will directly use algorithms that can operate on that representation.

The visibility sweep algorithm will be encapsulated in its own class. It will maintain appropriate data structures (sorted open walls especially) and use logic that reflects real-world situations. E.g., that we can have lots of runs of the algorithm in short proximity ("expose last path"), and that we tend to see very few endpoint duplications, a small number of open walls at any given point, and lots of far away walls completely hiding behind nearer walls (for complex topology). The algorithm will also avoid expensive operations (e.g., line intersection calculations)  where cheap and precise alternatives exist.

### Expected Benefits

The code will be easier to follow (if still quite specialized), and performance will get a boost for complex topologies.

### Additional Context

[...]

PR #4529, opened 2023-12-05
Fix MBL in non-noded situations

### LINKED ISSUE #4528: [Bug]: MBL polygons with corners touching edges can break movement
### Describe the Bug

When one region of MBL has a corner that touches the edge of another region, movement blocking can become confused and think a token is in MBL when it is not.

### To Reproduce

1. Open this campaign: [4528-mbl-corner-on-edge.cmpgn.renamed-to.zip](https://github.com/RPTools/maptool/files/13556432/4528-mbl-corner-on-edge.cmpgn.renamed-to.zip)
2. Try to move the Elf token.
3. See the movement stay at "0" and no path is drawn.

### Expected Behaviour

The token should be able to pathfind in the central region.

### Screenshots

Here is what the reproduction looks like:
image

### MapTool Info

1.14.1

### Desktop

Linux Mint 21.2

### Additional Context

_No response_

kwvanderlinde:

Fixes #4528

### Description of the Change

In the test case attached in the bug report, AWT returns the left and right triangles as separate paths, but the top and bottom triangles are treated as part of the surrounding rectangle by using a self-intersecting path. JTS' `Polygonizer` expects well-behaved input, which includes not allowing these self-intersection paths.

To solve this, we now make sure to node our geometry prior to polygonizing. This is done using a `SnapRoundingNoder` to make sure it is robust while also freeing us from having to enforce precision ourselves. And to make sure this sort of problem doesn't fly under the radar in the future, we log any errors that the polygonizer reports.

~~I've tested on some complex geometry and have not noticed any significant slowdown related to this change.~~

### Possible Drawbacks

Performance is reduced for expose last path when the paths are long.

### Documentation Notes

N/A

### Release Notes

- Fixed MBL when a corner of one region is on the edge of another.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4529)

### DIFF
--- a/src/main/java/net/rptools/lib/GeometryUtil.java
+++ b/src/main/java/net/rptools/lib/GeometryUtil.java
@@ -17,16 +17,23 @@ package net.rptools.lib;
 import java.awt.geom.Area;
 import java.awt.geom.Point2D;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.locationtech.jts.awt.ShapeReader;
 import org.locationtech.jts.geom.Coordinate;
 import org.locationtech.jts.geom.Geometry;
 import org.locationtech.jts.geom.GeometryFactory;
-import org.locationtech.jts.geom.LineString;
 import org.locationtech.jts.geom.PrecisionModel;
+import org.locationtech.jts.noding.NodableSegmentString;
+import org.locationtech.jts.noding.NodedSegmentString;
+import org.locationtech.jts.noding.SegmentString;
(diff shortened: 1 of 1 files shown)


kwvanderlinde:
After further testing, I did find a performance regression for expose last path. It's probably not noticable for most typical situations, but for long paths (~400 steps in this case) it does become noticable. This is due to how `GeometryUtil.toJts()` is applied to the vision at each step during expose last path, which causes the noding costs to really add up. Other situations are able to cache the results of the JTS conversion until topology changes, so they are not susceptible to seeing this cost.

The need for `GeometryUtil.toJts()` in this case is one of the many things that will end up being removed by #4506, so performance in this area should improve again once that is done.

cwisniew (approved):

PR #4249, opened 2023-08-16
Add slash command macro functions

cwisniew:
### Identify the Bug or Feature request
resolves #1691 

### Description of the Change
Adds the following macro functions
Set an alias
```
[slash.setAlias("alias", "command")]
```

Returns a JSON Object withe the defined aliases (defined by macro or /alias)
```
[slash.getAliases()]
```

Clears an alias
```
[slash.clearAlias()]
```

Add-ons can also define aliases in a file
`slash_commands.json`
With the following format
```json
{
  "slashCommands": [
    {
      "name": "testSlashCommand",
      "description": "A Test Slash Command",
      "command": "This is a test"
    }, 
    {
      "name": "testSlashCommand1",
      "description": "A Test Slash Command 1",
      "command": "This is a test 1"
    } 

  ]
}
```

This will take care of registering the aliases for you so that you do not need to create any onCampaignLoad macro, it will also separate the commands for your add on in the /alias command listing into their own heading like

CleanShot 2023-08-17 at 20 30 28@2x

Aliases set via macros and add ons have a different "lifetime" to those set using the /alias command.
When set by macro, the aliases are cleared when a new campaign is created, or a campaign is loaded. This is so campaign set aliases will not pollute each other. They are not saved as part of the campaign, so they must be recreated each time if not defined the slash_commands.json file in an add-on.

To maintain functional compatibility, aliases created with /alias command will persist across campaign load/creation and /savealias will not save alises created with `slash.setAlias()` or add-ons.

### Possible Drawbacks
See the above discussion about differences between aliases set with new macro functions and /alias

### Documentation Notes
See above

### Release Notes
-  Add new macro functions for dealing with slash commands
- Add ability for add-ons to define slash commands

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4249)

Issue #3367, opened 2022-02-15
[Feature]: Ability to functionally get/set Chat Color selection.


FullBleed:
### Feature Request

Users can select a color for their chat text via the chat window.  But functional access to get and set that selection isn't currently available and it makes using that selection in secondary resources impossible.

### The Solution you'd like

Add get/setChatColorSelection() functions.

### Alternatives that you've considered.

Adding to getInfo() would make retrieval possible, but would not allow functionally setting the color.

Or people can ignore MT's chat color selection and build all of their chat output macros to use a color selection that is set and saved somewhere else in the campaign.  Not a particularly user friendly way to deal with the issue though.

PR #4532, opened 2023-12-06
Added new chat function _chat_ that allows running slash commands in macros.

### LINKED ISSUE #2273: Function to spoof output like say, emote(s)
I sometimes write macros that emulate chat output through `broadcast` so it appears as an emote or a say command. There may be other default output types that would also be helpful to emulate.

When these default styles change, my framework's users and I have to update macros or they look weird in the updated version of MapTool.

I would like a function which returns a string formatted to appear as if it were output with `/say`, `/me`, or `/mes`, and be able to spoof a player, token, or arbitrary name. 

`spoof( output_string, [output_type], [target], [target_type], [map_name])`

Parameter | Values | Default | Description
-|-|-|-
`output_string` | *any string* | *none* | The string to be formatted
`output_type` | 0 \| "s" <br> 1 \| "emote" \| "me"<br> 2 \| "emotes" \| "mes" \| "say" | 0 | Format `output_string` as `say`, `me`, or `mes`.
`target` | Token ID, Token Name, Player Name, *any string*| *current token* unless no current token exists, then *current player*.  | The target which output will appear to come from.
`target_type` | 0 \| "token"<br> 1 \| "string" | "token" unless no current token exists, then "string". | The target is a token name or ID or an arbitrary string (such as a player name).
`map_name` | Existing map name | Current map | Map to check for a `target` token. Ignored if the `target_type` is not a token.

Example usage:
```
[h: vEmote = spoof("voice booms through a speaker overhead.", "emotes", "Baron McMoneyface", 0, "0. NPC Tokens")]
[h: vSay = spoof("You will never defeat the INVISIBLE HAND!", "say", "Baron McMoneyface", 0, "0. NPC Tokens")]
[h: broadcast(vEmote + vSay)]
```
Expected output:
image

Functions for each respective command would be ok, like `say()`, `emote()`, `emotes()`. Perhaps such functions could call `spoof()` with the `output_type` set appropriately.

1. Updating my framework each MT update which affects output. 
2. Abandon matching defaults or other libraries/frameworks and develop my own scheme for output formatting.

There may be a simpler/more graceful way to achieve what I want; I went for a kitchen-sink function, but maybe a better way exists.

**Additional context**
This approach could be designed to be compatible with future tweaks to the default styles, would let output conform with themes, and would let output automatically respect any future user defined color options. I'm thinking in particular of the changes to `/say` in 1.7 and of #2271.

bubblobill:
resolves #2273

### Description of the Change
Moved AboutMacro to same location as other slash command macros.
New function slash(command, content) allows running chat slash commands in macros.
e.g. [chat("me", "does something")]
*Elf does something

Potential security issue with access to user file system with some commands. Excluded /loadalias, /savealias, /experiments, /savetokenstates, and /loadtokenstates.

Also exluded /alias, and /clearaliases as they are user specific.

Excluded 
### Possible Drawbacks
Relies on chat macro permissions.
Should be robust.

### Documentation Notes
Chat macros now available through new function **chat()**

### Release Notes
Chat macros now available through new function **chat()**

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4532)


bubblobill:
#3367

Irarara:
Gonna say, I don't really like the idea of GMs being able to force the execution of slash commands on connected clients, which this would make possible. This opens up changing some client settings and manipulating things that (imo) only the local client should have control over (for instance, chat color and aliases) and it feels like it could be used maliciously. 

At the very least, only a subset of slash commands should work with the function, but I'm not really convinced this is desirable at all. It doesn't actually address #2273 and I'm not really sure what the other use cases for this are. #3367 was mentioned, but, personally, I'm not on board with the idea of setting chat color via macros, and this doesn't help with getting the color.

bubblobill:
It's mostly a response to people wanting to get emote and other formatted outputs via macro. Easy enough to filter it to limit functionality.

On the other side, being able to create a nice form for setting chat colour, toggling perlin noise, etc. is not a bad thing.

bubblobill:
Hold up. Just realised I haven't added them to the parser so they won't get very far.

cwisniew:
@bubblobill since there is already a slash.* prefix as part of #4249. This should probably be renamed to slash.exec() or slash.run() to save confusion.

bubblobill:
How about I just call it chat() since all it does is pass the text as if it was chat direct input?

cwisniew:
> How about I just call it chat() since all it does is pass the text as if it was chat direct input?

but I also agree with @Irarara there should be some things that shouldn't be allowed to be executed like saving aliases so you would need to modify it not to just pass it as if it was chat.

bubblobill:
> > How about I just call it chat() since all it does is pass the text as if it was chat direct input?
> 
> but I also agree with @Irarara there should be some things that shouldn't be allowed to be executed like saving aliases so you would need to modify it not to just pass it as if it was chat.

You just told me there may be confusion with the feature that sets aliases. tsk tsk tsk.

Which features will we allow/or deny? (whichever list is shorter)

cwisniew:
> > > How about I just call it chat() since all it does is pass the text as if it was chat direct input?
> > 
> > 
> > but I also agree with @Irarara there should be some things that shouldn't be allowed to be executed like saving aliases so you would need to modify it not to just pass it as if it was chat.
> 
> You just told me there may be confusion with the feature that sets aliases. tsk tsk tsk.
> 
> Which features will we allow/or deny? (whichever list is shorter)

I don't see how my statements are mutually exclusive :)

We probably shouldn't allow
/loadalias, /savealias, /experiments, /loadtokenstates as they access the user's filesystem. Probably not /alias, /clearalias either as they are the user's aliases, the above-mentioned alias slash command functionality can be used to create aliases from macros that don't interfere with user aliases. There seems no point for /help or /about to be executed so might as well exclude them too.


There is a net.rptools.maptool.client.macro.MacroDefinition annotation that you can add a field to to indicate if it is allowed or not as that has all the other information about slash commands.

bubblobill:
@cwisniew 
Changed to **chat()**
Updated with blacklist for; "alias", "cc", "chatcolor","clearaliases", "loadaliases", "savealiases", "exp", "exper", "experiments", "loadtokenstates", "tsl", "savetokenstates", and "tss".
Should now pass muster

cwisniew:
> @cwisniew Changed to **chat()** Updated with blacklist for; "alias", "cc", "chatcolor","clearaliases", "loadaliases", "savealiases", "exp", "exper", "experiments", "loadtokenstates", "tsl", "savetokenstates", and "tss". Should now pass muster

Almost passes muster :)
There are a few if statements that done have braces, even single line if's should include the { } just so everything is standard... In other news after this comment I need to do a web search for voodoo doll counters.

bubblobill:
Talk about being retentive. 
Isn't there a coding style guide somewhere that says code should be concise and easy to read?
I'ma gonna look for an alternative to shitless.

You'll notice I even fixed an "if" without braces that wasn't mine.

cwisniew:
> Talk about being retentive. Isn't there a coding style guide somewhere that says code should be concise and easy to read? I'ma gonna look for an alternative to shitless.
> 

Part of being easy to read is consistency... Yeah the google java format has issues enforcing that rule :(

> You'll notice I even fixed an "if" without braces that wasn't mine.

Fantastic! I shall prepare my dinasour stamp!

cwisniew (approved):

cwisniew:
CleanShot 2024-01-02 at 23 44 29@2x

Issue #3691, opened 2022-09-29
[Refactoring]: Split up ZoneRenderer to enable an easier transition to other rendering implementation


kwvanderlinde:
### Describe the problem

The `ZoneRenderer` class is both large and too interconnected with itself. Making changes to it is always risky as it's not clear which parts of the `ZoneRenderer` will be affected and which will not. On top of that, large and fundamental changes such as moving to different renderers (e.g., libGDX) are particularly difficult, at least not without duplicating core logic and bringing associated clutter.

### The improvement you'd like to see

I'd like to see the `ZoneRenderer` broken down into parts that can work together but are much more managable.

1. The rendering logic under `ZoneRenderer#paintComponent()`/`ZoneRenderer::renderZone()` is split up into separate render layers.
    - Each layers would maintain its own internal state (e.g., caching light sources in the lighting layer) so that the `ZoneRenderer` has less state.
    - The `ZoneRenderer` would have much less logic in it, and the rendering operation would be simplified to drawing each render layer in turn.
    - This would make rendering changes more focused as modifying one render layer is not as likely to interfere with another.
2. `ZoneRenderer`s nature as a Swing component is removed. Instead, the `ZoneRenderer` can expose a reference to some Swing component, but which component is determined by something other than the `ZoneRenderer`.
    - Another rendering implementation (like libGDX) could then provide a different Swing component than the current one.
    - The `ZoneRenderer` would only be responsible for providing an interface to the visible zone, while the Swing components would be responsible for effecting renders.
3. Define a graphics context that provides the necessary fundamental operations we need to draw a zone.
    - E.g., operations for drawing an image at a certain location, for transforming the view port, and for setting clips.
    - We can make more than one implementation. E.g., a Swing implementation that interfaces with a `Graphics2D`, while a libGDX implementation could do whatever it needed to do.
    - The render layers would use this graphics context instead of relying directly on either a `Graphics2D` or any libGDX stuff.
4. Pull rendering logic out of non-renderer components.
    - E.g., remove `Grid#draw()` and `Grid#drawCoordinatesOverlay()` and instead make that the responsibility of a render layer.

The end result (if all my ducks are in a row) is that:
- The `ZoneRenderer` defines some core logic, such as what order to render things in, what actually needs to be rendered, etc.
- The logic behind rendering is determined by the various render layers. E.g., deciding where to paint light sources in what shape, etc.
- The details of how to render each element is determined by our graphics context.
[...]

PR #4538, opened 2023-12-07
Factor out more rendering logic into separate renderers

### LINKED ISSUE #4429: [Bug]: Soft FOW outline is rendered overtop hard FOW
### Describe the Bug

If auto-expose FOW is disabled, moving a token can result in its visible area overlapping the hard FOW. If that happens, the soft FOW outline is rendered overtop of the hard FOW.

### To Reproduce

1. Disable auto-expose fog in the **Preferences** dialog.
2. Create a new campaign
3. Set map vision to Night
4. Change the map fog to a color other than black.
5. Draw some VBL on the map.
6. Expose an area of the map.
7. Place a PC token with darkvision in the exposed area
8. Enable player view
9. Observe a black outline where the soft FOW would start if the area were exposed.
10. Right click the token, Expose > Visible Area.
    - Now things look "normal", the soft FOW does not overlap with the hard FOW anywhere.

### Expected Behaviour

The FOW outline should be constrained to the exposed FOW area.

### Screenshots

Here is an example from the campaign attached below: 
image

### MapTool Info

1.13.2, 1.14.1

### Desktop

Linux Mint 21.2

### Additional Context

Here is a campaign with the problem scenario already present: [fog-1.14.1.cmpgn.renamed-to.zip](https://github.com/RPTools/maptool/files/13401544/fog-1.14.1.cmpgn.renamed-to.zip)

kwvanderlinde:

Progresses #3691
Fixes #4429

### Description of the Change

Rendering of lights, auras, player darkness, fog of war, vision, and debug shapes are now handled by separate renderer classes. For the most part, logic was simply copied over, but some simplifications and alterations were made.

Points of interest:
- The new `RenderHelper` takes care of setting up `Graphics2D` with initial clips and transforms. It can optionally use a temporary buffer as a render target. Each of the new renderers uses the `RenderHelper`.
- Fog rendering case work has been consolidated and simplified to make it a bit more direct in terms of what is happening.
- The player-specific logic for vision overlays has been eliminated. The clip set there was not used (it was always overwritten prior to use), and the logic was broken anyways due to mixing screen space and world space areas.
- The debug shape renderer shows that shapes are translucent so they don't completely occlude what is rendered behind it.

### Possible Drawbacks

Should be none.

### Documentation Notes

N/A

### Release Notes

- More modularization of the zone renderer for fog, lights, and vision.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4538)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/ZoneView.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/ZoneView.java
@@ -610,12 +610,7 @@ public class ZoneView {
    * @param token the token to get the visible area of.
    * @return the visible area of a token, including the effect of other lights.
    */
-  public Area getVisibleArea(Token token, PlayerView view) {
-    // Sanity
-    if (token == null) {
-      return null;
-    }
-
+  public Area getVisibleArea(@Nonnull Token token, PlayerView view) {
     // Cache ?
     Map<GUID, Area> tokenVisionCache =
         tokenVisionCachePerView.computeIfAbsent(view, v -> new HashMap<>());
--- a/src/main/java/net/rptools/maptool/client/ui/zone/renderer/DarknessRenderer.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/renderer/DarknessRenderer.java
@@ -1,0 +1,57 @@
+/*
+ * This software Copyright by the RPTools.net development team, and
+ * licensed under the Affero GPL Version 3 or, at your option, any later
+ * version.
+ *
+ * MapTool Source Code is distributed in the hope that it will be
+ * useful, but WITHOUT ANY WARRANTY; without even the implied warranty
+ * of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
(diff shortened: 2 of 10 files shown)


bubblobill:
Personal preference, when defining colours, include alpha. That way it's more universally applicable and a little bit future-proofed.
From
 `palette = new Color[] {Color.red, Color.green, Color.blue};`
 to
 `palette = new Color[] {Color.red, Color.green, Color.blue, Color.alpha};`

From
`var fillColor = color.darker();
      fillColor = new Color(fillColor.getRed(), fillColor.getGreen(), fillColor.getBlue(), 76);`
to
`var fillColor = color.darker();
      fillColor = new Color(fillColor.getRed(), fillColor.getGreen(), fillColor.getBlue(), fillColor.getAlpha() / 3f );`

kwvanderlinde:
Sure, makes sense, made the change. Though I'll point out `palette` is just a sequence of colours to cycle through, not itself a color with an alpha.

cwisniew (approved):

cwisniew (approved):

PR #4557, opened 2023-12-12
Fix note rendering when labels are present.

kwvanderlinde:

Improves #3691

### Description of the Change

With the switch to `setRenderingHints()` for restoring graphics state, we ran into a bug in `Graphics2D` where its cached font state is not invalidated since it doesn't detect that a change has occurred. This causes the wrong state to be used back in `paintComponent()`, but only if the problematic state is in play (e.g., if a label has been rendered).

The new approach has `renderZone()`  work on a copy of the `Graphics2D` object, so that `paintComponent()` isn't affected by any changes to its state. This also removes the need for `renderZone()` to reset any state, including rendering hints and clips. To be sure it works properly, `paintComponent()` has to set its font prior to rendering the notes now.

### Possible Drawbacks

Should be none.

### Documentation Notes

N/A

### Release Notes

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4557)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/renderer/ZoneRenderer.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/renderer/ZoneRenderer.java
@@ -819,9 +819,12 @@ public class ZoneRenderer extends JComponent
             timer.stop("paintComponent:createView");
 
             renderZone(bufferG2d, pl);
-            int noteVPos = 20;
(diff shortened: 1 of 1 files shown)


bubblobill (approved):
Strange bug.
The fix seems sound to me.

cwisniew (approved):

PR #4579, opened 2023-12-22
Auras and Lights Through VBL

### LINKED ISSUE #4578: [Feature]: Auras and Light Through Walls
### Describe the Problem

Some systems have effects that do not adhere to line of sight - ie. "all creatures within 60feet" - not "all visible creatures within 60 feet". Auras and lights that would benefit towards assisting with these cases do not exist.

### The Solution you'd like

Auras and Lights gain a new option "PASSTHRU" where VBL is ignored when determining the light area.

Measuring from every relevant token - but that's not as quick and easy as a visible circle.

### Additional Context

_No response_

ColdAnkles:
closes #4578

### Description of the Change
Adds basic option in light campaign properties to allow auras/lights to phase through VBL.

### Possible Drawbacks
Complicates future, more robust systems.

### Documentation Notes
Adds "IGNORE-VBL" as option key in Campaign Light Properties for disabling VBL effect on lights/auras.

### Release Notes
Added option for disabling VBL effects on lights/auras.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4579)


bubblobill:
No. PASSTHRU is a bad keyword. It's not a real word, it's bastardised bad English, which makes it harder to translate. 
No objection to the functionality, I'm assuming there is a use-case which is why it was written, but it needs a better keyword. 
Imagine someone with bad English trying to explain it to someone with no English. 
"Unblocked" is better but there is probably a better word than that.

bubblobill:
New thought. Why limit this to auras? It's Vision Blocking, not Light Blocking. A Car park light can light up the whole car park because of its height.

src/main/java/net/rptools/maptool/client/ui/zone/FogUtil.java
```diff
@@ -96,10 +96,19 @@ public class FogUtil {
      */
     List<Geometry> visibleAreas = new ArrayList<>();
     final List<Function<VisionBlockingAccumulator, Boolean>> topologyConsumers = new ArrayList<>();
-    topologyConsumers.add(acc -> acc.addWallBlocking(topology));
-    topologyConsumers.add(acc -> acc.addHillBlocking(hillVbl));
-    topologyConsumers.add(acc -> acc.addPitBlocking(pitVbl));
-    topologyConsumers.add(acc -> acc.addCoverBlocking(coverVbl));
+    if (topology != null) {
```
kwvanderlinde: No need to support `null` here. Callers should pass an empty tree if no blocking should be applied. But also see my other comment about why the new calls with `null` don't even need to be made.

src/main/java/net/rptools/maptool/client/ui/zone/ZoneView.java
```diff
@@ -347,14 +347,21 @@ private List<ContributedLight> calculateLitAreaForLightSource(
...
-            lightSourceArea,
-            getTopologyTree(Zone.TopologyType.WALL_VBL),
-            getTopologyTree(Zone.TopologyType.HILL_VBL),
-            getTopologyTree(Zone.TopologyType.PIT_VBL),
-            getTopologyTree(Zone.TopologyType.COVER_VBL));
+    Area lightSourceVisibleArea = null;
+
+    if (lightSource.isPassThroughVBL()) {
+      lightSourceVisibleArea =
+          FogUtil.calculateVisibility(p, lightSourceArea, null, null, null, null);
```
kwvanderlinde: This call can be removed as it will just return a copy of `lightSourceArea` back again.

bubblobill (approved):
Not familiar enough with light code to spot errors, but I see no holes in it.

src/main/java/net/rptools/maptool/client/functions/TokenLightFunctions.java
```diff
@@ -322,6 +322,10 @@ private static LightSource createUniqueLightSource(
             : LightSource.Type.NORMAL;
     final boolean scaleWithToken =
         lightSourceDef.has("scale") ? lightSourceDef.get("scale").getAsBoolean() : false;
+    final boolean ignoreVBL =
+        lightSourceDef.has("ignores-vbl")
+            ? lightSourceDef.get("ignores-vbl").getAsBoolean()
```
kwvanderlinde, line 327: We should be consistent with the campaign properties syntax about whether to use `ignore-vbl` or `ignores-vbl` otherwise it will lead to confusion. I slightly prefer `ignores-vbl`, but doesn't really matter.

The flag should also be added to the result in `lightSourceToJson()`, and should be added to the result of `getInfo("campaign")` (impl is in `getInfoFunctions()`).

src/main/java/net/rptools/maptool/util/SightSyntax.java
kwvanderlinde, line 1: Need to add `IGNORE-VBL` in the stringify results.

kwvanderlinde, line 1: On second thought, I don't think this flag should be applied to sights as it will lead to confusion. It won't make the sight itself ignore VBL, it will just make make the associated personal light ignore VBL. Since lights are clipped to visible areas for players but not for GMs, this will lead to GMs thinking the sight itself ignores VBL (even though it's just the personal light that does), while players will see the personal lights as though they respect VBL, despite the flag.

... or we make sights respect their personal light's IGNORE-VBl, but I'm not sure that's the right choice. Opinions welcome.

kwvanderlinde (changes requested):

src/main/java/net/rptools/maptool/util/SightSyntax.java
ColdAnkles: I was originally going to work in x-ray vision, but decided against it, and have removed this less useful legacy of that bad idea.

ColdAnkles:
I think I got all the required updates

kwvanderlinde (approved):

bubblobill:
Damnit. I just had a thought. It should be IGNORES-VB
a) We need to stop sticking the L on the abbreviation when it is unnecessary, because, English, and
b) If His Majesty implements layering/elevation, there is a difference between VB & VBL,
Can you tweak the parser to accept VB as well as VBL?

cwisniew:
> Damnit. I just had a thought. It should be IGNORES-VB a) We need to stop sticking the L on the abbreviation when it is unnecessary, because, English, and b) If His Majesty implements layering/elevation, there is a difference between VB & VBL, Can you tweak the parser to accept VB as well as VBL?

If there are going to be layer groups, there would still be a layer in each group for vision blocking. Mind you VBL has been less than accurate since vision blocking could be included on tokens etc. But part of me just wonders if you are being sponsored by CUB in your quest to rebrand everything as VB :)

bubblobill:
Yes, but what if my light is to transcend layers? Then it's just a beer with no L
BTW. Help text and translation file has been updated to include this.

PR #4596, opened 2023-12-30
Alternate Encoding Handling for Addon Content

### LINKED ISSUE #4594: [Bug]: Asset Constructor Doesn't Handle UTF-16 Text Files When Loading Addon
### Describe the Bug

When loading an Addon that contains a file encoded in UTF-16 text, will include control characters in string conversion.

### To Reproduce

1. Create a MapTool addon
2. Include in the addon a JSON file encoded in UTF-16
3. Attempt to load the error
4. Receive a MalformedJSON exception

### Expected Behaviour

Addon loaded as expected

### Screenshots

_No response_

### MapTool Info

1.14.3

### Desktop

Windows 11

### Additional Context

/maptool/src/main/java/net/rptools/maptool/model/Asset.java: Line 474

ColdAnkles:
fixes #4594

### Description of the Change
Adds an ugly try/catch chain that attempts to determine the encoding of the byte array instead of assuming UTF8

### Possible Drawbacks
Only handles ASCII, UTF-8, UTF-16, or Latin1 - and doesn't read file if can't determine encoding (null).

### Documentation Notes
Uses Standard Java Charset Decoding, assuming different encodings until it matches or fails entirely.

### Release Notes
Added Encoding Checks for Addon File Import

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4596)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/Asset.java
+++ b/src/main/java/net/rptools/maptool/model/Asset.java
@@ -494,7 +494,7 @@ public final class Asset {
             try {
               decoder = StandardCharsets.ISO_8859_1.newDecoder();
               decodedString = decoder.decode(ByteBuffer.wrap(data)).toString();
-            } catch (Exception eThree) {
+            } catch (Exception eFour) {
               decodedString = null;
             }
           }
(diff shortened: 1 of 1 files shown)


cwisniew:
There is no need to try it as ASCII first as UTF8 decoder should read an ASCII file fine. Also not sure about ISO_8859_1 as we don't want to go down the path of supporting detecting and reading every character set out there, UTF 8 and 16 are enough (well more than enough really :) )

cwisniew (approved):

PR #4616, opened 2024-01-03
Remove leftover bits from the campaign export removal

### LINKED ISSUE #4047: [Maintenance]: Remove campaign export functionality
### Describe the problem

The campaign export has an awkward existence in today's MapTool.
1. The set of supported versions seems arbitrary to a user (it only allows 1.4.0 through 1.5.1). There's no support for exporting to more recent versions that users are more likely to encounter today.
2. The supported versions predate the 1.8.3 security fix. We've sternly warned users not to use such versions of MapTool, so it seems irresponsible to offer this invitation to downgrade.
3. The functionality has not been maintained and no longer functions properly. The exporter essentially operates by modifying the data to remove things the older versions don't expect. This has not been kept up over time, so there is now lots of stuff that has been added which is not removed by the exporter. I cannot today take my 1.13.0 campaign, export it to 1.4.0.0 and expect it to work.
4. The campaign exporter is a maintenance burden. While not a large amount of code, it pops up when doing searches and the like, and just serves to make developers second-guess whether they're supposed to do anything with it.

The campaign export is removed in its entirety.

### Expected Benefits

One less thing to have to think about and wade through.

### Additional Context

I'm also noticing `ModelVersionManager` is in a similarly awkward position, though it's invisible to the user and only handles upgrades. Still, it only has logic for really old versions (1.3.51 - 1.3.78), is more invasive than the campaign export logic, and I believe the handful of XML transforms that it does can instead be handled via `readResolve()`.

kwvanderlinde:

Improves #4047

### Description of the Change

PR #4065 removed the code for campaign exports, but the translation strings and a form file were left behind. This PR fixes that up.

One string `msg.confirm.campaignExported` was mistakenly still in use as the confirmation for exporting campaign repository files. This has been replaced with `msg.confirm.campaignRepoExported` with a more accurate message.

### Possible Drawbacks

None

### Documentation Notes

N/A

### Release Notes

- Fixed confirmation message shown when exporting a campaign repository.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4616)


bubblobill (approved):
You caught all the strings I could find.

cwisniew (approved):

PR #4621, opened 2024-01-08
Input() to handle `lib://` asset paths

### LINKED ISSUE #4620: [Feature]: Input() Icon to Accept lib:// image assets
### Describe the Problem

Input() can accept `asset://` URIs to add images to some Input options, but does not accept `lib://` URIs.

### The Solution you'd like

Input function to handle both `asset://` and `lib://` URIs seamlessly.

Add workaround image tokens to allow for `asset://` URIs usage.
Do some insane token spawning and set tokenImages (since that accepts `lib://` image paths) to translate to a `asset://` URI

### Additional Context

_No response_

ColdAnkles:
resolves #4620

### Description of the Change
Updates regex to capture asset string and pass through AssetResolver to handle both `asset://` and `lib://` paths

### Possible Drawbacks
None

### Documentation Notes
Regex capture includes the 'asset://' OR 'lib://' portion of the URI
Now depends on Asset Resolver

### Release Notes
Added ability for Input() to handle `lib://` URIs

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4621)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/InputFunction.java
+++ b/src/main/java/net/rptools/maptool/client/functions/InputFunction.java
@@ -1273,7 +1273,9 @@ public class InputFunction extends AbstractFunction {
   /** Gets icon from the asset manager. Code copied and modified from EditTokenDialog.java */
   private ImageIcon getIcon(String id, int size, ImageObserver io) {
     // Extract the MD5Key from the URL
-    if (id == null) return null;
+    if (id == null) {
+      return null;
+    }
     var assetKey = new AssetResolver().getAssetKey(id);
     String assetId = null;
     if (assetKey.isPresent()) {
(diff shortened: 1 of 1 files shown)


src/main/java/net/rptools/maptool/client/functions/InputFunction.java
```diff
@@ -1272,10 +1274,20 @@ public void checkParameters(String functionName, List<Object> parameters)
   private ImageIcon getIcon(String id, int size, ImageObserver io) {
     // Extract the MD5Key from the URL
     if (id == null) return null;
```
cwisniew: I know it was like this before you edited it, but if you don't mind can you change 
```
if (id == null) return null;
```
to 
```
if (id == null) {
  return null;
}
```
Just so it meets the coding standards and is consistent, and we don't have to create a separate PR to address that line y itself :)

Thanks

cwisniew (changes requested):

cwisniew (approved):

PR #4629, opened 2024-01-11
Updated FunctionUtil to handle `lib://` assets via AssetResolver

### LINKED ISSUE #4620: [Feature]: Input() Icon to Accept lib:// image assets

### LINKED ISSUE #4626: [Bug]: createMap does not accept add-on URIs
### Describe the Bug

It is not possible to use the new `createMap` function with an image from an add-on. 
When I attempt to use the function with e.g. these parameters: 
```javascript
[h:lobbyImage = "lib://net.dovesoft.trinity-continuum/images/LobbyBackground.png"]
[h: lobbyId = createMap("01. Lobby", 
    json.set("{}",
        "player visible", json.true,
        "vision type", "Off",
        "vision distance", 1000,
        "lighting style", "OVERTOP",
        "has fog", json.false,
        "ai rounding", "None",
        "map asset", lobbyImage
    )
)]
```

I get the following exception: 
```
20:41:53.233 DEBUG net.rptools.maptool.client.MapToolLineParser - net.rptools.parser.ParserException: createMap: "lib://net.dovesoft.trinity-continuum/images/StoryGuideScreen.png" is not a valid asset ID or URL
20:41:53.233 ERROR net.rptools.maptool.client.swing.MapToolEventQueue - java.util.concurrent.CompletionException: net.rptools.parser.ParserException: createMap: "lib://net.dovesoft.trinity-continuum/images/LobbyBackground.png" is not a valid asset ID or URL
java.util.concurrent.CompletionException: net.rptools.parser.ParserException: createMap: "lib://net.dovesoft.trinity-continuum/images/LobbyBackground.png" is not a valid asset ID or URL
    at net.rptools.maptool.model.library.addon.AddOnLibrary.lambda$callMTSFunction$15(AddOnLibrary.java:653) ~[MapTool.jar:1.14.3]
    at java.awt.event.InvocationEvent.dispatch(Unknown Source) [?:?]
    at java.awt.EventQueue.dispatchEventImpl(Unknown Source) [?:?]
```
The image url is known to work, as I have tested it with `[h:broadcast("</img>")]`, which showed the image nicely in the chat window. 

### To Reproduce

1. Create an add-on
2. Add an image in the public folder and an onInit.mts file to the mtscript folder. 
3. In the onInit.mts add this: 
```javascript
[h:lobbyImage = "lib://test-library-namespace/LobbyBackground.png"]
[h: lobbyId = createMap("01. Lobby", 
    json.set("{}",
        "player visible", json.true,
        "vision type", "Off",
        "vision distance", 1000,
        "lighting style", "OVERTOP",
        "has fog", json.false,
        "ai rounding", "None",
        "map asset", lobbyImage
    )
)]
```
Where the lobbyImage variable points to the correct image with the correct namespace. 
4. Import the add-on 
5. Observe the exception. 

### Expected Behaviour

I expect the lib:// URI to be treated exactly as an asset:// URI. Even more so, as the ´createMap` function is so newly added. 

### Screenshots

_No response_

### MapTool Info

1.14.3

### Desktop

Windows 11

### Additional Context

_No response_

ColdAnkles:
fixes #4626
resolves #4620 in a smarter fashion

### Description of the Change
Updates FunctionUtil.getAssetKeyFromString()  to handle both `asset://` and `lib://` asset URLs
Propagates that usage to both createMap, and input functions.

### Possible Drawbacks
None

### Documentation Notes
FunctionUtil now depends on AssetResolver imports
InputFunction now depends on FunctionUtil

### Release Notes
Fixed several inconsistent inabilities to use `lib://` assets

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4629)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/MapFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/MapFunctions.java
@@ -20,6 +20,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
+import net.rptools.lib.MD5Key;
 import net.rptools.maptool.client.AppPreferences;
 import net.rptools.maptool.client.MapTool;
 import net.rptools.maptool.client.MapToolUtil;
@@ -32,6 +33,7 @@ import net.rptools.maptool.model.InvalidGUIDException;
 import net.rptools.maptool.model.Zone;
 import net.rptools.maptool.model.ZoneFactory;
 import net.rptools.maptool.model.drawing.DrawablePaint;
+import net.rptools.maptool.util.AssetResolver;
 import net.rptools.maptool.util.FunctionUtil;
 import net.rptools.parser.Parser;
 import net.rptools.parser.ParserException;
@@ -280,13 +282,14 @@ public class MapFunctions extends AbstractFunction {
 
       if (config.has("map asset")) {
         final var mapAssetId = config.getAsJsonPrimitive("map asset").getAsString();
-        final var mapAssetKey = FunctionUtil.getAssetKeyFromString(mapAssetId);
-        if (mapAssetKey == null) {
-          throw new ParserException(
-              I18N.getText("macro.function.map.invalidAsset", functionName, mapAssetId));
+        final var mapAssetKey = new AssetResolver().getAssetKey(mapAssetId);
+        MD5Key assetMD5 = null;
+        if (mapAssetKey.isPresent()) {
+          assetMD5 = new MD5Key(mapAssetKey.get().toString());
+        } else {
+          assetMD5 = new MD5Key(mapAssetId);
         }
-
-        final var mapAsset = AssetManager.getAsset(mapAssetKey);
+        final var mapAsset = AssetManager.getAsset(assetMD5);
         if (mapAsset != null) {
           AssetManager.putAsset(mapAsset);
           if (!MapTool.isHostingServer()) {
@@ -294,7 +297,7 @@ public class MapFunctions extends AbstractFunction {
           }
         }
 
-        newMap.setMapAsset(mapAssetKey);
+        newMap.setMapAsset(assetMD5);
       }
 
       MapTool.addZone(newMap, false);
--- a/src/main/java/net/rptools/maptool/client/functions/MapFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/MapFunctions.java
@@ -20,7 +20,6 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
-import net.rptools.lib.MD5Key;
 import net.rptools.maptool.client.AppPreferences;
 import net.rptools.maptool.client.MapTool;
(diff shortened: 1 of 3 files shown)


src/main/java/net/rptools/maptool/client/functions/MapFunctions.java
```diff
@@ -280,21 +282,22 @@ public Object childEvaluate(
 
       if (config.has("map asset")) {
         final var mapAssetId = config.getAsJsonPrimitive("map asset").getAsString();
-        final var mapAssetKey = FunctionUtil.getAssetKeyFromString(mapAssetId);
```
kwvanderlinde: If we instead change `FunctionUtil.getAssetKeyFromString()`, than all callers will support `lib:` URIs. I think that's just here and some table functions, but the point stands :slightly_smiling_face:

src/main/java/net/rptools/maptool/client/functions/MapFunctions.java
```diff
@@ -280,21 +282,22 @@ public Object childEvaluate(
...
-        if (mapAssetKey == null) {
-          throw new ParserException(
-              I18N.getText("macro.function.map.invalidAsset", functionName, mapAssetId));
+        final var mapAssetKey = new AssetResolver().getAssetKey(mapAssetId);
+        MD5Key assetMD5 = null;
+        if (mapAssetKey.isPresent()) {
+          assetMD5 = new MD5Key(mapAssetKey.get().toString());
+        } else {
+          assetMD5 = new MD5Key(mapAssetId);
```
kwvanderlinde: I haven't tried it out, but this looks like it hides any failure when the user passes an invalid MD5. Or it would fail later on, but then the message would be less informative. We should prefer giving a clear and direct error message.

src/main/java/net/rptools/maptool/client/functions/MapFunctions.java
ColdAnkles: That is smarter, and now done. Rewrote my previous PR for Input functions to use the same.

src/main/java/net/rptools/maptool/client/functions/MapFunctions.java
ColdAnkles: I've reinstated the exception with the error

ColdAnkles:
If we can find any similar uses of `asset://` only I can update for those as well.

bubblobill:
sooo... if I have an image: token, what happens?

ColdAnkles:
> sooo... if I have an image: token, what happens?

Was you'd get a null, but now it should work, though I suspect there's a better way to implement.

Probably should discourage using Image token names directly in this way, since afaik token names aren't unique.

cwisniew (approved):

Issue #4638, opened 2024-01-15
[Bug]: Missing Icons from Rods iconset


thelsing:
### Describe the Bug

There are some missing icons in the icon set. 

### To Reproduce

1. Start MT.
2. See ugly up-arrow in top left.

### Expected Behaviour

Rods iconsets contains replacements for all classic icons.

thelsing:
After PR #4639 there are still some icons missing:
ACTION_CANCEL classic: net/rptools/maptool/client/image/cancel_sm.png
ACTION_CLOSE classic: net/rptools/maptool/client/image/collapse.png
ACTION_NEW_SMALL classic: net/rptools/maptool/client/image/add_sm.png
ACTION_OPEN classic: net/rptools/maptool/client/image/expand.png
ADD_RESSOURCE_RPTOOLS classic: net/rptools/maptool/client/image/rptools_icon.png
ADD_RESSOURCE_WEB classic: net/rptools/maptool/client/image/download.png
ASSETPANEL_PDF classic: net/rptools/maptool/client/image/pdf_icon.png
ASSETPANEL_PDF_FOLDER classic: net/rptools/maptool/client/image/pdf_folder.png
ASSETPANEL_SEARCH classic: net/rptools/maptool/client/image/zoom.png
CHAT_NOTIFICATION classic: net/rptools/maptool/client/image/chat-blue.png
CHAT_SCROLL_LOCK_OFF classic: net/rptools/maptool/client/image/comments_delete.png
CHAT_SMILEY classic: net/rptools/maptool/client/image/smiley/emsmile.png
DRAWPANEL_AREA_DRAW classic: net/rptools/maptool/client/image/tool/drawpanel-poly.png
DRAWPANEL_AREA_ERASE classic: net/rptools/maptool/client/image/tool/drawpanel-poly-erase.png
DRAWPANEL_ELLIPSE_DRAW classic: net/rptools/maptool/client/image/tool/drawpanel-ellipse.png
DRAWPANEL_ELLIPSE_ERASE classic: net/rptools/maptool/client/image/tool/drawpanel-ellipse-erase.png
DRAWPANEL_LINE_DRAW classic: net/rptools/maptool/client/image/tool/drawpanel-line.png
DRAWPANEL_LINE_ERASE classic: net/rptools/maptool/client/image/tool/drawpanel-line-erase.png
DRAWPANEL_POLYGON_DRAW classic: net/rptools/maptool/client/image/tool/drawpanel-poly.png
[...]

PR #4639, opened 2024-01-15
Add more icons from Rod Takehara

thelsing:
related to  #4638

### Description of the Change

This PR add some more icons that Rod provided and updates flatlaf to 3.3. As flatlaf dropped the *Contrast* themes, I removed them from MT too.

### Possible Drawbacks

Users might use those themes.

### Release Notes

- added more icons to icon theme "Rod Takehara"

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4639)


cwisniew:
I specifically haven't upgraded to 3.3 of flatlaf because of the dropped themes. On my list of things to do (but low priority as flatlaf 3.3 doesn't seem to add to much for us) is check to see if those themes are still available from https://plugins.jetbrains.com/search?tags=Theme and include them (like @bubblobill  theme was included).

thelsing:
I didn't find any of the "Contrast" variants in the jetbrains search. Only the normal variants. IMO if anyone really wants one of the removed themes back they can raise an issue. 
The 3.3 version of flatlaf claims to use less memory with svg icons because they use a different svg lib. But I have no problem with reverting the flatlaf version bump.

bubblobill:
Contrast themes are more likely to be used by people that need them than most of the many themes we have.
Best to hunt down the themes and include them because they will ask to have them back.
Accessibility is one area we can nail the competition.

FullBleed:
I use the Monokai Pro Contrast (Material) theme.

Wonder why the contrast themes were dropped...

thelsing:
I pulled the *Contrast* themes from flatlaf and added them to MT.

kwvanderlinde:
> I pulled the _Contrast_ themes from flatlaf and added them to MT.

Careful with the licensing here ...

thelsing:
> > I pulled the _Contrast_ themes from flatlaf and added them to MT.
> 
> Careful with the licensing here ...

Flatlaf uses apache license, so I see no problem with using the code.

kwvanderlinde:
> > > I pulled the _Contrast_ themes from flatlaf and added them to MT.
> > 
> > 
> > Careful with the licensing here ...
> 
> Flatlaf uses apache license, so I see no problem with using the code.

That's my understanding as well, but there are requirements on retaining the original copyright notices and other things: https://github.com/JFormDesigner/FlatLaf/blob/0291dd5416b6bbf8ea2f3d5e4f7a4f22d6368e36/LICENSE#L89-L128

cwisniew:
@thelsing there is a directory doc/3rd-party-libs that is angood a place as any of you need to include a text file with license file and what it applies to

thelsing:
It seems spotless changed the copyright headers, I'll change them back. We already have multiple files with the apache 2.0 license in doc/3rd-party-libs.

cwisniew (approved):

PR #4662, opened 2024-01-25
Update Map and Token Labels to allow users to define how they look

cwisniew:
fixes ##4660

#### Token Labels
These are not the default colours, which are closer to the current ones, but I have changed them for these screenshots to make them more obvious. These are all set on the local client's side. (These are not the default colours I just needed something that stood out as being very different when testing)

Preferences Dialog

#### Map Labels
Labels that are from pre 1.15 campaigns default to have the background colour (well one of them) from the fake 3D "bubble". These can not be changed client side due to the problems 

CleanShot 2024-02-03 at 15 16 05@2x

CleanShot 2024-02-03 at 15 16 48@2x

CleanShot 2024-01-26 at 05 54 50@2x

I have also added comments to the majority of the files I touched that didn't have comments (adding comments to all the methods and fields the renderer and app preferences probably need an issue to address that so I don't completely annoy whom ever will review this change :)), hoping to set a trend with this but doubt it will :)

### Possible Drawbacks
Some will miss the fake 3D bubble effect; I am sure someone out there prefers fake 3D bubbles over being able to read and configure the look.

### Documentation Notes
See above

### Release Notes
- Added the ability to configure Token labels.
- Added the ability to configure Map labels.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4662)


FullBleed:
It looks like there is extra padding on the top of the text boxes.

cwisniew:
> there is extra padding on the top of the text boxes.

Fonts are going to font :(
If you look at the smaller font labels, they are centred (in fact, the smaller one above has a little too much padding at the bottom), but as they scale, you might get more padding at the top or bottom. I fiddled with it a lot but was unable to get it always to line up correctly in every occurance

FullBleed:
> Fonts are going to font :( If you look at the smaller font labels, they are centred (in fact, the smaller one above has a little too much padding at the bottom), but as they scale, you might get more padding at the top or bottom. I fiddled with it a lot but was unable to get it always to line up correctly in every occurance

Yeah, I see that the smaller the font it starts to drift up and increase the padding below.  Seems like a difference between the padding and font sizes calcs.  It looks like the fonts are using the point system... would using EM or true pixel size normalize things?

Does running 4k exacerbate the issue?

cwisniew:
> > Fonts are going to font :( If you look at the smaller font labels, they are centred (in fact, the smaller one above has a little too much padding at the bottom), but as they scale, you might get more padding at the top or bottom. I fiddled with it a lot but was unable to get it always to line up correctly in every occurance
> 
> Yeah, I see that the smaller the font it starts to drift up and increase the padding below. Seems like a difference between the padding and font sizes calcs. It looks like the fonts are using the point system... would using EM or true pixel size normalize things?
> 
> Does running 4k exacerbate the issue?

Cant use EM its not web browser based, It's just a case of asking the renderer how big the font is and then using that info to draw the box/centre. It is what it is, trying to do it another way will lead to problems on someones display.

kwvanderlinde:
Code changes look good.

As for style, I find the new labels aren't well defined compared to the background, which makes it quite hard for me to focus on them. IMO we should keep a border and a bit of shape with the new flat look, something like this:
image

FullBleed:
> Code changes look good.
> 
> As for style, I find the new labels aren't well defined compared to the background, which makes it quite hard for me to focus on them. IMO we should keep a border and a bit of shape with the new flat look, something like this: 

I like that a lot better, too.

cwisniew:
> Code changes look good.
> 
> As for style, I find the new labels aren't well defined compared to the background, which makes it quite hard for me to focus on them. IMO we should keep a border and a bit of shape with the new flat look, something like this: image

Updated with borders and arcs, both adjustable, can you do another review

kwvanderlinde (approved):

PR #4685, opened 2024-02-08
Mark .form files as inputs to :instrumentForms task

### LINKED ISSUE #4389: [Refactoring]: Avoid deleting .class files every build
### Describe the problem

During development, if I try to run a second MapTool instance via Gradle while another is already running, the first is likely to start generating `java.lang.NoClassDefFoundError` during the build due to certain GUI `.class` files having been removed.

I would like the instrumented and uninstrumented versions of each `.class` file to be kept separate, and the defensive deleting of `.class` files to be removed.

### Expected Benefits

The developer experience will be improved in two ways:
1. We will be able to run multiple MT instances without the builds interfering with each other.
2. Builds will be faster as the `compileJava` task will not be constantly out of date.

Also there won't be a need to maintain a hardcoded list of `.class` files to delete in order to avoid duplicate `$$$getFont$$$` methods.

### Additional Context

_No response_

kwvanderlinde:

Improves on PR #4390 for #4389

### Description of the Change

I've noticed that the current `:instrumentForms` task does not recognize changes to IntelliJ GUI files. This makes dev iteration more clunky than it needs to be when making frequent UI changes. So the situation was only mildly improved by the previous PR.

This PR makes it so the class files are reinstrumented whenever a `.form` file is changed. This way they won't report as up-to-date unless neither Java nor form files are modified.

### Possible Drawbacks

Should be none (build only).

### Documentation Notes

N/A

### Release Notes

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4685)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -180,6 +180,11 @@ def instrumentForms = tasks.register('instrumentForms', Sync) {
     from compileJava.destinationDirectory
     into sourceSets.main.java.destinationDirectory
 
+    def formsSet = objects.sourceDirectorySet('forms', 'IntelliJ form files')
+    formsSet.srcDirs sourceSets.main.java.sourceDirectories
+    formsSet.include '**/*.form'
+    inputs.files(formsSet.files)
+
     doLast {
         ant.echo "Instrumenting IntelliJ GUI forms"
         ant.taskdef(


kwvanderlinde:
This is a part of Gradle I'm not super familiar with. If anyone knows a more idiomatic way to grab the `.form` files as task inputs, I'll update accordingly.

cwisniew (approved):

PR #4753, opened 2024-04-12
Pass zones to templates so they don't have to look them up

### LINKED ISSUE #4524: [Bug]: Burst & Blast templates on a map do not reflect changes to map scale
### Describe the Bug

Burst & Blast templates on a map do not reflect changes to map scale.

If done via [Map -> Edit Map... -> Pixels per Cell] this issue is seen immediately.
If done via [Map -> Adjust grid... -> Grid size] this issue is not seen immediately (will raise a separate bug for that issue!), but the ultimately Burst & Blast templates do not change either.

### To Reproduce

Create a new map and add all 8 template types from the Template Tools toolbar onto the map.
Map -> Edit Map.. 
Change Pixels per Cell to 50.
Click OK
All templates except Blast and Burst types adapt to match the new map scale

### Expected Behaviour

Burst & Blast templates behave the same as other templates and change size to reflect changes to map scale changes.

### Screenshots

Original map with the 8 template types:
image

Same map after changing the Pixels per Cell to 50 (Blast and Burst templates in the bottom-left and bottom-center positions which have not changed size to reflect the new scale, unlike other template types)
image

### MapTool Info

Occurs in 1.13.2 & 1.14.3

### Desktop

_No response_

### Additional Context

_No response_

### LINKED ISSUE #4526: [Bug]: If a map with templates is copied and the copy's [Map -> Edit map… -> Pixels per Cell] is changed, templates sizes change on the original map
### Describe the Bug

If a map with templates is copied and the copy's [Map -> Edit map… -> Pixels per Cell] is changed, templates sizes change on the original map.

### To Reproduce

Create a map and add all 8 template types.
Make a copy of this map.
Change the map scale using [Map -> Edit map… -> Pixels per Cell] to 50.
Select the original map (which should have a Pixels per Cell of 100)
Observe how the original templates have change size (well apart from Burst and Blast templates (which now in this case https://github.com/RPTools/maptool/issues/4524 works in this issue's favour!)

### Expected Behaviour

Changing the grid size on one map should not affect templates on another map (even if they are copies).
Drawing IDs for templates in a campaign should be unique (even if they are copies).

### Screenshots

Original "Grasslands" Map:
image

"Copy of Grasslands" map with  changed Pixels per Cell = 50.   Templates have changed size (well, apart from Burst & Blast templates)
image

Now select original "Grasslands" map, and see how templates have incorrectly changed size 
image

### MapTool Info

Occurs in 1.13.2 & 1.14.3

### Desktop

Windows

### Additional Context

_No response_

### LINKED ISSUE #4527: [Bug]: The curious case of disappearing non-Cone, Blast, and Burst Templates on maps
### Describe the Bug

Templates which are of a type that are not Cone, Blast, nor Burst can sometimes just disappear from a map and from the draw explorer preview.  This seems similar to https://github.com/RPTools/maptool/issues/3693 perhaps but can be instigated in several other ways:

**Scenario 1**
- importing a *.rpmap file that has non-Cone/Blast/Burst templates on it.

**Scenario 2**
- if a map with templates is copied and the copy is then deleted, some templates on original map become invisible (and will also raise an error if [Map -> Adjust Grid…] was open at the time of deletion - maybe covered by https://github.com/RPTools/maptool/issues/4523).

### To Reproduce

**Scenario 1**
- Create a new map (e.g. "Grasslands") and add all 8 template types from the Template Tools.
- Export the map
- Import the map
- Notice now that only three of the 8 different template types are displayed, either in the map itself, or in the Draw Explorer preview.  Templates other than Cone, Blast, and Burst seemingly disappear from the map!
 
**Scenario 2**
- Create a new map (e.g. "Grasslands") and add all 8 template types from the Template Tools.
- Copy the "Grasslands" map.
- Delete the newly copied "Copy of Grasslands" Map.
- Notice now that only three of the 8 different template types are displayed, either in the map itself, or in the Draw Explorer preview.  Templates other than Cone, Blast, and Burst seemingly disappear from the map!

### Expected Behaviour

Templates should remain visible on a map (and in the draw explorer preview) when they haven't been deleted!

### Screenshots

Original Map with templates:
image

Templates put on their rings of invisibility:
image

Error message (if adjust grid... box was open when the copy was deleted):
image

### MapTool Info

Occurs in 1.13.2 & 1.14.3

### Desktop

Windows

### Additional Context

- If you add a non-cone/blast/burst template to the original map after copying it, but before deleting the copied map, these do not disappear from the original map.
- If you make a copy and also a copy of the copy, and delete the copy of the copy, the templates disappear on both the original and the copy maps.  
- If you make a copy and also a copy of the copy, and delete a map that is not the last copied map, then the templates do not disappear.
- Some ways these templates can become visible again I have discovered is if you make another copy of the map, add a new map, or save the campaign and reopen it.

kwvanderlinde:

Fixes #4524
Fixes #4527

### Description of the Change

The main idea with this change is to fix #4527 by passing a `Zone` to templates rather than requiring them to look the zone up themselves. This is done by adding a `Zone` parameter to these `Drawable` methods, and other related methods:
- `draw()`
- `getBounds()`
- `getArea()`

Regular drawings ignore the new parameter, but templates require it in order to scale properly without relying on global state.

There were some remaining zone lookups in `RadiusCellTemplate`, `BlastTemplate`, and `BurstTemplate`. These all delegated to a `ShapeDrawable` as a renderer, and keeping that renderer in sync with the template state required knowledge of the zone when modifying the template. But these renderer did not add any value to these templates -- `RadiusCellTemplate` didn't use its renderer, while `BlastTemplate` and `BurstTemplate` were just as well to draw their own shape. Removing these and the need for maintaining extra state is the resolution to #4524.

With those changes done, templates no longer require their zone ID in any way. So that is now removed from `AbstractTemplate` and the associated DTOs. This simplifies other parts of the code (e.g., `Zone` deserialization) as we don't have to worry about updating template zone IDs.

### Possible Drawbacks

Should be none.

### Documentation Notes

N/A

### Release Notes

- Fixed a bug where burst and blast templates would not scale when changing the the map grid size.
- Fixed a bug where templates would disappear if duplicated to another map and then the duplicate was deleted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4753)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
@@ -251,7 +251,7 @@ public class DrawingFunctions extends AbstractFunction {
     dinfo.addProperty("name", d.getName());
     dinfo.addProperty("layer", el.getDrawable().getLayer().name());
     dinfo.addProperty("type", getDrawbleType(d));
-    dinfo.add("bounds", boundsToJSON(d));
+    dinfo.add("bounds", boundsToJSON(map, d));
     dinfo.addProperty("penColor", paintToString(el.getPen().getPaint()));
     dinfo.addProperty("fillColor", paintToString(el.getPen().getBackgroundPaint()));
     dinfo.addProperty("opacity", el.getPen().getOpacity());
@@ -262,12 +262,12 @@ public class DrawingFunctions extends AbstractFunction {
     return dinfo;
   }
 
-  private JsonObject boundsToJSON(AbstractDrawing d) {
+  private JsonObject boundsToJSON(Zone map, AbstractDrawing d) {
     JsonObject binfo = new JsonObject();
-    binfo.addProperty("x", d.getBounds().x);
-    binfo.addProperty("y", d.getBounds().y);
-    binfo.addProperty("width", d.getBounds().width);
-    binfo.addProperty("height", d.getBounds().height);
+    binfo.addProperty("x", d.getBounds(map).x);
+    binfo.addProperty("y", d.getBounds(map).y);
+    binfo.addProperty("width", d.getBounds(map).width);
+    binfo.addProperty("height", d.getBounds(map).height);
     return binfo;
   }
 
--- a/src/main/java/net/rptools/maptool/client/functions/DrawingMiscFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/DrawingMiscFunctions.java
@@ -108,7 +108,7 @@ public class DrawingMiscFunctions extends DrawingFunctions {
   private JsonArray getCrossedPoints(final Zone map, final DrawnElement de, final String pathStr) {
     List<Map<String, Integer>> pathPoints = convertJSONStringToList(pathStr);
     JsonArray returnPoints = new JsonArray();
-    Area a = de.getDrawable().getArea();
+    Area a = de.getDrawable().getArea(map);
     int cnt = 0;
     Point previousPoint = new Point();
     for (Map<String, Integer> entry : pathPoints) {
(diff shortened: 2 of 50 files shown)


cwisniew (approved):

kwvanderlinde:
Fixed the merge conflicts.

PR #4770, opened 2024-04-30
fix `setTableEntry()` performance

### LINKED ISSUE #4769: [Bug]: setTableEntry extremely slow
### Describe the Bug

When adding 3200 elements to a new table using `setTableEntry()`, MT can consume significant cpu and even more memory (upwards of 48GB!).

### To Reproduce

Execute the follow macro and note the execution time.  Then comment out or delete the `setTableEntry()` and run it again.
```
[h: pattern = "Dex"]
[h: replaceWith = "DEX"]
[h: tableName = "Beasts"]
[h: stringIndex = "40"]

[h, COUNT(3200), CODE:
    {
        [h: stringList = table(tableName, roll.count)]
        [h: value = listGet(stringList, stringIndex, "~")]
        [h: value = replace(value, pattern, replaceWith)]
        [h: stringList = listReplace(stringList, stringIndex, value, "~")]
        [h: setTableEntry(tableName, roll.count, stringList)]            
    }
]
```

### Expected Behaviour

The `setTableEntry()` function should overwrite an existing entry if it's there, or add a new one if it isn't.  (The latter is new functionality that doesn't exist in the current version, so the wiki will need to be updated.)  Unfortunately, I don't see a way to change the return code so that it can indicate whether an entry was added or replaced and still maintain backward compatibility.

### Screenshots

_No response_

### MapTool Info

All

### Desktop

Any

### Additional Context

See [this Discord discussion](https://discord.com/channels/296230822262865920/494598206626463755/1234001822817980460) for the initial report from **Full Bleed**.

Azhrei:
Closes #4769 

### Description of the Change
The original code for `setTableEntry()` was executing an O(n) algorithm to add a single element to the table, causing massive slowdown on larger tables as well as excessive memory churn (OP claimed 32-48GB was required to add 3200 elements to a table).

To correct this, the original `LookupTable` implementation was modified to add a `setTableEntry()` method that will overwrite an existing entry.  The original implementation used `addTableEntry()` to copy the existing table to a new one as overwriting a single entry was not possible.

Also included are refactoring of redundant code blocks into a private method, `checkTableAccess()`, as well as refactoring to move large code blocks for each subfunction into their own private methods.

### Possible Drawbacks
There shouldn't be any.

### Documentation Notes
None.

### Release Notes
- fixed performance issue with `setTableEntry()` when used with large tables

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4770)


FullBleed:
Nice.

I'm going to wait for this to get in before trying to build my nearly 7000 record NPC table (virtually every stat blocked NPC in every published PF1e book through 2020).  As is, I'm not even sure I *could* do it with my 64GB system without significant pain and suffering.

Much obliged.

src/main/java/net/rptools/maptool/model/LookupTable.java
```diff
@@ -55,31 +106,69 @@ public LookupTable(LookupTable table) {
...
   public LookupEntry getLookup() throws ParserException {
     return getLookup(null);
   }
 
-  public String getRoll() {
-    return getDefaultRoll();
-  }
+  //  public String getRoll() {
+  //    return getDefaultRoll();
+  //  }
```
cwisniew, line 152: Please remove function rather than just commenting it out

src/main/java/net/rptools/maptool/model/LookupTable.java
```diff
@@ -55,31 +106,69 @@ public LookupTable(LookupTable table) {
...
-  public void setName(String name) {
+  /**
+   * Set the table name.
+   *
+   * <p>This method should not be called after the table is added to the campaign properties, as the
+   * <code>Map<></code> therein relies on the name for the hash code.
+   *
+   * @param name name of the table
+   */
+  public void setName(@org.jetbrains.annotations.Nullable String name) {
```
cwisniew, line 162: There is a mixture of @Nullable and @org.jetbrains.annotations.Nullable, even the imported one is the same class as the fully qualified one (which I can't tell if it is from the review) it's confusing to anyone reading the code. This is occurring in multiple places in the file

src/main/java/net/rptools/maptool/model/LookupTable.java
```diff
@@ -194,33 +312,40 @@ private int constrainRoll(int val) {
...
+        if (unpicked.length != 0) {
+          Result result = expressionParser.evaluate("d" + unpicked.length);
+          int index = Integer.parseInt(result.getValue().toString()) - 1;
+          unpicked[index].picked = true;
+          return unpicked[index].getValue();
+        }
       } catch (ParserException e) {
-        MapTool.showError("Error getting default roll for Pick Once table ", e);
-        return (NO_PICKS_LEFT);
+        MapTool.showError("Error getting default roll for Pick Once table: " + name, e);
```
cwisniew, line 343: I know it wasn't before but since you are changing the message this is perfect opportunity to put it into the i18n properties file

src/main/java/net/rptools/maptool/client/functions/LookupTableFunction.java
```diff
@@ -84,367 +90,521 @@ public Object childEvaluate(
...
+    } else if ("setTablePickOnce".equalsIgnoreCase(function)) {
+      return setTablePickOnce(function, params);
+    } else if ("getTablePickOnce".equalsIgnoreCase(function)) {
+      return getTablePickOnce(function, params);
+    } else if ("getTablePicksLeft".equalsIgnoreCase(function)) {
+      return getTablePicksLeft(function, params);
+    } else { // if tbl, table, tblImage or tableImage
+      return tbl_and_table(function, params);
+    }
+  }
```
cwisniew, line 139: It's clearer if you convert the function variable to lower (or upper) case and then just use a switch statement, or better yet a switch expression if you are just going to return the value of a function in each else if.

src/main/java/net/rptools/maptool/client/functions/LookupTableFunction.java
```diff
@@ -84,367 +90,521 @@ public Object childEvaluate(
...
 
-      checkTrusted(function);
-      FunctionUtil.checkNumberParam("setTableImage", params, 2, 2);
-      String name = params.get(0).toString();
-      MD5Key asset = FunctionUtil.getAssetKeyFromString(params.get(1).toString());
-      LookupTable lookupTable = getMaptoolTable(name, function);
-      lookupTable.setTableImage(asset);
-      MapTool.serverCommand().updateCampaign(MapTool.getCampaign().getCampaignProperties());
-      return "";
+  private @Nullable Serializable tbl_and_table(String function, List<Object> params)
```
cwisniew, line 141: The name of the function does not follow naming standards

src/main/java/net/rptools/maptool/client/functions/LookupTableFunction.java
```diff
@@ -84,367 +90,521 @@ public Object childEvaluate(
...
-      MapTool.serverCommand().updateCampaign(MapTool.getCampaign().getCampaignProperties());
-      return "";
+    LookupTable lookupTable = checkTableAccess(function, name);
+    LookupEntry result = lookupTable.getLookup(roll);
+    if (result == null) {
+      return null;
+    }
 
-    } else if ("setTableEntry".equalsIgnoreCase(function)) {
+    assert result.getValue() != null;
```
cwisniew, line 157: If this should never happen, please place a comment describing why so people reading the code know its not just a mistake that should be replaced, as otherwise, it should be a more meaningful error reported.

src/main/java/net/rptools/maptool/client/functions/LookupTableFunction.java
```diff
@@ -84,367 +90,521 @@ public Object childEvaluate(
...
-      MapTool.serverCommand().updateCampaign(MapTool.getCampaign().getCampaignProperties());
-      return lookupTable.getPickOnce() ? BigDecimal.ONE : BigDecimal.ZERO;
+  private @NotNull Object getTableEntry(String function, List<Object> params)
+      throws ParserException {
+    FunctionUtil.checkNumberParam(function, params, 2, 2);
+    String name = params.get(0).toString();
+    LookupTable lookupTable = checkTableAccess(name, function);
+    String roll = params.get(1).toString();
+    LookupEntry entry = lookupTable.getLookupDirect(roll);
+    if (entry == null) return "";
```
cwisniew, line 301: Does not follow formatting standards. The same formatting standards issue crops up multiple places in this change

src/main/java/net/rptools/maptool/client/functions/LookupTableFunction.java
```diff
@@ -84,367 +90,521 @@ public Object childEvaluate(
...
-      } else if ("tableImage".equalsIgnoreCase(function)
-          || "tblImage"
-              .equalsIgnoreCase(function)) { // We want the image URI through tblImage or tableImage
+        rowindex++;
+      }
+      MapTool.serverCommand().updateCampaign(MapTool.getCampaign().getCampaignProperties());
+      return counter == rowindex ? new JsonPrimitive(counter) : errorRows;
+    }
+    // FIXME Report error using I18N
+    throw new ParserException("Second parameter must be a JSON Array");
```
cwisniew, line 533: Should be extracted to i18n

cwisniew (changes requested):

src/main/java/net/rptools/maptool/client/functions/LookupTableFunction.java
Jmr3366: @Azhrei I think this may work (so as to maybe get this in to 1.15.1.
`throw new ParserException(
                I18N.getText("macro.function.json.onlyArray"));`

Azhrei:
Hurricane Milton knocked out power, but I'll try to get to this before the end of the weekend, assuming power is back up.  Otherwise, I'll convert it to Draft, which I think takes it out of this queue...

src/main/java/net/rptools/maptool/model/LookupTable.java
Azhrei: These are IntelliJ recommendations.  It apparently thinks they are not equivalent and wouldn't allow one to be used in place of the other without producing a warning.  I wanted to eliminate all warnings, so...  Maybe it's a bug in the version of IntelliJ?  I can update and try again.

src/main/java/net/rptools/maptool/client/functions/LookupTableFunction.java
Azhrei: Agreed.  I'm surprised IntelliJ didn't suggest that.

src/main/java/net/rptools/maptool/client/functions/LookupTableFunction.java
Azhrei: Hm, shouldn't spotlessApply take care of the formatting?  Running spotless was the last commit in the PR.

(Or maybe there are things spotless doesn't correct/change and must be done manually?  If so, is there a list of such things that spotless doesn't fix but it should?  Apparently single-line IF statements is one of them; what about using braces when a true or false block only contains a single statement?)

Azhrei:
It looks like changing the PR to Draft only disables the ability to perform a Review, but doesn't make it disappear out of the PR queue.  (I had hoped that it would only appear to the person(s) assigned to it.)  I likely won't be able to work on it until 10/15 ET, so close it if you feel it's necessary and I'll resubmit later.

src/main/java/net/rptools/maptool/client/functions/LookupTableFunction.java
cwisniew: The Google formatting library that Spotless uses doesn't deal with single statement ifs because of reasons... (to keep Android developers happy as the coding guidelines for Android are different on that one point).

There may be others but they haven't come up before (this one has come up several times)

FullBleed:
How far off of completion is this?  RC's are being built for the next version and it would be nice to test this.

Azhrei:
I'm in crunch mode with a holiday deadline.  I can get back to this next week, Dec 16th week.  If someone else wants to pick up the PR, please do so!

FullBleed:
@kwvanderlinde @cwisniew 

Any chance one of you could take a look at this and push it across the finish line so we can see it in 1.16?

FullBleed:
So this has been killed?

Azhrei:
Accidentally, I guess.  I had to merge in the current develop and there were a lot of conflicts, so I deleted the branch and repushed.  I didn't realize that I'd submitted the PR from that branch.  I'll submit a new PR in the next day or two, as I go through the comments here...

FullBleed:
Hate to keep pinging this but I keep fighting with this issue in my framework and it seems like this PR was "close" to completion...

kwvanderlinde:
@FullBleed I have recently started look at this, and have reviewed this PR to see how things stand.

In terms of existing macros, this PR doesn't appear to make any performance improvements. I.e., as long you're using `setTableEntry()`, `table()`, etc., macros will still be quite slow even with these changes. So I've opened PR #5822 to fix some of that, though there's more improvements that I'd like to make separately, and still quite a lot of overhead even then.

The introduction of the `loadTable()` function would have been the only performance win, and it's significant. I opened FR #5821 for that, and hope to get to it pretty quickly.

Issue #4654, opened 2024-01-21
[Bug]: Lib:Token can no longer be found


ThomasMHenderson:
### Describe the Bug

When starting MapTool 1.14.3, the first time I try to load a campaign file, maptool takes quite a while to do so (15 seconds for a campaign with no contents: campaign file is 15 kB).

- If the campaign does not have a library token, it loads correctly.
- If the campaign does have a library token, it fails to load, with an error message which begins
"net.rptools.maptool.model.library.LibraryNotValidException: Lib:Token can no longer be found."

Subsequent attempts to load campaign files in the same instance of MapTool work perfectly: campaign files load quickly and correctly, whether they have a library token or not,

### To Reproduce

1. Create a new campaign
2. Add a blank token called, say, "Lib:Test"
3. Save the campaign and close maptool
4. Launch maptool and load this campaign

### Expected Behaviour

The campaign file should load.

### MapTool Info

1.14.3

### Desktop

Windows 11 Pro

### Additional Context

[maptool.log](https://github.com/RPTools/maptool/files/14000447/maptool.log)

Rijeclem:
I've been having the same error ever since switching to 1.14 from 1.13. The error doesn't always trigger, it was like 50% chance of triggering so i would just open up the campaign file again. But lately it's been more like 95%. Even if i delete my lib token and save it that way, the error continues this time not even showing error and opening a blank maptool file instead of the campaign file. 

Another thing i found is that the "95% chance of erroring" is retroactive meaning that i save my campaign files regularly while working on my framework and the older versions of my campaign files have the same chance of error. So it isn't a very broken macro triggering, i use oncampaignload for defining functions and i tried to delete all my defined functions and save it that way. Didn't solve the problem.

FullBleed:
I think you're going to have to share the problem campaign file.  If this was a common issue we'd be seeing a lot more reports.

kwvanderlinde:
I believe this is the same underlying issues as #4542.

kwvanderlinde:
Fixed since #4787.

Issue #4542, opened 2023-12-08
[Bug]: Campaign sometimes unloaded immediately after loading.


kwvanderlinde:
### Describe the Bug

Sometimes when I open a campaign, it opens and shows a map for a moment before disappearing and being replaced with a blank campaign. No errors are shown or logged.

### To Reproduce

1. Find a nice campaign in your file explorer.
2. Open the campaign from your file explorer.
3. If you're unlucky, see the campaign disappear after a moment.

### Expected Behaviour

The selected campaign should remain open until I replace it.

### MapTool Info

1.14.1

### Desktop

Linux Mint 21.2

### Additional Context

Not all campaigns have this issue, but I don't know yet what makes the difference. Here is a test campaign of mine that consistently shows the problem for me: [4354-graaljs.cmpgn.renamed-to.zip](https://github.com/RPTools/maptool/files/13607549/4354-graaljs.cmpgn.renamed-to.zip) _In this campaign there is an error that comes up, but it seems to be a symptom of the campaign be unloaded rather than a problem in its own right._

I remember hitting this issue a week or so ago when opening a campaign from within MapTool instead of from the file explorer. But I tried to reproduce it now and couldn't, so I'm not sure anymore if I'm right about that.

kwvanderlinde:
Okay, so the lib:token missing error that I get with the attached campaign is the reason that campaign is _consistently_ dropped. It stops being consistent at 15ebc6caf, I assume due to some differences in error handling (haven't actually looked at it). But the issue exists regardless, that errors is a result of the underlying problem, it's not the cause.

This seems be the rough startup sequence causing the problem:
1. MapTool process starts.
2. EDT queue ->
    1. New default campaign is created and set as current.
    2. New personal server is started.
    3. Client is connected to the personal server. Implicitly generates a `SetCampaignMsg` and sends to the client.
        - Client handles this message async on another thread, then queus and EDT action for the actual logic.
    4. EDT queue ->
        1. Load the requested campaign.
        2. Push `SetCampaignMsg` to the server.
            - Client does not receive as we exclude the sending client for `SetCampaignMsg`.
        3. Synchronously set requested campaign locally on the client.

Since (2.iv) runs on the EDT thread, if we get there before the client message handler queues its EDT action for the first `SetCampaignMsg`, then the default campaign will end up overwriting the requested campaign.

We can make this problem consistent by simulating a delay in the client message handling thread, e.g.:
```diff
[...]

Issue #3756, opened 2022-11-28
[Bug]: Starting server immediately after loading large campaign throws error


aeauseth:
### Describe the Bug

Immediately starting the server (before the map has a chance to refresh) will show the "Server has disconnected" error.  Subsequent "You are already running a server" and "Could not restart personal server" errors will show but @thelsing suggested these are follow-up errors.  This is likely only to occur with larger campains as they can be slow (30+ seconds) to load. #3753 is supposed to improve load times.

Despite these errors, the server does seem to start and clients can connect.

![ServerHasDicsonnected](https://user-images.githubusercontent.com/13549766/204370663-637b4883-2d31-403d-9f3e-967011f78cde.png)

![startserverpng](https://user-images.githubusercontent.com/13549766/204372050-afee19b1-655a-4b57-b997-1a3c5376eb62.png)

### To Reproduce

1. Load a large campaign.
2. Immediately start the server before the map has a chance to refresh.

### Expected Behaviour

Should not be able to start a server if not ready to do so.  Consider one or more of the following:

1. Disabling teh start server menu option until server is ready to start without errors.
2. Alert that the server is not ready to be started.  
3. Progress bar or other delay when server start is requested, delaying until ready to actually start the server.

### MapTool Info

Maptool v1.12.2

### Desktop

windows

### Additional Context

[maptool.log](https://github.com/RPTools/maptool/files/10107063/maptool.log)

FullBleed:
If I may add... 

The "please wait for map to refresh" warning after starting a server has always been something of a mystery to me... Once I start a server I'm always having to tell my player, "Hold on... wait... don't connect yet..."

*twiddle thumbs*

"Ok... go ahead and connect."

It would be nice if the process of starting a server didn't need such a warning.  I'd prefer it if it didn't say "Server Started" until it was actually ready/safe to receive connections.

adventuremagic123:
Yes, I think the entire UI should be locked until rhe campaign is loaded.  A message should be constantly displayed somewhere saying "Campaign loading ...".

Issue #3790, opened 2022-12-26
[Refactoring]: Server initialization


thelsing:
### Describe the problem

Currently server initialization is a mess. We start a server and immediately try to connect to it. If the server is already running the first step fails, but we still try to connect to the server. If we failed the connection because the server is already running we get confusion messages ("Player already exists" if we used the same name the existing server uses") or even the "Please notify your GM of your PIN .." because we connected to some server via webrtc.

I often see the "Failed to start personal server" message when something went wrong with server start.

The huge catch bloc for many different exceptions in the START_SERVER action isn't nice too.  I'd like to remove that.

### The improvement you'd like to see

IMO we need some clean chain of events with onComplete and onError handlers. We already have something similar for establishing client connections. 
We could also use ListenableFutures or something similar.

### Expected Benefits

Clean way to init server, better error handling, code that is easier to understand.

thelsing:
We should also get rid of the "please wait until map reloads" and block the screen until server start and connection to it is done.

FullBleed:
> We should also get rid of the "please wait until map reloads" and block the screen until server start and connection to it is done.

Definitely.  Has always been a bugaboo.

PR #4771, opened 2024-05-05
Refactor client-side and server-side state, split player database responsibilities, and enforce better lifecycle for servers and clients

kwvanderlinde:

Address #4519 , #4542, #4654.

Probably also addresses #3756 but was having a hard time reproducing that exactly.

### Description of the Change

This changes is primarilty a refactoring that better separates client and server concepts and enforces better lifecylces of both. But it touches on a few separate parts of the codebase, so I'll cover them separately below.

#### Client-side changes

Mainly this pulls various parts of the `MapTool` global state into a new `MapToolClient` class. This class represents the client-side of a connection, analogous to `MapToolServer` for the server-side. This new class keeps track of the connection itself, the local player, the current player database, the current campaign, etc. This allows us to clearly associate each one and set up the necessary dependencies between them, rather than having each look up the current global state and hope that whatever they find was created for the same connection.

#### Server-side changes

This reworks personal servers to avoid most possibilities for failure and to generally simplify the case of a personal server. Before, we would run a real server on a random port with its own copy of the campaign, and would have to keep that campaign in sync with the client by sending it network messages. Now we have a trivial `PersonalServer` class that does not need its own campaign copy, and therefore does not need to respond to any messages. This avoids several potential failure states, and makes personal server creation an instant process, which in turn avoids timing issues that can occur when loading large campaigns.

#### Player database

At the intersection of the above changes was the `PlayerDatabase`, and more importantly, the `PlayerDatabaseFactory`. Before, we kept track of the current database in `PlayerDatabaseFactory` itself. Now `MapToolClient` keeps its own `PlayerDatabase`, and `MapToolServer` keeps its own `PlayerDatabase` (for local servers they are the same database). This means `PlayerDatabaseFactory` is now just a factory and not a vehicle for application state, which in turn allowed us to cut out a lot of code dedicated to notifying and keeping track of the databases.

`PlayerDatabase` itself was reworked a bit as well. Some methods were specific to server-side databases, so those have been moved into their own `ServerSidePlayerDatabase`. This simplifies `LocalPlayerDatabase` (which is purely client-side) and means we don't have to try and interpret some of these server-side methods in a client-side way.

#### Lifecylce improvements

With things a bit more separated, we have a clearer process for starting servers, connecting to servers, and disconnecting.

When starting a server, we now create the server (personal or other local server), create a corresponding client to connect to it, then start the server and then set the client. When connecting to a remote server, we null out the local server and create a client to connect to the remote serve, then start it. Doing it this way means the client and the server will be set together, rather than having periods of time where one or the other isn't set, or might be set to old state.

On the disconnecting side, we have altered the relationship between `MapTool.stopServer()` and `MapTool.disconnect()`:
- `disconnect()` only deals with cleaning up client-side connection state, including adjustments to the UI as needed. It no longer does any server teardown such as UPnP port closure or accouncer shutdown.
- `stopServer()` is where server state is cleaned up (if a local server is running). It is a mirror imge of `startServer()`, cleaning up everything that `startServer()` set up. `stopServer()` no longer cleans up connection state through `disconnect()`.

This better allows us to handle both expected and unexpected disconnects, and avoids the redundant "You have disconnected." messages that show up when disconnecting for a server.

### Release Notes

- Fixed a bug where starting a server on a used port when bring MapTool in an inescapable bad state.
- Fixed a bug where campaigns are sometimes unloaded immediately after MapTool starts.
- Fixed a bug where Lib:Tokens sometimes disapppear after MapTool starts.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4771)


thelsing:
Just wanted to review this, but still draft. :)
Redoing the server start code was on my todo list, too. [I wanted to make it more async with CompleteableFuture because webrtc connections are established though event.](https://github.com/thelsing/maptool/tree/feature-serverconnect) 
Maybe something like this could be incorporated?

kwvanderlinde:
> Just wanted to review this, but still draft. :)

Undrafted now :slightly_smiling_face:

> Redoing the server start code was on my todo list, too. [I wanted to make it more async with CompleteableFuture because webrtc connections are established though event.](https://github.com/thelsing/maptool/tree/feature-serverconnect?rgh-link-date=2024-05-06T18%3A52%3A07Z)
> Maybe something like this could be incorporated?

I am generally in favour of what I'm seeing on your branch, but would like to keep this PR more focused (it's already pretty large). I actually have a lot of other potential changes written up for `clientserver` code, but decided to cut them out of this PR just to keep it managable.

Maybe we can talk about this further somewhere else (#3790 I guess), but I also have some ideas on how servers can be reworked that in some ways lines up with your branch there, but in other ways is different. The main thing in my mind is that it should be trivial for a client to connect to its own local server, regardless of whatever connection mechanism remote clients have to use. A separate concern to that is whether the server starts up successfully so that remote clients can actually connect.

cwisniew (approved):

cwisniew:
> Just wanted to review this, but still draft. :) Redoing the server start code was on my todo list, too. [I wanted to make it more async with CompleteableFuture because webrtc connections are established though event.](https://github.com/thelsing/maptool/tree/feature-serverconnect?rgh-link-date=2024-05-06T18%3A52%3A07Z) Maybe something like this could be incorporated?

I agree with @kwvanderlinde lets do it in a separate PR

PR #4777, opened 2024-05-11
More client lifecycle improvements

### LINKED ISSUE #4776: [Bug]: Ill-timed heartbearts cause handshake failures
### Describe the Bug

If a client is in the middle of a handhsake, and the time comes for a heartbeat to be sent to the server, the handshake will be corrupted by the heartbeat and the connection will fail.

### To Reproduce

1. Start a server on one machine
2. On another machine, run MapTool under a debugger.
    1. Set a breakpoint at the top of `ClientHandshake.handleMessage`
    2. Connect to the server
    3. Wait on the breakpoint for more than 20 seconds.
    4. Remove the breakpoint and continue.
3. If all went well, the server will have received reported a `MESSAGETYPE_NOT_SET` and the client will be shown an error about incorrect passwords.

### Expected Behaviour

Heartbeats should not be sent to the server during handshake. Handshakes complete very quickly after establishing a connection, and so there shouldn't even be an opportunity to lose the server unless something very drastic is happening.

### Screenshots

_No response_

### MapTool Info

develop

### Desktop

Linux Mint 21.3

### Additional Context

N/A

kwvanderlinde:

Fixes #4776, follow up to PR #4771

### Description of the Change

The main change here is the bug fix for #4776: `ServerCommand` will now only send messages if the client has finished its handshake. This is done by giving `MapToolClient` an explicit set of states: `New` when first created; `Started` after the handshake is started on a `New` client;`Connected` when the handshake completes for a `Started` client; `Closed` whenever `close()` is called regardless of previous state.

Second is that handshakes are future-based now which really cleans up the amount of code dedicated to observer them and getting results.

Third is some minor fixes and clean up related to the player database UI.

### Possible Drawbacks

Hopefully none ...

### Documentation Notes

N/A

### Release Notes

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4777)


kwvanderlinde:
I've pared this down to the more essential stuff. I'll do a follow-up PR afteward that really streamlines server init in general.

src/main/java/net/rptools/maptool/server/ServerHandshake.java
```diff
@@ -205,16 +162,17 @@ private synchronized String getEasyConnectName() {
...
     // with invalid handshake.
   }
 
-  private void sendMessage(HandshakeMsg message) {
+  private void sendMessage(State newState, HandshakeMsg message) {
+    setCurrentState(newState);
+
     var msgType = message.getMessageTypeCase();
-    log.info("Server sent to " + connection.getId() + ": " + msgType);
+    log.debug("Sevver sent to {}: {}", connection.getId(), msgType);
```
thelsing: typo

thelsing (approved):
LGTM beside the one typo.

PR #4802, opened 2024-05-26
Do not run the InfoTextSwingWorker on the EDT thread

### LINKED ISSUE #4801: [Bug]: Opening the Gather Debug Information dialog hangs the application
### Describe the Bug

When using Help > Gather Debug Information..., MapTool becomes unresponsive until the dialog is populated.

Worse yet, on 1.15.0-rc.3, the dialog never populates, meaning MapTool is permanently hanged when using ths dialog.

### To Reproduce

1. Click on Help > Gather Debug Information.
2. Observe a blank dialog open
3. Attempt to pan the map while the dialog is still blank. Nothing will happen.
4. Attempt to pan the map once the dialog is populated. This time it will work.

### Expected Behaviour

The application should remain responsive while the debug info is gathered.

### Screenshots

_No response_

### MapTool Info

1.13.2, 1.15.0-rc.3

### Desktop

Linux Mint 21.3

### Additional Context

_No response_

kwvanderlinde:

Fixes #4801

### Description of the Change

This changes the data gathering task for Gather Debug Information... to run on a background thread rather than the EDT thread. The dialog as a result now shows the text "Gathering information, please wait..." until the dialog is populated, rather than being blank. The dialog also no longer hangs indefinitely.

The reason we previously got away with running that task on the EDT thread is that `SwingWorker` would always run the `done()` method after the internal `FutureTask<>` completes, so the `get()` method would have a result to return. But since Java 21, it runs `done()` as part of the `FutureTask<>` callback if running on the EDT thread, i.e., it runs `done()` before the task completes. The `get()` call therefore blocks as it has no result yet, and the task is blocked from completing and providing those results.

I audited the rest of our `SwingWorker` tasks, but they are all run via `.execute()` and thus are safe from this issue. Only one task (`RenderPathWorker`) is run in a dfiferent manner, but it is on its own non-EDT executor. One unnecessary usage of that worker was found and removed though.

### Possible Drawbacks

None

### Documentation Notes

N/A

### Release Notes

- Fixed a bug where Gather Debug Information... would permanently hang MapTool.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4802)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/SysInfoDialog.java
+++ b/src/main/java/net/rptools/maptool/client/ui/SysInfoDialog.java
@@ -47,7 +47,7 @@ public class SysInfoDialog {
     infoTextArea.setWrapStyleWord(true);
     infoTextArea.setFont(new Font("Monospaced", Font.PLAIN, 13));
     infoTextArea.setText(I18N.getText("action.gatherDebugInfoWait"));
-    EventQueue.invokeLater(new InfoTextSwingWorker());
+    new InfoTextSwingWorker().execute();
 
     JScrollPane scrollPane = new JScrollPane(infoTextArea);
     scrollPane.setHorizontalScrollBarPolicy(31);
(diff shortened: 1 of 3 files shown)


kwvanderlinde:
Targetting the 1.15 release with this one so bugs can be properly reported for that version.

cwisniew (approved):

PR #4826, opened 2024-06-06
Fix the problems with out crowin updates

cwisniew:
The checks for crowdin changes are failing due to test and key differences.

### Description of the Change
Need to create this PR to bring things back into line before i18n feeds will sync up again 

### Possible Drawbacks
There will be some files with problems due to file name case, but this wont properly be able to be fixed until after this is merged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4826)


bubblobill (approved):
It's passing the tests and has localised language files. I say do it before it breaks again.

PR #4838, opened 2024-06-15
VBL Excluding

### LINKED ISSUE #4836: [Feature]: VBL Immunity/Exclusion
### Describe the Problem

Vision is purely an on/off - if vision is on for the map, tokens without vison can see nothing. There's no method for preventing certain types of VBL from effecting the vision for a certain token.

### The Solution you'd like

Tokens can become immune to/exclude specific VBL types/areas from their vision processing.

Do the four kinds of Map VBL effect this tokens visibility. (Wall, Hill, Pit, and Cover) - Perhaps this is better configured in the vision/sight configuration?

Configure which tokens have their VBL ignored - allowing tokens to see through other tokens - desirable for when you want the token VBL for determining cover from other tokens - but want PCs immune to VBL of fellow PCs.

Using macros to enable/disable specific VBL depending on turn order - not great for performance, but does kind of work.

### Additional Context

Not the most necessary feature, but helps with a few edge cases I think I might be coming on soon/eventually.

ColdAnkles:
closes #4836

### Description of the Change
Unfinished Implementation of VBL Excluding per token. Needs review by someone who knows vision systems better

### Possible Drawbacks
Forces frequent regeneration of Vision Areas - performance detriment

### Documentation Notes
Adds macro functions to get and set map and token VBL immunity on a token.
When set, that token's vision is not effected by that map/token topology.

### Release Notes
Added ability for token to ignore specific topology.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4838)


cwisniew:
@kwvanderlinde  do you wish to have a go at reviewing this as you have had your head in the code in this area more recently than I have? If you don't have time I can do it.

kwvanderlinde:
Yeah I'll have a look. Though I think this could use some discussion at the feature level - it's related to a bunch of other FRs that are better backed by use cases, so I'd like to see how it works together with some of those other ideas.

I'll try to get some of that discussion going soon.

ColdAnkles:
At this point, this is probably the best I can do with this. I think it's usable (or salvageable) as is - aside from the ugly GUI bits.

ColdAnkles:
god I hate merge conflicts

ColdAnkles:
@kwvanderlinde - you might need to give me some hints on how to mesh my intention with your nice new features.
I think I'll need to take the exclusions as from `getTokenVisibleArea()` and pass them to `zone.prepareNodedTopologies()` - but `zone.getMasks()` does both regular and token topology, and excluding seems to be a single UID to exclude one token only.

kwvanderlinde:
As I mentioned in #4836, I don't believe filtering on VBL type is the right way to go. Instead we should be aiming to attach real semantics to vision and vision blocking so that users can control the behaviour based on intent.

PR #4858, opened 2024-07-14
Remove property files with same name differing case to let crowdin to send only one set

cwisniew:
Remove i18n files that have same name but different case so that crowdin can send only the correct case
* src/main/resources/net/rptools/maptool/language/i18n_en_au.properties
* src/main/resources/net/rptools/maptool/language/i18n_en_gb.properties
* src/main/resources/net/rptools/maptool/language/i18n_pt_br.properties

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4858)


bubblobill (approved):
Nothing dramatic, just housekeeping

Issue #4849, opened 2024-06-29
[Feature]: Configure token size categories


Baaaaaz:
### Describe the Problem

MapTool has no functionality enabling a GM to customise or configure token size categories.

* currently token size names for a square/iso grid are seemingly named after D&D 3.0 sizes, gridless uses numbers, and hex uses a fraction or a name.  Ref: [https://wiki.rptools.info/index.php/token_Size](https://wiki.rptools.info/index.php/token_Size)
* game systems are likely to use different size names or have a different number of size categories entirely but MapTool does not support size customisation leading to a disconnect between MapTool sizes and the games system sizes.

### The Solution you'd like
(In priority order 1= highest, 4=lowest)

1. Similar to other items under Campaign Properties, as a GM I'd like to be able to manually configure, save, export, and import token sizes to suit my game system of choice, such as the ability to:
* add new sizes, e.g. adding a grid/iso size 'Big' with a scale of 1.5 between 'Medium' and 'Large'
* rename a size rather than being forced to use numbers (gridless), D&D 3.0 size names (grid/iso), or a mixture of fractions/names (hex) in both MapTool GUI and MT Script.
* set a display name to be shown in the MapTool token context menu size item (if different to the size name)
* remove/hide redundant sizes so they do not show under the MapTool token context menu size item
* change a size scale, e.g. Colossal from 1.0 (6x6) to 1.0 (8x8), or Fine from 0.5 to 0.25
* delete sizes

2. Adding campaign size categories, names, display name, values to `getInfo("campaign")` results for each grid type.

3. When importing predefined Campaign Properties (e.g. Dungeons and Dragons: 5th Edition), set the sizes correctly for that system.

4. MT Script functions to support automating relevant features above in point 1.

### Alternatives that you've considered.

1. Having some data mapping residing somewhere in the campaign file that can translate the game system size into MapTool size, but then the sizes show in the MapTool GUI still use MapTool sizes. 

### Additional Context

D&D 3.0 came out in 2000, so MapTool needs to get with the times and size up supporting sizes from other games systems! e.g.

* [D&D 3rd Edition Sizes](https://www.dandwiki.com/wiki/3e_SRD:Movement,_Position,_and_Distance#Big_and_Little_Creatures_in_Combat)
* [D&D 3.5 Edition Sizes](https://www.dandwiki.com/wiki/SRD:Table_of_Creature_Size_and_Scale)
[...]

Issue #4856, opened 2024-07-09
[Bug]: Token footprint is kept when changing to native size


kwvanderlinde:
### Describe the Bug

If a _Medium_ token (1x1 cell) is changed to _Native_ size, the footprint stays as one cell. If a _Gargantuan_ token (4x4 cells) is changed to _Native_ size, its footprint stays as 4x4 cells despite the user seeing no reference to a _Gargantuan_ size.

### To Reproduce

1. Drag two tokens onto a map with square grid 
2. Change one token to size _Medium_ and the other to size _Gargantuan_
3. Change both tokens to size _Native_.
4. Drag each token. Notice the originally-medium token still occupies one cell, while the originally-Gargantuan token occupies 16 cells

### Expected Behaviour

The token footprint should be based on its actual size. If the token visibly covers 9 cells, that should be its footprint. Alternatively, natively-sized tokens could simply occupy one cell at its center.

### Screenshots

This shows the two apparently-identical tokens with their differing footprints:
![image](https://github.com/RPTools/maptool/assets/7492219/68ff0e8c-6ec1-4141-b77b-e6ce1413c744)

### MapTool Info

1.14.3

### Desktop

Linux Mint 21.3

bubblobill:
- Native size footprints should probably be the size of the drawn image instead of one cell.
- I can actually see benefits in this if it was something you could do deliberately. e.g. A token has an effect that covers a certain footprint. You could then use macros to apply the effect on other tokens within the footprint.
- We really need a footprint editor dialogue

bubblobill:
No. What you see is not necessarily related to the occupied space. 
WRT retention of footprint. That should just need `isSnapToScale` being set to false when it is changed to free-size
Between footprints changing, where do we need to update the occupied cells cache?

Issue #4830, opened 2024-06-11
[Bug]: Hex grid lightting broken if grid is adjusted before any lights are displayed


kwvanderlinde:
### Describe the Bug

When MapTool is first opened, if the map is changed to a hex grid and the grid is adjusted before showing any grid lights, then grid lights will evermore not have the correct shape.

### To Reproduce

1. Open MapTool
2. Edit the map and change to a vertical hex
3. Adjust the grid, setting the size to 100 and the 2nd dimension to 400 (which will change to ~200 due to #2840).
4. Edit campaign properties and add this light: `grid: grid 20#000000`
5. Add a token to the map
6. Attach the `grid` light to the token
7. See the strange light shape that occurs.

### Expected Behaviour

The grid lighting should be a simple area with no holes.

### Screenshots

![image](https://github.com/RPTools/maptool/assets/7492219/c16caa7c-00ea-4110-a44b-9954d039e6f4)

### MapTool Info

1.13.2, develop

### Desktop

Linux Mint 21.3

PR #4886, opened 2024-09-01
Token Footprint Functions and Campaign Properties

ColdAnkles:
Basic Implementation of #4849, submitted for review

### Description of the Change
Moves Token Footprints from being Static and Part of Campaign Properties.
MTScript functions for getting and setting Footprints

### Possible Drawbacks
Initial naive implementation, possibly with some weirdness from my dev environment

### Documentation Notes
getTokenFootprints()
getTokenFootprints(gridType)
getTokenFootprints(gridType, footprintName)

getGridTypes()

getFootprintNames()
getFootprintNames(gridType)

removeTokenFootprint(gridType, footprintName)

setTokenFootprint(gridType, footprintName, JSONFootPrintData)

resetFootprintsToDefault()

### Release Notes
Custom Token Footprint Support

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4886)


src/main/java/net/rptools/maptool/model/CampaignProperties.java
```diff
@@ -163,6 +161,11 @@ public void mergeInto(CampaignProperties properties) {
     properties.tokenStates.putAll(tokenStates);
     properties.tokenBars.putAll(tokenBars);
     properties.defaultTokenPropertyType = defaultTokenPropertyType;
+    if (properties.gridFootprints != null) {
+      properties.gridFootprints.putAll(gridFootprints);
+    } else {
+      properties.gridFootprints = gridFootprints;
+    }
```
cwisniew: Should create a new HashMap instead of assigning so that if its modified after the call it wont cause issues.

src/main/java/net/rptools/maptool/model/CampaignProperties.java
```diff
@@ -144,6 +139,9 @@ public CampaignProperties(CampaignProperties properties) {
       characterSheets.put(type, properties.characterSheets.get(type));
     }
     defaultTokenPropertyType = properties.defaultTokenPropertyType;
+    if (properties.gridFootprints != null) {
+      gridFootprints = properties.gridFootprints;
+    }
```
cwisniew: Better to clear and add all rather than assigning a passed in hashmap

src/main/java/net/rptools/maptool/model/CampaignProperties.java
```diff
@@ -352,6 +356,66 @@ private void initTokenTypeMap() {
     tokenTypeMap.put(getDefaultTokenPropertyType(), list);
   }
 
+  // Undesired reimplementation from Grid.java
```
cwisniew: What is desired then? :)

src/main/java/net/rptools/maptool/model/TokenFootprint.java
```diff
@@ -76,15 +83,33 @@ public String toString() {
...
+      jsonPoint.addProperty("y", currentCell.y);
+      occupiedString.add(jsonPoint);
+    }
+    jsonRep.addProperty("name", name);
+    jsonRep.add("cells", occupiedString);
+    jsonRep.addProperty("scale", scale);
+    jsonRep.addProperty("localizedName", localizedName);
+    return jsonRep;
+  }
+
```
cwisniew: This should be part of the macro function code not here

src/main/java/net/rptools/maptool/model/TokenFootprint.java
```diff
@@ -76,15 +83,33 @@ public String toString() {
...
     translatorList.add(translator);
   }
 
   public Set<CellPoint> getOccupiedCells(CellPoint centerPoint) {
     Set<CellPoint> occupiedSet = new HashSet<CellPoint>();
 
-    // Implied
-    occupiedSet.add(centerPoint);
+    // not implied any more, explicitly part of footprint data
+    // occupiedSet.add(centerPoint);
```
cwisniew: just remove it, having such comments just makes things more confusing :)

cwisniew (changes requested):
This really needs a UI, functionality like this shouldn't be locked behind macros, also going by past experience macro only functions don't get tested until its too late to change anything if its broken,

cwisniew (changes requested):
This really needs a UI, functionality like this shouldn't be locked behind macros, also going by past experience macro only functions don't get tested until its too late to change anything if its broken,

cwisniew:
The tests this fails will also need to be fixed up

src/main/java/net/rptools/maptool/model/TokenFootprint.java
ColdAnkles: I don't mind either way, but would like to know why

cwisniew: Since the functionality is part of the macro and not the footprint I prefer it in the macro code. It's just to stop macro code infiltrating other classes.

ColdAnkles:
Agreed on the UI front, but it's not something I'm very good at, so since @bubblobill offered to do the GUI in #4849 if I got the backend, this PR is solely for backend enablement of the incoming GUI portion.
The macro parts were mostly because I'm capable of implementing them, and I needed something to help test with. (and they'd be desired functionality anyway).
I also added a quick and dirty mitigation for #4856 - though would like something better (as in generating a custom footprint that appears to cover the approximate area). Needs more thought.

On the tests failing: I've taken a look and can't see anything obvious between what I've changed and which tests fail

src/main/java/net/rptools/maptool/model/TokenFootprint.java
```diff
@@ -68,6 +70,9 @@ public TokenFootPrintDto toDto() {
     dto.setId(id.toString());
     dto.setIsDefault(isDefault);
     dto.setScale(scale);
+    if (localizedName != null) {
+      dto.setLocalizedName(localizedName);
+    }
```
cwisniew, line 76: The localised name probably shouldn't be transferred or saved, since it could be different on different clients

cwisniew (changes requested):

cwisniew:
> 
> On the tests failing: I've taken a look and can't see anything obvious between what I've changed and which tests fail

Taking a quick look my guess would be the calls to CampaignProperties.loadFootprints(), the XML probably not loading correctly as its getting xstream errors. Try remove the localizedName in the classe (which shouldn't be saved anyway) and see if that fixes the issue. Or if you need that variable put transient in front of it so it is not persisted

you can run the tests locally with .\gradlew.bat test. (or ide) to test

bubblobill:
I noticed you didn't include the grid type "isometric hex". One day someone is going to implement it beyond including it in the code add as an option.  :)

cwisniew:
> I noticed you didn't include the grid type "isometric hex". One day someone is going to implement it beyond including it in the code add as an option. :)

Thanks for volunteering... Just know I believe in you @bubblobill  :)

ColdAnkles:
> > On the tests failing: I've taken a look and can't see anything obvious between what I've changed and which tests fail
> 
> Taking a quick look my guess would be the calls to CampaignProperties.loadFootprints(), the XML probably not loading correctly as its getting xstream errors. Try remove the localizedName in the classe (which shouldn't be saved anyway) and see if that fixes the issue. Or if you need that variable put transient in front of it so it is not persisted
> 
> you can run the tests locally with .\gradlew.bat test. (or ide) to test

Okay, its definitely the footprint loading - removing the localizedName variable or making it transient didn't make a difference, but omitting all the calls to footprint loading from the initFootprints step did. Not sure what to try next.

At any rate: the reason for my choices with localizedName existing separate to name is to allow for a framework to define sizes/footprints and have the display/localized name overwritten freely by someone else without effecting ability to address footprints by their "true" name - and also not having to make said localization dependant on MapTool - though it does work on the assumption that servers and clients are working in the same locale.
Also it doesn't work as intended yet.

cwisniew:
> Okay, its definitely the footprint loading - removing the localizedName variable or making it transient didn't make a difference, but omitting all the calls to footprint loading from the initFootprints step did. Not sure what to try next.
> 
There is probably no good reason to have default footprints in external serialised files in any case, probably just make classes with the default values


> At any rate: the reason for my choices with localizedName existing separate to name is to allow for a framework to define sizes/footprints and have the display/localized name overwritten freely by someone else without effecting ability to address footprints by their "true" name - and also not having to make said localization dependant on MapTool - though it does work on the assumption that servers and clients are working in the same locale. Also it doesn't work as intended yet.

This is not a hreat assumption though there will be cases where client is  different from server. Really it should be something that is localised within MT and users don't need to localise themselves

ColdAnkles:
> This is not a hreat assumption though there will be cases where client is different from server. Really it should be something that is localised within MT and users don't need to localise themselves

including localizations for every possible word/phrase people might choose to name a token size?

cwisniew:
> > This is not a hreat assumption though there will be cases where client is different from server. Really it should be something that is localised within MT and users don't need to localise themselves
> 
> including localizations for every possible word/phrase people might choose to name a token size?

Ah gotchya, it's a super supplied name? Then that is fine

ColdAnkles:
Okay, so excising the XML based defaults and replacing with code based works fine and tests pass.

However I've noticed that some footprints on hex grids are busted, and I can't see how my changes effect this

bubblobill:
There are about 5 principal methods of referencing grids, each with a zillion variants. The one we use is probably the least useful for compatibility and interoperability between grid types with the possible exception of polar coordinates. 
Every time I considered doing what you are doing, I baulked because the xml makes no sense.

ColdAnkles:
Here's what I'm seeing re: Hex Grid weirdness
image

Standard `Large` footprint on both hex grids do this kind of thing, yet the coordinates in the footprint are unchanged

bubblobill:
Didn't someone recently fix something relating to footprints? Maybe in #4830?

ColdAnkles:
It's definitely something related to my changes - `develop` and `1.14.3` don't have the issue

bubblobill:
Swap your vertical and horizontal hex logic.

ColdAnkles:
> Swap your vertical and horizontal hex logic.

I'm not following (unless you meant to swap `makeVertHex` and `makeHorizHex` in which case I did, and had no effect.)

bubblobill:
The hex that is out of place could have resulted from being calculated using the wrong grid type. Was hoping it was an easy fix due to a bad offset calc.

ColdAnkles:
as best I can tell, for both grids, every second row of cells has off-by-one coordinates
and I'm a little lost on where to go find the source of the issue

bubblobill:
This is correct. I thought your problem may be the you are getting the offset on the wrong row/column. 
Try incrementing it by one and see if the shape ends up correct. You may simply be getting even offsets on odd rows and vice versa.

bubblobill:
Go here and follow the link to redBlobGames. Read his stuff on hex grids. 

https://github.com/RPTools/maptool/wiki/On-Maps,-Grids,-Objects,-and-Location

ColdAnkles:
Assuming my understanding of hex grids is functional, here's a better diagram from what I'm observing:
image
What's up with `(-1,-1)` under actual? That entire row is off by one.

Baaaaaz:
May need to adjust your expectation! ;)  

MapTool seems to actually use the following coordinates for pointy-top hex grids: 
image

and for flat-top hex grids:
image

These align with the hex "odd-r" and "odd-q" offset  coordinate systems mentioned here:
https://www.redblobgames.com/grids/hexagons/#coordinates-offset

ColdAnkles:
oh barf, I hate that
Do we have any reason to keep it that way?
(and thanks and kudos for the really clear and helpful diagram)

thelsing:
I think the way it is quite logical.

Baaaaaz:
> oh barf, I hate that Do we have any reason to keep it that way? (and thanks and kudos for the really clear and helpful diagram)

No worries about the diagram - I just expanded on your better diagram idea! 👍 

Maybe a separate FR to introduce a whole new (optional) alternative to offset hex grid coordinate referencing, but it is possible to translate between offset and axial coordinates: https://www.redblobgames.com/grids/hexagons/#conversions-offset

ColdAnkles:
The problem now is that footprints are defined as a number of offsets from a reference point - the standard "Large" size token is comprised of the cells `(0,0), (1,0), (0,1)` which works A-OK when the reference point is also `(0,0)` since adding offsets to that results in identical set. But when the reference point is `(-1,0)` you end up with the set `(-1,0), (0,0), (-1,1)` instead of `(-1,0), (0,1), (-1,1)` - two entirely different shapes.
I have no idea how it was working previously and why it's different now, since I'm using the same list of offsets as were originally used.
Is it worth telling the `getOccupiedCells` function what grid you're on and attempting to handle the offset there, or something else?

bubblobill:
If you read the redblob stuff about coordinate systems you will see that there are much better systems than the odd-r/q crap we went with. 

Since I have been advocating we upgrade to cubic ever since I read about it I think this is a great opportunity to implement it. You wouldn't need a new PR to change the hard-coded xml to fit a sensible system, and adding a third axis will make future stuff much easier. Calculating radii and rings and other shapes programatically is much easier with a decent coordinate system. 
He even provides code for conversion between systems so a tiny bit of infill code can keep things running that don't rely on it. 
You can do all your operations on sensible stuff and just spit out the ugly for every other function to use.

bubblobill:
> oh barf, I hate that
> 
> Do we have any reason to keep it that way?
> 
> (and thanks and kudos for the really clear and helpful diagram)

No reason. Currently every size is just a bigger circle. You could throw out the existing stuff and generate them on startup in a couple of ms.

bubblobill:
> I think the way it is quite logical. 

Just not optimal.

thelsing:
I'm all for using a proper  coordinate system (y-up) instead of the pixel-based (0,0) is top left. But this should be done everywhere and not just in a small part like tokenfootprint.

bubblobill:
> I'm all for using a proper coordinate system (y-up) instead of the pixel-based (0,0) is top left. But this should be done everywhere and not just in a small part like tokenfootprint.

Gotta start somewhere, and nobody wants to rewrite everything all at once.

bubblobill:
So... Mr Ankles. You wish to grant supreme power to users to determine shapes.
Now, as we all know by now, with great power comes much irresponsibility.
One wonders what Mr @kwvanderlinde will want to know about things like this
image

1. The shape is not conveniently convex.
2. It has no evident centre other than the point all cells are relative to.
3. It wants to have eyeballs in the blue hex.
4. It has a glow in the dark tail.

I propose footprints contain the following data;

1. The cell from which vision is determined.
2. The cell from which light is calculated.
3. A point (not cell) on which rotation occurs, assumed to be the origin cell centre unless defined.

Just saying, this GUI is getting out-of-hand.

image

ColdAnkles:
> I propose footprints contain the following data;
> 
>     1. The cell from which vision is determined.
> 
>     2. The cell from which light is calculated.
> 
>     3. A point (not cell) on which rotation occurs, assumed to be the origin cell centre unless defined.

I did some quick tests - looks like they originate from the point at the centre of the relative x/y rectangle bounding box.

Fundamentally I agree, but unsure if I want to get into extending footprints with that functionality at this point.

At any rate, I solved my offset grid issues and will open a new issue proposing changes to hex grid coordinate systems.

bubblobill:
Not sure its fixed yet. I don't know if I'm doing it right yet, but this isn't the desired result
image

bubblobill:
Minor annoyance - you cannot set the canonical name for a footprint, only the localised name.
This is fine for people doing it for themselves, but annoying for wanting to edit the existing cases for broader use. Going to have to replace it with a new one which is extra fiddling.

ColdAnkles:
That's something of a consequence of my choice to use the canonical name as the key for the map as well - the only way to do it would be to remove the existing named footprint and replace with an almost identical footprint.

bubblobill:
Yup, just extra fiddling.
GUI is nearly there (old pic).
image
I have it on your fork so I am wary of pushing it until it is working otherwise you will need to exclude it from your commits.

bubblobill:
Footprint functions: 

1. People are going to ask for "Isometric" so you should include it, even if it only returns square.
2. You use mod in your offset translator calculations. Unwise as there is always the potential for a negative number to screw things up. Probably better to use bitwise And. (I know it's copy pasta, but that doesn't mean the original can't be improved on).

ColdAnkles:
Some Feedback:
If you delete the only footprint marked as a default, you start getting errors -> set the first footprint in the list to be default if no defaults exist?
There's something I don't understand happening with the naming of footprints and the appending of underscores to them and this causing other footprints to disappear.
Its not immediately clear how the cell selection works; I worked out it was Select A, then B sets cells on A->B path as footprint - maybe use blank button to swap between path select and single cell select?
Cancel button didn't work.

bubblobill:
> Some Feedback: If you delete the only footprint marked as a default, you start getting errors -> set the first footprint in the list to be default if no defaults exist? 

It is already supposed to do that

> There's something I don't understand happening with the naming of footprints and the appending of underscores 

Supposed to ensure unique names. Been redone a few times and I got distracted so it's probably between two different implementations

> Cancel button didn't work.

Not sure if it's connected yet, but `escape` works.

Functionality is supposed to be:

- Typing in the text field: Changes the name of the currently selected footprint
- Add: Duplicates the current footprint with an amended name
- Revert: Goes back to the original footprint if it exists. Not intended to work on new footprints but it might.
- Clicking: Long story
Firstly. The editing panel has gone through many iterations, in it's present form it accepts three arguments, a grid type name (nullable), a footprint, and a cell set (also nullable)(overrides getOccupiedCells()).
- creates an instance of each type of grid
- creates a new zone and zone renderer
- switches between grids according to passed arguments
- paints cell highlights for each of the footprint's cell set
- paints a waypoint marker on the origin cell that everything is relative to
- creates a ZoneWalker

The clicking: 
- mouselistener gets cell under mouse from the zone renderer, sends it to modifyFootprintCells
- any clicks on cell(0,0) are rejected
- any other cell is then added to or removed from the current cell set
- if removed, the zone walker is reset to the origin point only
- when added it is also added as a waypoint to the zone walker and the walker's cell path is added to the cell set. This started as an attempt to ensure that there are no isolated cells but it doesn't work. Basically I'm tinkering with ways to make sure we end up with a monolithic geometry with no holes.

bubblobill:
Surprised you didn't notice the abusable scale setting
image

bubblobill:
Since there were no earth-shattering changes required in the [Discord discussion](https://discord.com/channels/296230822262865920/1289853548028366869), and nobody saw fit to dispute my unilateral decision making, I think we should be pushing this forward.

___Copy of those unilateral decisions___

Resurrecting this because I really want the editable footprints.
## Firstly, disambiguation: ##
- Token *anchor, scaleSize, scaleX, scaleY*, and *imageRotation* relate purely to the token **image** being rendered. 
- They can be applied to anything that is not free or native sized.
- Their relationship to the footprint is:
    - The anchor is relative to the footprint centre.
    - With everything at scale 1.0, the image fills the footprint.
- Token facing rotates the footprint itself.

## Secondly, Properties of a Footprint ##
Origin Cell:
- Footprints have an origin cell from which all other cells radiate.
- The origin cell is defined as having its _centre_ at 0,0
- It follows that sub-cell-sized footprints shrink uniformly within the cell, equidistant from the edges.
- For purposes of snap-to-grid this is the snapping point and it snaps to the cell centre.
Inherent Points:
- Snap point, as above.
- Bounds centre, the centre of the rectangular bounds.
- Geometric centre, usually the mean of all vertices (centroid). The centre of the area defined by the footprint.
Optional Points: probably a Map of point Sets.
- Centre of rotation.
- Perception points, e.g. normal vision.
- Emanation points, e.g. normal light.

### Defaults ###
- Any unset point uses the bounds-centre. This mirrors the current way of things.

### Optional Points ###
- Can exist on a footprint and/or a token allowing a token to override the return value.

### Unilateral Decision ###
No more integers - I am tired of rounding errors screwing things up.

ColdAnkles:
@bubblobill - I agree with basically all you've put forth, but am unsure whether any are worth pursuing immediately under this PR, or whether they should be postponed as a later enhancement
Also unsure on my own availability and ability to implement

bubblobill:
More a roadmap for the future and design guide. 
When dealing with graphics objects that are subject to offsets and rotation it is far easier to manipulate them if their reference point is their central point. 
I keep wishing the functions were available in the work I'm doing, so that's an immediate benefit for me, and simply being able to define footprints is an immediate benefit to GMs everywhere.
What you have currently is perfectly fine. I'm mostly paving the way for what comes next.

PR #4890, opened 2024-09-02
rest - return result vs ParserException

### LINKED ISSUE #2489: REST functions that trigger exceptions should package them as JSON and return them
As discussed in the Discord server, it would be helpful if exceptions that occur during REST functions were packaged as JSON and returned.  As it stands, a macro using a REST function cannot determine that there was a timeout or other potentially transient error.

The simplest approach is likely to create a JSON object and set the `state` field to an invalid HTTP value (such as -1) as a flag that the rest of the JSON is exception data.  That allows the macro to test a single field in the return value to determine an error occurred.  The rest of the fields can be exception data.

**Additional context**
Perhaps this shouldn't apply to all exceptions, though.  All possible exceptions should be looked at to determine if any should be left uncaught.

Jmr3366:
Closes #2489 
`rest` commands throw parser exceptions and do not provide a `return` result that can be checked for failure.

### Description of the Change
`rest` commands that fail, will now provide a `return` of the same error previously provided through `ParserException`.
This can provide a graceful return for macro writers to check and respond accordingly.

### Possible Drawbacks
With this _feature_, macros that do not perform error checking on `rest` returns, where previously the macro would break due to the `ParserException`, will now continue execution.  This will most likely cause the macro to fail due to bad/wrong data types (eg. errors below vs expected data).

### Release Notes
Example of failures:
`[r: rest.get(url)]`
on a failure / timeout: `Unable to process function "{0}", HTTP Status Code: {1}`
on bad response: `Unable to process function "{0}", An Exception has occurred: {1} `

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4890)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/RESTfulFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/RESTfulFunctions.java
@@ -220,8 +220,7 @@ public class RESTfulFunctions extends AbstractFunction {
     // Execute the call and check the response...
     try (Response response = client.newCall(request).execute()) {
       if (!response.isSuccessful() && !fullResponse) {
-        throw new ParserException(
-            I18N.getText("macro.function.rest.error.response", functionName, response.code()));
+        return I18N.getText("macro.function.rest.error.response", functionName, response.code());
       }
 
       if (fullResponse) {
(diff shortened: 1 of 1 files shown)


cwisniew:
The feature request asks for the status to be returned in json format if it fails, this came will just return the translated string that would appear in the error message displayed when reporting the exception

Jmr3366:
It is not "What they'd like", you are correct.
However it is adequate by allowing _macro handling_ and does respond to the specific points mentioned; if it's a timeout or something else.

cwisniew:
> It is not "What they'd like", you are correct. However it is adequate by allowing _macro handling_ and does respond to the specific points mentioned; if it's a timeout or something else.

If it is a translated string then how will coders know unless the include checks for all the different translations?
I don't think I can approve this based on the feature request

PR #4900, opened 2024-09-08
Rework javascript getProperty API to handle missing properties better

### LINKED ISSUE #3028: [1.10 JavaScript] Token.getProperty does not return default value and does not process calculated properties.
**Describe the bug**
Token.getProperty does not return default value and does not process calculated properties.

**To Reproduce**
1. Create a set of default campaign properties where strength is 10 and strmod is `{Floor((Strength-10)/2)}`
2. Create the following javascript code in a macro:
```
        MapTool.chat.broadcast(MapTool.tokens.getSelected().getProperty("Strength"));
        MapTool.chat.broadcast(MapTool.tokens.getSelected().getProperty("StrMod"));
```
3. Create a token, select it, and run the above javascript with `js.evalNS`
4. `null` will appear in chat twice, because it doesn't use the default value.
5. Confirm this by changing strength to 12 and strmod to `{Strength+2}` on the token and re-running the javascript code on the token.
6. You will then get `12` and `{Strength+2}` in the chat window.
7. This is closer, but still not correct

**Expected behavior**
If you create a macro with
```
[r:getProperty("strength")]
[r:getProperty("strmod")]
```
and run it on a fresh token, you get `10` and `0` which would be the expected output of the same javascript code above.

**MapTool Info**
- Version: 1.10.0 javascript
- Install: Upgrade

 - OS: Windows
 - Version 10

fishface60:

closes https://github.com/RPTools/maptool/issues/3028

### Description of the Change

This changes the javascript `getProperty` function to look up the value in the campaign property type defaults if it's missing, and return `null` instead of `"null"` if you don't have permission to get that token's properties or the campaign doesn't have that property either, for improved ergonomics of null-handling in javascript.

This also adds `getRawProperty` which returns `null` if it's not set, and `getEvaluatedProperty` which may evaluate the value as a macro.

### Possible Drawbacks

If users of getProperty were checking for the value being missing with something like:

```javascript
let value = token.getProperty(name);
if (value == "null") {
    value = "10";
}
```

then this will now end up with value being `null`.

`getProperty` is the most desirable name for the most common operation. It's possible that most campaigns use properties that should be evaluated so should use the shorter name, and the form of `getProperty` that only falls back to the default value should be called something like `getUnevaluatedProperty` or `getDefaultedProperty`.

### Release Notes

* Made `token.getProperty` return `null` instead of the string `"null"` when properties are missing or not permitted to be read.
* Make `token.getProperty` return the default value for the token type if the value was not set on the token.
* Added `token.getRawProperty` for uses that shouldn't fall back to defaults and `token.getEvaluatedProperty` for uses that should use calculated defaults.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4900)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/script/javascript/api/JSAPIToken.java
+++ b/src/main/java/net/rptools/maptool/client/script/javascript/api/JSAPIToken.java
@@ -121,6 +121,21 @@ public class JSAPIToken implements MapToolJSAPIInterface {
   }
 
   @HostAccess.Export
+  public String getRawProperty(String name) {
+    boolean trusted = JSScriptEngine.inTrustedContext();
+    String playerId = MapTool.getPlayer().getName();
+    if (trusted || token.isOwner(playerId)) {
+      Object val = this.token.getProperty(name);
+      // Short-circuit to returning null so we don't return "null"
+      if (val == null) {
+        return null;
+      }
+      return "" + val;
+    }
+    return null;
+  }
+
+  @HostAccess.Export
   public String getProperty(String name) {
     boolean trusted = JSScriptEngine.inTrustedContext();
     String playerId = MapTool.getPlayer().getName();
@@ -130,6 +145,7 @@ public class JSAPIToken implements MapToolJSAPIInterface {
       // since it's not useful to return nulls and require
       // javascript to have to handle defaults when getInfo isn't even bound,
       // especially when the value gets unset if it matches the default.
+      // Evaluation is not performed automatically, use getEvaluatedProperty for that.
       if (val == null) {
         List<TokenProperty> propertyList =
             MapTool.getCampaign()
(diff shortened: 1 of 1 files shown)


fishface60:
I'll have a poke at CI failures this evening.

cwisniew:
> I'll have a poke at CI failures this evening.

They are formatting errors you can fix with 
./gradlew spotlessApply (or .\grqladlew.bat  spotlessApply of using windows)

fishface60:
> > I'll have a poke at CI failures this evening.
> 
> They are formatting errors you can fix with ./gradlew spotlessApply (or .\grqladlew.bat spotlessApply of using windows)

I even told myself "You should run that before submitting". I blame staying up until 11PM because the current token behaviour had been bothering me for long enough and doing bindings token status bars brought me close enough to the code to feel like I could do something about it.

cwisniew (approved):

PR #4903, opened 2024-09-12
Provide functionality for add-on development

### LINKED ISSUE #3460: [Feature]: One click/command addon loading
### Feature Request

I am writing an addon for maptool and i commonly do the steps, end editing -> call make to zip my addon -> switch to maptool -> go to file menu -> click on the addon libraries menu -> click on my addon -> click remove -> click on add addon -> navigate through 5 directories -> double click my addon -> finally i can test. 

### The Solution you'd like

I would like to have my addon immediately updated on a single command/interaction after proper setup. so that the flow is. end editing -> single short command/click/something similarly simple -> testing
there are some ways to accomplish this,
one could be to have maptool listen for commands so that i can just add a command to load my addon to my makefile. 
another would be to expand the macro api so that a macro that calls make and them loads the new version of my library is possible so using the new version of the addon is as simple as clicking a macro. 
Or add a option for maptool to watch a mtlib file so that whenever it changes it updates my addon, in this case i would only need to call make.
Any one of the three options presented would be enough.

Continue living with a fixable annoyance.

### Additional Context

i know this is minor but I am willing to fix this myself the only problem is that i am unfamiliar with the addon part of the code base. and i do not know which of the three options would be easiest. so any direction and tips would be appreciated.

cwisniew:
fixes #3460

### Description of the Change

This change adds the following features to the development tab of the dialog from File -> Add On Libraries.
* Directory to monitor for Add-on development
* Ability to import Add-ons within this directory
* Create the directory structure and example json files for new add on
* Create a .mtlib file based on the selected directory

### Possible Drawbacks
This still requires the user to import the add-on from the directory by clicking a button; add-on importing for large libraries is not something that is suitable to be done automatically on every file change. This is especially the case for windows where the file system often reports multiple change events for a single change. But it is one step closer to providing an API that an IDE could use to trigger importing of an add-on.

### Documentation Notes

This change adds the following features to the development tab of the dialog from File -> Add On Libraries.
* Directory to monitor for Add-on development
* Ability to import Add-ons within this directory
* Create the directory structure and example json files for new add on
* Create a .mtlib file based on the selected directory

CleanShot 2024-09-12 at 23 18 01@2x

CleanShot 2024-09-12 at 23 26 47@2x
### Release Notes
- Provide functionality for add-on development

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4903)


thelsing:
Would it also be possible to autoinstall an Add-on from a github-repo? i.e. download the repo as zip, rename to mtlib and import?

cwisniew:
> Would it also be possible to autoinstall an Add-on from a github-repo? i.e. download the repo as zip, rename to mtlib and import?

This is the eventual goal (its why there is a git URL in the add on) but not as part of this change

bubblobill (approved):
No obvious holes. Still has sufficient distancing to avoid security concerns.

src/main/java/net/rptools/maptool/model/library/addon/AddOnLibraryManager.java
```diff
@@ -193,4 +210,105 @@ public CompletableFuture<Set<Library>> getLegacyEventTargets(String eventName) {
...
                 .collect(Collectors.toSet()));
   }
+
+  /** Initializes the add-on library manager. */
+  public void init() {
+    externalAddOnLibraryManager = new ExternalAddOnLibraryManager(this);
+    String path = AppPreferences.externalAddOnLibrariesPath.get();
+
+    try {
+      externalAddOnLibraryManager.setExternalLibraryPath(Path.of(path));
```
kwvanderlinde, line 220: @cwisniew This is throwing an NPE when the preference isn't set since the default is `null`.

cwisniew: Hmm ok, I will fix it after work tonight (in a couple hours)

PR #4921, opened 2024-09-13
Add a Handlebars LibraryTemplateLoader and rework file loading to support partial templates

### LINKED ISSUE #4920: [Feature]: Handlebars partial template support would be helpful
### Describe the Problem

Stat sheets potentially share a lot of reusable components or it would merely be convenient to split the template into multiple files.

Handlebars supports this with the Partial Templates feature, which uses the `{{> path}}` syntax to load a template named `path`.

### The Solution you'd like

MapTool implements an appropriate template loader to support loading templates from addons.

* Make templates out of templates in my own build system and slowly drive myself insane.
* Accept the duplication between related templates that are very large.

### Additional Context

_No response_

fishface60:

closes https://github.com/RPTools/maptool/issues/4920

### Description of the Change

This has two major changes:

1. Moves the code for injecting the base tag to after templates have been applied instead of when they are read.

   This is necessary because the syntax for including partial templates is `{{> path}}`
   but the Jsoup code that adds the `<base>` tag normalises this strictly invalid HTML to `{{&gt; path}}`
   which makes it an invalid Handlebars template.
   
2. Adds a Handlebars [TemplateLoader](https://javadoc.io/static/com.github.jknack/handlebars/4.3.1/com/github/jknack/handlebars/io/TemplateLoader.html) that loads from lib:// URLs.

   Because the templates are already publicly exported from add-ons,
   and they can load css from other files with relative URLs
   it is natural to support loading partial templates the same way.
   
   It is implemented as a URL loader which is aware of the path of the template being loaded
   and interprets relative paths as relative to the directory.
   As-implemented, absolute paths can load content from other add-ons, but this could be prevented by adjusting the prefix and current paths.

See <https://gitlab.com/richardmaw/maptool-bar-test/-/tree/main/mtlib/library/public/sheets/basic> for an example stat sheet that uses partial templates.

### Possible Drawbacks

lib:// URIs were used to implement it as a URLTemplateLoader because handlebars.java already had a URL-based loader
and it was less intrusive to pass the URL around.

It would instead be possible to define a loader that uses a reference to the library which would have the benefit of allowing non-public templates, and bypasses the URL fetch machinery, but it may be useful to permit add-ons to depend on each other.

### Documentation Notes

In <https://wiki.rptools.info/index.php/Stat_Sheet_Tutorial> there is a comment:

> The `&` CSS selector didn't work when I created the examples, but may be available by the time you read this. It doesn't hurt to try it and see.

Which was probably due to the base tag being added before templates are parsed.

### Release Notes

- Fixed handlebars templates being mis-parsed if they were not strictly HTML.
- Added support for loading partial handlebars templates with the `{{> path}}` syntax, which includes the contents of the named partial template file in the current template. If the path is an absolute path the first component is the library namespace. If the path is a relative path it is in the same directory as the template.
- Adds the `include` helper, e.g. `{{include "foo" bar=baz}}` will include the partial template `"foo"` with the context altered to set bar to the value of baz in the current context.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4921)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/MacroDialogFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/MacroDialogFunctions.java
@@ -35,13 +35,11 @@ import net.rptools.maptool.language.I18N;
 import net.rptools.maptool.model.library.Library;
 import net.rptools.maptool.model.library.LibraryManager;
 import net.rptools.maptool.util.FunctionUtil;
+import net.rptools.maptool.util.HTMLUtil;
 import net.rptools.parser.Parser;
 import net.rptools.parser.ParserException;
 import net.rptools.parser.VariableResolver;
 import net.rptools.parser.function.AbstractFunction;
-import org.jsoup.Jsoup;
-import org.jsoup.nodes.Element;
-import org.jsoup.parser.Tag;
 
 public class MacroDialogFunctions extends AbstractFunction {
   private static final MacroDialogFunctions instance = new MacroDialogFunctions();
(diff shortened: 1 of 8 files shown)


cwisniew (approved):

cwisniew:
@fishface60 I know this is an old but I noticed a problem with one of the changes while looking at something for @bubblobill  (so you can blame him :) ).

I noticed in 
https://github.com/RPTools/maptool/blob/bd70c32dd9ddf403247bad2a08b627756bad2227/src/main/java/net/rptools/maptool/util/HandlebarsUtil.java#L94-L100

There is a hardcoded `lib://`, but for it to work in the future all of the URL stuff needs to be relative URL's only and not rely on lib:// prefix. Sorry I should have caught this during the review

fishface60:
> There is a hardcoded `lib://`, but for it to work in the future all of the URL stuff needs to be relative URL's only and not rely on lib:// prefix. Sorry I should have caught this during the review

Ok, thanks for letting me know.

I've been a bit out if the loop in following MapTool development for a while due to my laptop dying, is the URL stuff you speak of merged already and I need to fix this, or is there a development branch/MR I need to develop a fix against to not block merging it?

fishface60:
Ok, found time to take a look at it again.

https://github.com/RPTools/maptool/pull/5461 broke partial template loading since it inserted the glue at file load again after I had changed things so that the glue was added by the caller.

I'll take a look at what those changes do first in case it's broken anything else, then think about what changes need to happen for partial templates.

fishface60:
Ok, that looks to be because plumbing in content-type detection to prevent mangling in https://github.com/RPTools/maptool/pull/5489 identifies the handlebars as HTML, so there probably aren't issues with other file types loaded via URL.

PR #4968, opened 2024-10-02
Fixed State/Bar images resizing on rotated tokens

### LINKED ISSUE #4944: [Bug]: States being resized based on rotation
### Describe the Bug

When Tokens rotate, the state images on them get resized, which can make the Token look a completely different size than it is.  I have attached an example.

In this example, all of the Tokens are the same size, but they look like they're different sizes because their states get bigger when they're facing vertices.  Expected behavior would be that the states remain the same size regardless of the facing of the Token.

### To Reproduce

Add a state with an image to a token.  Rotate the token so it is facing a square vertex.

### Expected Behaviour

For states to remain the same size regardless of the facing of the Token, so it doesn't look like Tokens are all different sizes when they're rotated.

### Screenshots

image

### MapTool Info

1.14.3

### Desktop

Windows

### Additional Context

_No response_

bubblobill:
fixes #4944

### Description of the Change
Changed render area for states and bars.
Area size is now based on non-rotated token bounds.
Now renders to a square area based on token's max dimension for non isometric tokens.

### Possible Drawbacks
none

### Documentation Notes
n/a

### Release Notes
Fixed State/Bar images resizing on rotated tokens

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4968)


cwisniew:
@bubblobill ruh roh it doesn't compile

maptool\src\main\java\net\rptools\maptool\client\ui\zone\renderer\ZoneRenderer.java:2604: error: incompatible types: <null> cannot be converted to int
      tmpToken.setFacing(null);

bubblobill:
Ja, stoopid computer. 
Fixed now.

cwisniew (approved):

PR #4981, opened 2024-10-05
Decompose `Grid.getShapedArea()` so we can reuse logic between grids

### LINKED ISSUE #4980: [Bug]: Incorrect shapes for beam and hex lights on isometric grids
### Describe the Bug

On isometric grids, hex lights are rendered as an ellipse and beams do not get visually shorted when lying vertically on the screen.

### To Reproduce

1. Create a map with isometric grid
2. Define these lights in the campaign properties:
3. Add a couple tokens to the map and equip the above lights.
    - The hex light will show up as a circle.
4. Rotate a token with a beam light equiped.
    - When horizontal, the length of the beam is correct.
    - When vertical, the beam length doesn't change, i.e., it is quite a bit longer than it should be.

### Expected Behaviour

Hex lights should be hexagonal. Beam lights should be foreshortened according to the grid.

### Screenshots

Here two beams offset by 90° to show the lack of perspective:
image

And here is a "hex" light:
image

### MapTool Info

1.15.2

### Desktop

Linux Mint 22

### Additional Context

_No response_

kwvanderlinde:

Fixes #4980

### Description of the Change

This refactors `Grid.getShapedArea()` for more consistency between grid types. The main goal was to allow `IsometricGrid` to reuse the main logic and transform the results rather than having to duplicate and specialize most of the logic for its unique circumstance.

With this change, isometric grid lighting behaves the same as for other grid types, save for the footprint area and 45° difference in the meaning of token facing. This means hex lights render as hexes, and beams are properly foreshortened. It also means no special logic for the other shapes, so we can be more confident that those will remain consistent into the future.

### Possible Drawbacks

None. Unless folks want to complain about hexes being hexagonal.

### Documentation Notes

N/A

### Release Notes

- Fixed a bug where hex lights were rendered as circles on isometric grids
- Fixed a bug where beam lights were not foreshortened on isometric grids

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4981)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/Grid.java
+++ b/src/main/java/net/rptools/maptool/model/Grid.java
@@ -387,59 +387,63 @@ public abstract class Grid implements Cloneable {
     fireGridChanged();
   }
 
+  // region Light shapes
+
   /**
-   * Called by SightType and Light class to return a vision area based upon a specified distance
+   * Get the grid-relative angle of the token based on its facing.
    *
-   * @param shape CIRCLE, GRID, SQUARE, CONE or LINE
-   * @param token Used to position the shape and to provide footprint
-   * @param range As specified in the vision or light definition
-   * @param arcAngle Only used by cone
-   * @param offsetAngle Arc distance from facing, only used by cone
-   * @param scaleWithToken used to increase the area based on token footprint
-   * @return Area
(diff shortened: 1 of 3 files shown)


bubblobill (approved):
An improvement. One day this will all be pared down to something lean and sensible.

PR #4982, opened 2024-10-06
Do not disclose GM maps to players in status UI

### LINKED ISSUE #4595: [Bug]: Connections window disclosing GM maps to the players
### Describe the Bug

When the GM is hosting a server and on a map that the players can't see, the players' _Connection_ window will show the name of the GM's current map. This discloses information that players shouldn't know about.

### To Reproduce

1. Open an instance of MapTool, start a server as GM
2. Open a second instance, connect to the server as a player and open the Connections window.

### Expected Behaviour

Whether or not the map is hidden, players should not see which map the GM is on.

### Screenshots

This is the player's view:
image

In this example, the map "Grasslands2" is not visible to players.

### MapTool Info

1.14.3, 1.15.2

### Desktop

Linux Mint 21.2

### Additional Context

_No response_

kwvanderlinde:

Fixes #4595

### Description of the Change

For player clients, this hides the name of the GM's zone in the Connections window and status bar. Players can still see each other's zones, and GMs can still see everyone's zone.

### Possible Drawbacks

None

### Documentation Notes

In the Connections window, players cannot see the map any GM is on. Same for the status bar indicator.

### Release Notes

- Fixed a bug where names of hidden maps are shown to players in the Connections window when a GM is on the map.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/4982)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/swing/PlayersLoadingStatusBar.java
+++ b/src/main/java/net/rptools/maptool/client/swing/PlayersLoadingStatusBar.java
@@ -56,11 +56,11 @@ public class PlayersLoadingStatusBar extends JLabel {
         new StringBuilder(I18N.getText("ConnectionStatusPanel.playersLoadedZone", loaded, total));
 
     for (Player player : players) {
+      // GMs can see everyone's zone, players can only see each other's.
+      var showZone = MapTool.getPlayer().isGM() || !player.isGM();
+
       String text;
-      if (player.isGM()) {
-        // Don't display zone information.
-        text = player.toString();
-      } else {
+      if (showZone) {
         var zone =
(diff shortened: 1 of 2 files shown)


cwisniew:
Although having more than one GM connected is rare, we should still probably show the information to GMs.

kwvanderlinde:
Fair enough. I didn't think it was too important for GMs to keep track of each other, but I will add it in.

cwisniew:
> Fair enough. I didn't think it was too important for GMs to keep track of each other, but I will add it in.

Its not so much keeping track of which map they were on, but the original request that generated this functionality was about knowing if maps had loaded for other users. So I can see people wanting that feature back

kwvanderlinde:
Updated so GMs can once again see each other's zone.

cwisniew (approved):

Issue #4961, opened 2024-09-30
[Bug]: NPE when clients mouse-off an NPC token they don't own


fishface60:
### Describe the Bug

```
Exception in thread "JavaFX Application Thread" java.lang.NullPointerException: Cannot invoke "net.rptools.maptool.client.ui.htmlframe.HTMLOverlayManager.updateContents(String, boolean)" because "overlay" is null
	at net.rptools.maptool.client.ui.sheet.stats.StatSheet.lambda$clearContent$1([StatSheet.java:77](http://statsheet.java:77/))
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$10([PlatformImpl.java:456](http://platformimpl.java:456/))
	at java.base/[java.security](http://java.security/).AccessController.doPrivileged(Unknown Source)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$11([PlatformImpl.java:455](http://platformimpl.java:455/))
	at [com.sun.glass](http://com.sun.glass/).ui.InvokeLaterDispatcher$[Future.run](http://future.run/)([InvokeLaterDispatcher.java:95](http://invokelaterdispatcher.java:95/))
	at [com.sun.glass](http://com.sun.glass/).ui.gtk.GtkApplication._runLoop(Native Method)
	at [com.sun.glass](http://com.sun.glass/).ui.gtk.GtkApplication.lambda$runLoop$11([GtkApplication.java:316](http://gtkapplication.java:316/))
	at java.base/[java.lang.Thread.run](http://java.lang.thread.run/)(Unknown Source)
```

### To Reproduce

1. Create a new campaign.
2. Add an add-on with stat sheets.
3. Place an NPC token
4. Start a server
5. Connect a player client to the server
6. Have the player repeatedly mouse-over the NPC token and mouse-off again.

### Expected Behaviour

No errors spammed to the client's logs

### MapTool Info

First observed in v1.14.3 but present in develop

### Desktop

Ubuntu 24.04.1 LTS

### Additional Context

This is probably sufficient to fix it.

```diff
diff --git a/src/main/java/net/rptools/maptool/client/ui/sheet/stats/StatSheet.java b/src/main/java/net/rptools/maptool/client/ui/sheet/stats/StatSheet.java
index 331e168ac..ed0a337ca 100644
--- a/src/main/java/net/rptools/maptool/client/ui/sheet/stats/StatSheet.java
[...]

Issue #4954, opened 2024-09-27
[Bug]: Fancy/new Stat Sheet option freezes Overlays


Akaku-Keninari:
### Describe the Bug

Now that my old issue with overlays just not calling forms is fixed, here is a new one: when I change a statsheet from legacy to any of the new built in options, my overlays flat stop responding to clicks.

### To Reproduce

1: Open test campaign
2: Impersonate Perra Token
3: Validate the buttons on the bottom of the screen interact
4: Go to campaign properties and change the veg (pc) statsheet to any option, like smoke
5: See buttons do not press, nor events run on them (mouseover power detail not firing either)

https://github.com/user-attachments/files/16872086/submitbugg.zip

### Expected Behaviour

Using a custom statsheet shouldn't block the whole screen's overlays transparently

### MapTool Info

1.15.2 installed with msi installer, old version removed

### Desktop

Windows 11

Jmr3366:
@Akaku-Keninari  can you confirm if you're still experiencing this?
I believe you made changes to your code to compensate, it would be great to know if you'd be able to revert it back, and test.
At least we'd know this update _also_ resolved your initial bug report and it can be closed.

Akaku-Keninari:
This seems to work in 1.16.1 now at least.

Jmr3366:
@Phergus would you be able to close this one?

PR #5007, opened 2024-10-18
Stat Sheet Bug Fixes

Jmr3366:
Fixes:
#4961 NPE on mouse off (needs testing, I can't validate)
#5008 Stat sheet not in front
#4954 User Overlays break with new statsheets

### Description of the Change
#4961 adds an empty var `if` check.  Needs testing as I don't have the OS to validate this change on.  
#5008 changed variable used from min to max
#4954 forces sorting for all overlays (removed `needsSorting` bool)

### Possible Drawbacks
None

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5007)


cwisniew:
This overlay may be used in other functionality in the future so the name can't be hard coded like this

Jmr3366:
~It needs more work, I'm still missing something~

edit
~Created a new _init_ in HTMLOverlayPanel using _invokeLater_ being executed in _initialize_~

kwvanderlinde:
> It needs more work, I'm still missing something

@Jmr3366 At a glance this look good to me. Is this comment still true?

Jmr3366:
I think it is all good but wanted you or Craig to QC it.

src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLOverlayPanel.java
```diff
@@ -117,6 +119,20 @@ public ConcurrentSkipListSet<HTMLOverlayManager> getOverlays() {
     return overlays.clone();
   }
 
+  public HTMLOverlayManager init() {
```
kwvanderlinde: I think this approach to fixing #4954 actually masks the issue rather than fixing it. Did some testing, I believe the only thing that needs to change is in `showOverlay()` - we need `needsSorting` to be true even for internal overlays. Otherwise the new overlay ends up in front of the `front` region.

Also I have a personal motivation to not have overlays automatically open :slightly_smiling_face: - see #5044. Overlays on my system cause a lot of input lag, and this would force me to endure it despite not using the stat sheet feature yet.

Jmr3366: You're correct on both points (masking,sorting) and I looked right past that, thank you sir!

kwvanderlinde (approved):

PR #5019, opened 2024-10-24
Fix various issues with transfering map topology to and from token topology

### LINKED ISSUE #5018: [Bug]: transferVBL() et al. with delete flag does not update tokens on other clients
### Describe the Bug

When calling `transferVBL()` with the delete flag set to `true`, the current token's Wall VBL will be moved to the map on the local client. But on remote clients, the Wall VBL will be added to the map but not cleared on the token. Reconnecting does not solve the issue, suggesting the server does not receive that update.

This also applies to the other `transfer*()` functions.

### To Reproduce

1. Add some Wall VBL to a token.
2. Start a server.
3. Connect a remote GM client. On this client select the topology tools so it's apparent what is happening.
4. On the host client, select the token and run this command: `[h: transferVBL(1, true, getSelected())]`
5. On the host client, see the topology has been transfered to the map and is no longer present on the token.
6. On the remote client, see the topology on both the map and the token. Edit the token to convince yourself if needed.

### Expected Behaviour

The token should not have any of the removed topology remaining on any client.

### Screenshots

_No response_

### MapTool Info

1.15.2

### Desktop

Linux Mint 22

### Additional Context

_No response_

kwvanderlinde:

Fixes #5018

### Description of the Change

There are a few issues fixed in this PR.

First is the fix for #5018 itself, which is done by avoiding any direct modification of a token's topology. Instead, the server command is used to update both local and remote clients in one place. The same pattern is done for map topology, though in all cases it was at least done correctly. This used to be the role of `TokenVBL.renderTopology()`, but the server commands is a much more natural place for it. The add/remove operations and commands have also been merged into a single update operation that can draw and erase.

Next is the fix for the NPE mentioned in the comments. Our protobuf messages do not support sending null token topology, so instead we send empty areas along. On the other end these empty areas are turned back into nulls since there are a few places that rely on null and I wasn't confident about nailing all those down.

Finally is the fix for the misaligned token topology also mentioned in the comments. This was simple a matter of incorporating the token anchor in the rotation used by `TokenVBL.getTopology_underToken()` and `TokenVBL.getMapTopology_transformed()`. These methods were also refactored to be remove unused calculations, to not duplicate each other, and to avoid unnecessary intermediate `Area` creation.

### Possible Drawbacks

All sunshine.

### Documentation Notes

N/A

### Release Notes

- Fixed a bug where token topology transfered to the map would not be removed on connected clients.
- Fixed a bug where topology transfered to a token would not align with where it was on the map.
- Fixed a bug where using `transferVbl()` on a token with no VBL would print a NullPointerException in chat.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5019)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ClientMessageHandler.java
+++ b/src/main/java/net/rptools/maptool/client/ClientMessageHandler.java
@@ -101,7 +101,7 @@ public class ClientMessageHandler implements MessageHandler {
       log.debug("{} got: {}", id, msgType);
 
       switch (msgType) {
-        case ADD_TOPOLOGY_MSG -> handle(msg.getAddTopologyMsg());
+        case UPDATE_TOPOLOGY_MSG -> handle(msg.getUpdateTopologyMsg());
         case BOOT_PLAYER_MSG -> handle(msg.getBootPlayerMsg());
         case CHANGE_ZONE_DISPLAY_NAME_MSG -> handle(msg.getChangeZoneDisplayNameMsg());
         case CLEAR_ALL_DRAWINGS_MSG -> handle(msg.getClearAllDrawingsMsg());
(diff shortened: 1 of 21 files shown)


kwvanderlinde:
This will conflict with #5011 because of the changes to tools. I suggest for any reviewer's sake that this PR be done first.

cwisniew (approved):

PR #5050, opened 2024-11-17
Add wall-based topology

### LINKED ISSUE #3485: [Feature]: Overlappable chunks of topology
### Feature Request

At the moment, map topology is "flat", i.e., any point on a map either has a given type of topology or it doesn't. This is fine for Wall VBL and MBL which strictly block vision and pathfinding respectively, but it limits the effects that can be achieved for Hill VBL and Pit VBL.

### The Solution you'd like

Make topology more like drawings in that there can be independent pieces of topology on the map. These would be able to be layered on top of one another without being merged into a single area.

Token topology could be modified to behave in this way. Any given token's topology is already separate from map topology and any other token topologies in the sense that it can be moved and manipulated and technically layered on top of each other. At the moment token topology is always merged with map topology on-the-fly when running the vision and pathfinding algorithms, but it seems natural to consider making them truly layered.

Another alternative is to Introduce a different/new system for line-of-sight and pathfinding. This system could focuses on higher-level concepts of like walls, doors, windows, etc. If such a system were based on drawing line segments rather than covering areas, it wouldn't have to worry about explicitly allowing layering as that could happen naturally.

### Additional Context

A use case for layered topology is a multi-level tower, as seen from the exterior. This could be represented by a few circles of Hill VBL layered on top of one another.

Similarly, a layered pit could be achieved by layering some nested circles of Pit VBL on top of one another.

kwvanderlinde:

Resolves #3485

### Description of the Change

This adds a new type of topology in the form of thin, editable walls. These walls always block vision and movement through them. All walls in a zone form a graph, represented by `WallTopology`. Existing topology continues to be modeled as `Area`, but are generally referred to as "mask topology" and used through the new `MaskTopology`

#### Modelling

The new `WallTopology` type represents all wall topology in a zone as a single graph. The edges are the walls and the vertices are the endpoints of the walls. Operations are provided for adding new walls, extending existing walls, splitting walls, and merging vertices.

#### Serialization

There is an XStream converter for `WallTopology` so that it's internal structure doesn't become a sticking point for future enchancements. Because we only have basic walls, the XML will look like this:
```
<...>
  <vertices>
    <vertex id="8A647D10EE1B4AFA8BEF4B93AEFB624D" x="0.0" y="0.0" />
    <vertex id="E3A964F13221428582E491AB3D3D3764" x="100.0" y="50.0" />
    <vertex id="C510571E42C74080AF34EC00D47FCF97" x="200.0" y="300.0" />
  </vertices>
  <walls>
    <wall from="8A647D10EE1B4AFA8BEF4B93AEFB624D" to="E3A964F13221428582E491AB3D3D3764"/>
    <wall from="E3A964F13221428582E491AB3D3D3764" to="C510571E42C74080AF34EC00D47FCF97"/>
  </walls>
</...>
```
In the future, expect to see additional fields added to walls. I expect vertices to remain simple, i.e., only have position.

#### Tooling

A new tool is dedicated to editing wall topology. It allows dragging existing walls around in their entirety, or by a single vertex. New walls can be drawn freestanding or connected to existing walls. Walls can be split in the middle if new connections are needed. Two vertices can be merged together with the result having all the walls from the originals; if there was an wall between the two vertices, it will be deleted. If a wall is deleted, any "dangling" vertices will be automatically removed.

I tried to make this tool at least slightly general through the `Rig` interface, which exposes access to handle (vertices) and other movable elements (walls). Right now it is only applied to the wall topology tool, but my hope is that drawings and mask topology can adapt the same concept to make those editable in the future.

#### Vision updates

Wall topology is fully supported in visibility operations, but required some reworking to account for new edge cases. A critical aspect for the vision sweep is that all intersection points be represented as endpoints, otherwise the results will not reflect the inputs. Previously we could ignore this issue by running the sweep on each mask type separately, since a mask cannot intersect itself. But wall topology _can_ intersect itself, so we can't ignore this any longer. And since we need to solve it anyways, we now handle all topologies in a single sweep.

For masks, the hierarchical `AreaTree` and `AreaMeta` are gone in favour of a plain list of `Polygon`. This list, together with the line segments from the `WallTopology`, are combined into a single `NodedTopology` that copies the input geometry and adds any intersection points as needed for the vision sweep. This `NodedTopology` is cached in the `Zone` (not `ZoneView`) and is invalidated any time either the wall topology or mask topology changes.

Now that masks have their own custom type in `MaskTopology`, and are no longer arranged in a tree structure, the logic in `VisionBlockingAccumulator` has been moved to `MaskTopology` and considers only a single polygon at a time. This new implementation is functionally identical to the original, though it may produce more segments than previously since it doesn't have the tree structure to help out.

#### Pathfinding updates

There is a mild change in behaviour with pathfinding. Previously, we would union all zone topology of a given type, including all token topologies, then subtract back out the key token's topology. Now we instead leave out the key token during the union. In most cases this won't make a difference, but it avoids an edge case where a token with MBL would ignore any map MBL that overlapped it.

Some extra work is needed to keep the pathfinding performance working well. It used to suffice to gather all topology into a single `PreparedGeometry` since it provides fast intersection checks. But this only worked because all masks would be represented as `Polygonal` geometries. With the additional fo walls, there are now `Lineal` geometries as well, and `PreparedGeoemtry` performance suffers greatly with the mixed types. The new `MovementBlockingTopology` is a drop-in replacement for `PreparedGeoemtry`, and internally wraps separate `PreparedGeometry` for the `Polygonal` and `Lineal` components.

#### Future work

In order to keep this change somewhat focused, as large as it is, it only adds the most basic walls. There is no directionality, no equivalent to terrain walls, no ability to configure vision and movement blocking, etc. I will be making new FRs to add such things until wall topology can be considered a fully functional alternative to mask topology.

### Possible Drawbacks

- More code to maintain.
- The UX for the tool is probably not the greatest.
- Topology has some different performance characteristics now. Seems to be similar or slightly better in the cases I can test, but it's possible there are existing campaigns that will perform worse in vision or pathfinding.

### Documentation Notes

Here's is an example of drawing some simple wall topology using the snap-to-grid feature:
simple-demo

Any of the handles can be grabbed and dragged elsewhere. 

Here's a portion of a demo UVTT import where the LOS info is used to make wall topology (used to be Wall VBL + MBL):
dd-example

### Release Notes

- Added a new type of topology in the form of editable walls.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5050)


FullBleed:
Looking forward to this.  I use the hollow shapes and polyline tool for the vast majority of my vbl... and solid shapes often to just close gaps.  This new approach should really streamline the process and make refinement much simpler.

Looks to be a sneaky BIG change for (I hope) the better!

cwisniew:
> Looking forward to this. I use the hollow shapes and polyline tool for the vast majority of my vbl... and solid shapes often to just close gaps. This new approach should really streamline the process and make refinement much simpler.
> 
> Looks to be a sneaky BIG change for (I hope) the better!

It will be for the better, well apart from the bit where it doubles down on the word Topology, but I have given up on trying to convince anyone that's the wrong word any more :)

@kwvanderline if I can make one request (before everyone else does) can we get the ability to only block movement (e.g. window), or I guess only block vision (curtain). Because it wont be long before that feature request is raised :)

kwvanderlinde:
> It will be for the better, well apart from the bit where it doubles down on the word Topology, but I have given up on trying to convince anyone that's the wrong word any more :)

I'm open to suggestions on names. This may be our best chance to eventually break free of it :slightly_smiling_face: Actually one perk of this implementation is that we at least don't write out the term "topology" anywhere when it comes to wall topology. So we could change it at any time as a refactoring + small UI change.

> @kwvanderline if I can make one request (before everyone else does) can we get the ability to only block movement (e.g. window), or I guess only block vision (curtain). Because it wont be long before that feature request is raised :)

I'd rather not actually... Certainly some FR to that effect is going to get raised quickly (I intend to do so myself), but it's not obvious that we should simply allow toggling movement or vision blocking for segments. Maybe that will be the right thing to do, but maybe it would be better to explicitly mark some segments as windows or doors or curtains or whatever else, and derive blocking and other behaviours from that. I'd also have to make various internal and UI additions which would add a fair bit to the effort (especially the UI part for me, even though it sounds so simple).

There are also some other basic features that I believe should be added to wall topology early on, but which would interact somehow with decisions around how movement and vision configuration are handled. Directionality of walls is one in the front of my mind. I'd rather be able to discuss some of these additions in dedicated FRs to better understand the pros and cons beforehand.

FullBleed:
wallScaping or mapScaping?

cwisniew (approved):

cwisniew (approved):

cwisniew (approved):

kwvanderlinde:
Fixed the conflicts.

PR #5051, opened 2024-11-18
Feature to lock an overlay

### LINKED ISSUE #3414: [Feature]: Allow GM to Lock Overlays.
### Feature Request

A GM might want to use the overlay feature to display somethings on a client that they don't want the client to be able to toggle off.

### The Solution you'd like

For overlay function consistency (i.e. setOverlayVisible) a setOverlyLock() function would probably do the trick:

Usage

setOverlayLock(name, visible)

Parameters

    name - Overlay name.
    lock - A 1 or 0 to lock the overlay visibility toggle in the Windows>Overlays navigation menu.

_No response_

### Additional Context

My use-case for this was to be able to slowly darken a player's screen due to certain events.  It's a great effect, but if players can toggle the overlay on and off at will they can subvert it (like giving them a toggle switch for FoW.)

Jmr3366:
Closes #3414 by adding a `locked` option in `overlay`

### Description of the Change
Extended the 2nd parameter offered in `overlay`.
Like `zorder=<num>` you can now add `locked=<0/1>`. 
`locked` defaults to `0` (not locked)

Locked overlays appear in the **Window** -> **Overlays** menu but are _greyed out_ and can't be _unchecked_
### Possible Drawbacks
None
### Release Notes
Examples:
`[overlay("Test","zOrder=10;locked=1;"):[ { content to display }]`

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5051)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/MacroDialogFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/MacroDialogFunctions.java
@@ -26,11 +26,8 @@ import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.ExecutionException;
 import java.util.regex.Pattern;
 import net.rptools.maptool.client.MapTool;
-import net.rptools.maptool.client.ui.htmlframe.HTMLDialog;
-import net.rptools.maptool.client.ui.htmlframe.HTMLFrame;
-import net.rptools.maptool.client.ui.htmlframe.HTMLFrameFactory;
+import net.rptools.maptool.client.ui.htmlframe.*;
 import net.rptools.maptool.client.ui.htmlframe.HTMLFrameFactory.FrameType;
-import net.rptools.maptool.client.ui.htmlframe.HTMLOverlayManager;
 import net.rptools.maptool.language.I18N;
 import net.rptools.maptool.model.library.Library;
 import net.rptools.maptool.model.library.LibraryManager;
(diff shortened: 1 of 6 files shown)


Jmr3366:
Is there a known process/procedure to mitigate this (sometimes it happens, other times not)?
image

kwvanderlinde:
> Is there a known process/procedure to mitigate this (sometimes it happens, other times not)? image

You can avoid it by not using closing keywords in the PR (e.g., just write "#3414" instead of "closes #3414"). After the jobs run the description can be updated.

But I think we ought to remove the job until we know what's wrong with it.

cwisniew:
> > Is there a known process/procedure to mitigate this (sometimes it happens, other times not)? image
> 
> You can avoid it by not using closing keywords in the PR (e.g., just write "#3414" instead of "closes #3414"). After the jobs run the description can be updated.
> 
> But I think we ought to remove the job until we know what's wrong with it.


I am in two minds as when it works it means I don't have to do anything before merging, when it fails it at least reminds me to assign the label to the PR

Otherwise I end up spending too much time on release chasing down all the PRs and assigning them labels so the release notes get created correctly. Is it perfect no but it's better than when it wasn't there. Also please don't leave off the keywords that close PRs as that creates more work and I will soon get annoyed enough I will start rejecting them :)

cwisniew (approved):

PR #5055, opened 2024-11-19
Add landing maps to campaigns

### LINKED ISSUE #75: Set Landing Map for Client Connect
User issue from Dorpond working with a new MapTool GM
"When he loaded up his campaign, we connected as players and arrived in a map he wasn't expecting us to be on. My thoughts went back to our past discussions of having a landing page, where all players connected arrive at. This could be as simple as a set (default) map, or be as elaborate as having a non-map, full text/writing capable (recapping last sessions, listing current quests, etc) landing page."

After discussion, the thoughts are to allow the GM to set a landing page for the campaign or default the players to the GMs current map when they connect to a server. 

Possible solutions include
1. Flagging a map as a landing page on the MapTool Map menu. Rules around this include only having one landing page per campaign and the map must be player visible. If the map is set to non-visible, the flag is removed.
2. Force the players to the current GM map (by default). Again, the map would need to be player visible.
3. Consistently force the players to the first visible map in the Map List.

These are likely in precedent-order. If a Landing Page is set then go there. If there is no Landing Page then go to the GM's map. If the GM map isn't player-visible, go to the first visible map in the list.

kwvanderlinde:

Resolves #75

### Description of the Change

Adds a zone ID to the `Campaign` to mark it as the landing map for players. When a remote player client connects, if the campaign has a landing map set, it will use that. Otherwise it will fall back to the current behaviour of finding the first player-visible map.

This does not affect GMs - they still open the first map in the list when they connect. And local clients will still try to stay on the current zone if possible.

### Possible Drawbacks

None

### Documentation Notes

Use the _Map > Landing map_ toggle to set the current map as the landing map:
landing-map-menu-item

### Release Notes

- Added a Map menu option to set the current map as a landing map for when players connect to the server.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5055)


cwisniew (approved):

FullBleed:
This is great!  I have a onCampaignLoad that handles this sort of thing but this is a long overdue feature to gives users control over what map users are sent to.  

Would like to see this supported with set/getLandingMap() functions and a checkbox in the New Map and Edit Dialogs.  Should I put in a freq for those?

cwisniew:
> This is great! I have a onCampaignLoad that handles this sort of thing but this is a long overdue feature to gives users control over what map users are sent to.
> 
> Would like to see this supported with set/getLandingMap() functions and a checkbox in the New Map and Edit Dialogs. Should I put in a freq for those?

Hold off maybe on the macro one as it would be good to get feedback on the related functionality for setting the starting viewable area first before creating the macros. UI changes are easy to make based on feedback, once a macro function is created its kinda stuck that way forever.

kwvanderlinde:
@FullBleed I'll tack on another PR for the edit map dialog. Will be pretty simple and I think natural to keep part of #75.

PR #5067, opened 2024-11-22
Bring back unique light sources (no new macro functions)

### LINKED ISSUE #3087: Add 'Implicit Lights'

kwvanderlinde:

Resolves #3087

### Description of the Change

This adds unique light sources, originally from #4356. Unlike before, this change focuses primarily on UI changes and does not include any macro functions for manipulating unique light sources.

#### UI changes

The only way to define unique light sources for now is through the **Edit Token** dialog's _Config_ tab. It has a textbox for entering light source definitions, following the same syntax as lights in the **Campaign Properties** dialog, minus the support for groups.

The token context menu's "Light Sources" sub-menu now shows a "Unique" group containing the token's unique light sources. If the token has none, this group will not be shown. Attaching and detaching unique lights works exactly the same as for campaign lights.

#### UVTT importer

The UVTT importer now defines and attaches a unique light source on the light tokens it produces. The range of the light source is taken from the light's `"range"` field. The color is taken from the `"color"` field, but only if the imported map does not have baked lighting. Clear lights are produced if lighting is baked since the light color is already present on the map. 

The light's properties are still written to the token's GM notes as JSON, but now all properties are written rather than a hand-picked subset.

There are two limitations on this part of the importer:
- The `"intensity"` field goes unused as we don't have a equivalent concept.
- The UVTT format includes no shape information for lights, so the shape will always be set to "circle".

#### Macro function changes

The only change to macro functions is to add the `"$token"` type to [`getLights()`](https://wiki.rptools.info/index.php/getLights), [`setLight()`](https://wiki.rptools.info/index.php/setLight), and [`hasLightSource`](https://wiki.rptools.info/index.php/hasLightSource). The `"*"` wildcard type will also include any unique light sources on the token in addition to those in the campaign.

### Possible Drawbacks

Existing macros may fail if they try to match up the results of `getLights("*")` with the lighting info in `getInfo("campaign")`, without validating that the keys exist. This would not immediately impact existing campaigns since it is only an issue if run against tokens with unique light sources added to them.

### Documentation Notes

How the Edit Token dialog looks for defining unique light sources:
defining via Edit Token dialog

How the token context menu looks for turning unique light sources on and off:
attaching via token context menu

For `getLights()` **type** parameter:
- If set to `"$token"`, unique light sources on the token are returned.
- If set to `"*"`, all light sources - including unique light sources on the token - are returned.

For `setLight()` **type** parameter:
- If set to `"$token"`, `name` refers to a unique light source on the token.

For `hasLightSource()` **type** parameter:
- If set to `"$token"`, only unique light sources on the token are checked.
- If set to `"*"`, all light sources - including unique light source on the token - are returned

### Release Notes

- Added unique lights to tokens, with support in UVTT importer for creating them on imported lights.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5067)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ServerCommandClientImpl.java
+++ b/src/main/java/net/rptools/maptool/client/ServerCommandClientImpl.java
@@ -641,6 +641,18 @@ public class ServerCommandClientImpl implements ServerCommand {
   }
 
   @Override
+  public void toggleLightSourceOnToken(Token token, boolean toggleOn, LightSource lightSource) {
+    var update = toggleOn ? Token.Update.addLightSource : Token.Update.removeLightSource;
+    // We only need to send the ID of the light source.
+    updateTokenProperty(
+        token,
+        update,
+        TokenPropertyValueDto.newBuilder()
+            .setLightSourceId(lightSource.getId().toString())
+            .build());
+  }
+
+  @Override
   public void setTokenTopology(Token token, @Nullable Area area, Zone.TopologyType topologyType) {
     if (area == null) {
       // Will be converted back to null on the other end.
@@ -706,14 +718,6 @@ public class ServerCommandClientImpl implements ServerCommand {
   public void updateTokenProperty(Token token, Token.Update update, String value) {
     updateTokenProperty(
         token, update, TokenPropertyValueDto.newBuilder().setStringValue(value).build());
-  }
-
-  @Override
-  public void updateTokenProperty(Token token, Token.Update update, LightSource value) {
-    updateTokenProperty(
-        token,
-        update,
-        TokenPropertyValueDto.newBuilder().setLightSourceId(value.getId().toString()).build());
   }
 
   @Override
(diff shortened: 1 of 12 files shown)


FullBleed:
That was quick.  Looks like a good implementation.

Quick question...

If a token has a light source and is saved.  Then imported into a campaign without that light source, will it convert to a "unique" light source on the token?

kwvanderlinde:
> If a token has a light source and is saved. Then imported into a campaign without that light source, will it convert to a "unique" light source on the token?

No. Tokens don't actually store campaign lights, they just make a note saying that a certain light is activated. So at import time there's nothing available to convert. This PR doesn't change that.

cwisniew (approved):

bubblobill:
Damn you. I had just redesigned the whole config tab. \*sigh\*

PR #5076, opened 2024-11-29
Handle rptools-maptool: URI for quick-join links

### LINKED ISSUE #5075: [Feature]: A URI handler could streamline inviting players to connect to your server
### Describe the Problem

When starting a game with my group I have to announce via discord that I have started the server.

Automatically prompting my players to join would be ideal, but manually prompting with a link that contains everything they need to join that they can just click would make the process smoother.

Apart from making it a single click, this could include any additional connection parameters and remove the need to go back and forth explaining how to connect.

### The Solution you'd like

When the GM starts the server a URI is copied to their clipboard, they can go to connection information to find copyable text and buttons for links to connect.

_No response_

### Additional Context

_No response_

fishface60:

fixes: https://github.com/RPTools/maptool/issues/5075
depends on: https://github.com/RPTools/maptool/pull/5143

### Description of the Change

This change includes some refactoring.

It is structured in sequential commits that aim to be individually understandable, but later changes do replace earlier ones, so the commits are intended to make reading the changes more easily digestable, but review of the code should be performed on the end result.

The command-line parser is extended to handle positional arguments being URIs pointing to a running server instead of just file paths to a campaign file.

The URIs are parsed for connection information and if a valid server is provided MapTool will connect to it.

The packaging on Linux is extended to add URI handlers.

The Connection Info dialog is extended to add buttons for copying the URIs to the clipboard.

### Possible Drawbacks

A slight slowdown for parsing file paths as URI first.

If the feature doesn't prove useful then it's additional code to maintain.

Requires an additional service running on rptools.net to bounce from an https:// URL that can be pasted into chat and automatically turned into a link into the rptools-maptool: scheme URI.

#### `rptools-maptool://` URIs

When MapTool is installed it installs URI scheme handlers for:

* `rptools-maptool+registry:///serverName`
* `rptools-maptool+lan:///lanId`
* `rptools-maptool+tcp://address:port/`

When MapTool is started by clicking on a link with this URI it will connect to that server.

The URIs may be copied to your clipboard by clicking on the "Copy" buttons in the Connection Info dialog.

example

Chat apps such as Discord do not provide the ability to create your own links, and instead will only turn text that it recognises as a link into a clickable link, so the button that looks like a mouse cursor may be clicked to copy a URL that will open a page in your web browser and then redirect to the MapTool link.

### Release Notes

* Added support for starting MapTool and immediately connecting to a server specified in a custom URI so GMs can paste a link in chat that can prompt players to connect by clicking on it.
  * URI support is currently limited to Linux clients.
* Added buttons in the Connection Info dialog to copy pregenerated URIs and URLs.
* Fixed the Connection Info dialog listing two IPv6 addresses instead of an IPv4 and an IPv6 address if your system has an IPv6 address.
* Implemented a missing check that the address being displayed is reachable, by checking if it can connect to RPTools.net
* Fix a potential hang if all of the websites for determining external IP are down.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5076)


kwvanderlinde:
This is shaping up nicely! Loving the cleanup you've done along the way!

fishface60:
I added an abstraction over `net.tsc.servicediscovery.ServiceFinder` because I need to perform a discovery request outside of the connect dialog and didn't want to duplicate deserialize and type conversion logic.

In the process of investigating what authority might be valid for service discovery I discovered it is possible to bind it to a given interface, which would theoretically be a suitable authority, but it is useless for providing a connection link.

The other option for what might be a valid authority would be to make the multicast address configurable, which we probably don't want to make configurable but probably should change it from `net.tsc.servicediscovery`'s default, since the address it uses is assigned to the Teredo IPv6 tunnel.

From my reading of the Teredo RFC spec MapTool won't cause Teredo to break because it expects to get a valid response after seeing a multicast datagram, but if a service using Teredo happens to use the same port I'm not so sure `net.tsc.servicediscovery` won't be confused.

fishface60:
This has grown to be a bit of a monster.

I've split the features I discovered I wanted along the way into separate PRs and there's a common refactoring PR, but it's still huge.

I could split it down further if preferred, but having a bunch of very granular PRs may not be an improvement so I didn't do so in advance.

cwisniew (approved):

cwisniew:

Issue #5097, opened 2024-12-08
[Feature]: Move token layout panel to its own tab in edit token dialogue


bubblobill:
**EditTokenDialog**
1. Added checkboxes for Flip-X, Y, and Iso
2. Incorporated Unique Light Sources text area
3. Moved token layout panel to it's own tab

Four additional classes:
1. **SpinnerSliderPaired** ties a JSpinner and a JSlider together with some bells and whistles
2. **TokenLayoutRenderPanel** contains the paint method and handles mouse events
4. **TokenLayoutRenderPanelHelper** handles the initialisation code for control elements and a bunch of methods for creating and drawing various elements and onscreen guides
5. **MathUtils** a generic utils class for doing math stuff. Math operations in other classes should probably be moved to it too.

**GraphicsUtil** added method _createGridShape()_ for generating a grid shape consistently.

**ImageUtil** added a bunch of methods for operations to reduce ZoneRenderer code duplication:
1. getIsoFigureHeightOffset
2. getTokenImage - returns token image or table image if iso figure
3. flipCartesian - for flipped token images
4. flipIsometric - for flipped token images
6. getRenderingHintsQuality - a single source for a number of higher quality rendering hints
7. rotateImage - rotates an image
8. getScaledTokenImage - accounts for ZoneRenderer zoom level, token sizeScale, scaleX, scaleY and footprint size and scale
9. getTokenRenderImage - does most of these in appropriate order to maintain best image quality

kwvanderlinde:
I've found a few issues when using isometric features with the new renderer.

***

When changing a map to an iso grid, iso-flipped tokens are not rendered in iso until their facing changes or some other significant update happens. Same applies in reverse: changing to a square grid does not cause the token to stop rendering as iso.

E.g., these tokens have identical configurations, but one had an update since switching to iso:
![Image](https://github.com/user-attachments/assets/b72ff328-983b-496f-b685-91ff6cc907e2)

***

In non-iso grids, iso-flipped tokens are sometimes rendered as iso and sometimes not. E.g., when dragging a token, the token itself is not drawn iso, but the dragged version is:
![Image](https://github.com/user-attachments/assets/d529ef3f-3be6-43e9-a0a9-e9edec18aa2f)

***

Image tables are not respected unless on an iso grid.

***

The token layout panel renders the token as iso even when the grid is not iso:
![Image](https://github.com/user-attachments/assets/4cebeb7e-4c02-4b3d-82ae-b661348b9e7b)

[...]

Issue #5096, opened 2024-12-08
[Feature]: Improved and enhanced token layout options


bubblobill:
### Describe the Problem

I'm mostly frustrated when there was already an issue I said I would have a go at addressing and can never find it again.

Token layout is a tiny panel in a cramped tab of the edit token dialog. It is imprecise and almost impossible to use on anything of size greater then 1 map grid size. It also lacks access to control over the independant scaling of a token image in the horizontal and vertical axes as well as rotation of the token issue.

The dialog also fails to contain all the available UI at larger text sizes, often putting the Ok/Cancel buttons off screen.

### The Solution you'd like

A bigger, more accessible, more sophisticated, and easier to use GUI

### Alternatives that you've considered.

None

### Additional Context

Also ties-in to #997, #1447, #3424,  #4886 , #4934, and no doubt a bunch of others
```[tasklist]
### Tasks
- [ ] https://github.com/RPTools/maptool/issues/5101
- [x] https://github.com/RPTools/maptool/issues/5103
- [ ] https://github.com/RPTools/maptool/issues/5097
- [x] Split token painting out of ZoneRenderer into new support class TokenRenderer
```

PR #5104, opened 2024-12-11
Token class updates to accommodate new layout features

### LINKED ISSUE #5103: Update Token class for new layout features
Add token property for imageRotation.
Update macro functions getTokenLayoutProps and setTokenLayoutProps
Refactoring on Token:
Tidy up some token property names. 
Reorder layout related properties to be more or less grouped

bubblobill:
resolves #5103 
Progresses #5096 
Progresses #3691
Includes most of #5097 

### Description of the Change
New property
New property macro functions
Refactoring

### Possible Drawbacks
none

### Documentation Notes
n/a

### Release Notes
n/a

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5104)


bubblobill:
Yay, it runs, and I am reminded why methods should be properly named the first time.
Now I just have to rewrite ZoneRenderer

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/EditTokenDialog.java
```diff
@@ -940,10 +968,10 @@ public boolean commit() {
...
 
     // Jamz: TODO check if topology changed on token first
-    MapTool.getFrame()
-        .getCurrentZoneRenderer()
-        .getZone()
-        .tokenMaskTopologyChanged(token.getMaskTopologyTypes());
+    //    MapTool.getFrame()
+    //        .getCurrentZoneRenderer()
+    //        .getZone()
+    //        .tokenMaskTopologyChanged(token.getMaskTopologyTypes());
```
cwisniew: Please remove rather than just commenting out :)

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
```diff
@@ -0,0 +1,1317 @@
...
+  public void setParentRoot(JRootPane rp) {
+    parentRoot = rp;
+  }
+
+  private String helpText = assembleHelpText();
+
+  // @formatter:off
+  // spotless:off
+    // Component Getters
+    public JComboBox getSizeCombo(){ if (sizeCombo == null) sizeCombo = (JComboBox) parent.getComponent("size"); return sizeCombo; }
```
cwisniew: just format these as expected rather than turning off spotless :)

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
```diff
@@ -0,0 +1,1317 @@
...
+    }
+  }
+
+  ImageIcon[] scaleIcons = new ImageIcon[3];
+
+  private void createButtonIcons() {
+    scaleIcons[0] = makeSVGIcon(ICON_SIZE, scaleStartSVG + scaleNSSVG + scaleEWSVG + scaleEndSVG);
+    scaleIcons[1] = makeSVGIcon(ICON_SIZE, scaleStartSVG + scaleEWSVG + scaleEndSVG);
+    scaleIcons[2] = makeSVGIcon(ICON_SIZE, scaleStartSVG + scaleNSSVG + scaleEndSVG);
+  }
```
cwisniew: Its better to add these as resource files and add the functions to load them ResourceManager class, let me know if you need a hand doing this

cwisniew (changes requested):

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
bubblobill: Not until it gets out of draft status

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/EditTokenDialog.java
bubblobill: Points at` // private CharSheetController controller;`

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
bubblobill: Inclined to disagree. As you can see in these images. Dark themes do not change the icon colours like they are supposed to.
image
image
The FlatLaF docs are very poor when it comes to creating responsive SVG. So unless I can get it to work I will probably end up hard-coding the colours into the SVG at run-time.

cwisniew: https://github.com/JFormDesigner/FlatLaf/issues/402
You have to make sure you are using one of the defined colours, like the current icons are.

No embedding SVG strings into the code please it makes it too hard to edit them

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/EditTokenDialog.java
cwisniew: You can remove that one while you are there if you like :) feel free to point out others you find :)

bubblobill:
In order to get tokens to paint properly I had to mess with ZoneRenderer. The renderTokens() code in ZR is a nightmare of repetitious code and lots of fragile non-optimal stuff.

Since ZR needs major splitting anyway, I created TokenRenderer. At present it only deals with painting tokens and I left all the weird stuff I don't understand properly behind.
TokenRenderer checks if there have been changes to the token that would impact the rendered image (layout props) and only creates new images when required. 

No idea if it is faster, but image operations are performed in the following sequence with the intention of maintaining the best quality; Flip X/Y -> Flip Iso -> Scale[X, Y, XY, Footprint, and Zoom (all at once to prevent multiple scaling lossiness)] -> Rotate.

Image caches are cleared on change of zoom. Not sure if it is necessary but it seemed safer.

There are also problems with token bounds and locations. The whole everything works this way except for squares annoys the crap out of me as Everything I do is based on the token's actual position. Images are transformed such that the origin lies at their centre +/- any anchor. Translate the graphics object to the token position, scale and rotate as required and everything is easy.
With bounds, tokens have a bounding area defined by their footprint AND a bounding area that covers the token image. Think of a tree whose picture includes its canopy but the space occupied by the trunk is its footprint. Currently getBounds on a token returns the size of the whole image including offsets and scaling.
I seriously wanted to change this because it results in other things being drawn incorrectly using StG.

This is what happens if halos are drawn on offset images. Both tokens are StG. This is what the shark should look like.
image
But it snaps to the wrong place.
image

Since it has always been the case I didn't mess with it. But if all this gets through then I am mightily inclined to change it.

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
bubblobill: Done. Code uglified.

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
bubblobill: Icons added to resources

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/EditTokenDialog.java
bubblobill: Think I caught them all

src/main/java/net/rptools/lib/MathUtil.java
```diff
@@ -0,0 +1,150 @@
...
+ * <http://www.gnu.org/licenses/> and specifically the Affero license
+ * text at <http://www.gnu.org/licenses/agpl.html>.
+ */
+package net.rptools.lib;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/* Utility class for useful mathematical methods */
+public class MathUtil {
```
kwvanderlinde, line 21: We can do without most of this class as the methods are either unused or can be simplified. I'll leave specific comments below.

src/main/java/net/rptools/lib/MathUtil.java
```diff
@@ -0,0 +1,150 @@
...
+  private static final Logger log = LogManager.getLogger(MathUtil.class);
+
+  /**
+   * Faster version of absolute for integers
```
kwvanderlinde: Any proof of this claim? `Math.abs()` calls are always inlined with platform-specific instructions, so there's no branching concerns with it.

Anyways, this method isn't used and can be removed.

src/main/java/net/rptools/lib/MathUtil.java
```diff
@@ -0,0 +1,150 @@
...
+   *
+   * @param val
+   * @return absolute value of val
+   */
+  public static int abs(int val) {
+    return (val >> 31 ^ val) - (val >> 31);
+  }
+
+  /**
+   * Faster version of absolute
```
kwvanderlinde: I guarantee this is slower than `Math.abs()`. Plus this method isn't used.

src/main/java/net/rptools/lib/MathUtil.java
```diff
@@ -0,0 +1,150 @@
...
+   *
+   * @param valueToMap value in the first range that needs to be converted
+   * @param in_min the minimum value for the original range
+   * @param in_max the maximum value for the original range
+   * @param out_min the minimum value for the target range
+   * @param out_max the maximum value for the target range
+   * @return the equivalent value of valueToMap in the target range
+   * @param <T>
+   */
+  public static <T extends Number> T mapToRange(
```
kwvanderlinde: The generics don't really gain us anything here since internally it has to check `float` vs `double` anyways. I.e., an overload would do just as well.

But it can also just be replaced with a purely `double` version. The one caller that uses `float` can cast the result.

src/main/java/net/rptools/lib/MathUtil.java
```diff
@@ -0,0 +1,150 @@
...
+                      * (outMax.floatValue() - outMin.floatValue())
+                      / (inMax.floatValue() - inMin.floatValue())
+                  + outMin.floatValue());
+    } else {
+      result =
+          (Number)
+              ((mapValue.doubleValue() - inMin.doubleValue())
+                      * (outMax.doubleValue() - outMin.doubleValue())
+                      / (inMax.doubleValue() - inMin.doubleValue())
+                  + outMin.doubleValue());
```
kwvanderlinde: This kind of lerp implementation tends to be inaccurate when `valueToMap == in_max`, due to floating point shenanigans. An alternative with accurate endpoints would be:
```java
var t = (valueToMap - in_min) / (in_max - in_min);
result = (1 - t) * out_min + t * out_max;
```

src/main/java/net/rptools/lib/MathUtil.java
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * Constrains an integer between an upper and lower limit
+   *
+   * @param value
+   * @param lowBound
+   * @param highBound
+   * @return
+   */
+  public static int constrainInt(int value, int lowBound, int highBound) {
```
kwvanderlinde: This method is the same as `Math.clamp(long, int, int)` and can be removed in favour of that.

src/main/java/net/rptools/lib/MathUtil.java
```diff
@@ -0,0 +1,150 @@
...
+  /**
+   * Constrains a Number between an upper and lower limit
+   *
+   * @param value
+   * @param lowBound
+   * @param highBound
+   * @return
+   * @param <T>
+   */
+  public static <T extends Number> T constrainNumber(T value, T lowBound, T highBound) {
```
kwvanderlinde: This method is equivalent to `Math.clamp(double, double, double)` and can be removed.

src/main/java/net/rptools/lib/MathUtil.java
```diff
@@ -0,0 +1,150 @@
...
+   * @param decimalPlaces number of decimal places to use
+   * @return truncated double value
+   */
+  public static double doublePrecision(double value, int decimalPlaces) {
+    double d = Double.parseDouble(String.format("%." + decimalPlaces + "f", value));
+    log.debug("value: " + value + ", decimalPlaces: " + decimalPlaces + " -> " + d);
+    return d;
+  }
+
+  public static boolean isDouble(Object o) {
```
kwvanderlinde: With the other proposed simplifications, `isDouble()`, `isFloat()` and `isNumber()` are not needed.

src/main/java/net/rptools/lib/MathUtil.java
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * Returns a truncated double with the specified number of decimal places
+   *
+   * @param value to be truncated
+   * @param decimalPlaces number of decimal places to use
+   * @return truncated double value
+   */
+  public static double doublePrecision(double value, int decimalPlaces) {
+    double d = Double.parseDouble(String.format("%." + decimalPlaces + "f", value));
```
kwvanderlinde: This is fine for the UI usage, but outside of that will be a pretty slow implementation. Can use this instead:
```
double divisor = Math.pow(10, -decimalPlaces);
double d = divisor * Math.round(value / divisor);
```

kwvanderlinde:
I took this for a test and found a bug regarding halos. A token with no halo is shown as having the "white" halo when the menu is opened.
white-halo

src/main/java/net/rptools/maptool/model/Token.java
```diff
@@ -292,16 +287,15 @@ public enum Update {
...
-  private transient Color haloColor;
+  private transient Color haloColor =
+      new Color(ImageUtil.negativeColourInt(AppPreferences.defaultGridColor.getDefault().getRGB()));
 
   private Integer visionOverlayColorValue;
   private transient Color visionOverlayColor;
 
   // Jamz: allow token alpha channel modification
-  private @Nonnull Float tokenOpacity = 1.0f;
+  private @Nonnull Float opacity = 1.0f;
```
kwvanderlinde: Can't rename fields or token settings in existing campaigns will be lost (applies to `opacity`, `isFlippedX`, `isFlippedY`, `isFlippedIso`).

It's fine that the accessor methods or DTO fields are renamed, just the model fields need to remain the same.

kwvanderlinde (changes requested):

kwvanderlinde:
Got partyway through reviewing. Will pick it up again later today.

The feature is looking quite good. The new layout functionality feels very nice to use.

src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
```diff
@@ -929,24 +930,85 @@ public Object childEvaluate(
...
-      Double scale = token.getSizeScale();
-      int xOffset = token.getAnchorX();
-      int yOffset = token.getAnchorY();
+      String scale = String.valueOf(token.getSizeScale());
+      String xOffset = String.valueOf(token.getAnchorX());
+      String yOffset = String.valueOf(token.getAnchorY());
+      String rotation = String.valueOf(token.getImageRotation());
+      String scaleX = String.valueOf(token.getScaleX());
+      String scaleY = String.valueOf(token.getScaleY());
+      String footprintScaleValue =
```
kwvanderlinde, line 942: These should all be left as numeric values. Otherwise the JSON will contain a bunch of strings that will then require further parsing by the caller.

src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
```diff
@@ -929,24 +930,85 @@ public Object childEvaluate(
...
-        return "scale=" + scale + delim + "xOffset=" + xOffset + delim + "yOffset=" + yOffset;
+        StringBuilder sb = new StringBuilder();
+        sb.append("scale=" + scale + delim);
+        sb.append("xOffset=" + xOffset + delim);
+        sb.append("yOffset=" + yOffset + delim);
+        sb.append("rotation=" + rotation + delim);
+        sb.append("scaleX=" + scaleX + delim);
+        sb.append("scaleY=" + scaleY + delim);
+        sb.append("footprintScale=" + footprintScaleValue);
+        return sb.toString();
```
kwvanderlinde, line 965: Since you're using a `StringBuilder` anyways, might as well avoid the (expensive) string concats and chain the appends. E.g.,
```java
sb.append("scale=").append(scale).append(delim);
```

src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
```diff
@@ -929,24 +930,85 @@ public Object childEvaluate(
...
+      }
+    }
+    /*
+     * setExtendedTokenLayoutProps(StrProp/JSON Object, token: currentToken(), mapName = current map)
+     */
+    if (functionName.equalsIgnoreCase("setExtendedTokenLayoutProps")) {
```
kwvanderlinde: Anyway this could be folded into `setTokenLayoutProps()`? The set of options parallels the output of `getTokenLayoutProps()` and so would be easier for macro writers to discover.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/tokenRender/TokenRenderer.java
kwvanderlinde, line 1: I'm not sure what changed, but I'm finding that "Flip > Isometric Plane" doesn't result in correct rendering. The token is typically rendered flat, but once it starts being dragged it becomes iso.
flip-iso

src/main/java/net/rptools/lib/image/ImageUtil.java
```diff
@@ -47,64 +42,135 @@
...
+      // combine with token scale properties
+      if (sX != 1 || sY != 1 || sXY != 1 || imageFootprintRatio != 1) {
+        int outputWidth = (int) Math.ceil(imgW * sXY * sX * imageFootprintRatio);
+        int outputHeight = (int) Math.ceil(imgH * sXY * sY * imageFootprintRatio);
+        token.setWidth(outputWidth);
+        token.setHeight(outputHeight);
+        try {
+          return ImageUtil.scaleBufferedImage(img, outputWidth, outputHeight);
+        } catch (Exception e) {
+          log.debug(e.getLocalizedMessage(), e);
```
kwvanderlinde: Is there an actual error that we should be concerned about here? At a glance, I don't think the `ResampleOp` will throw anything in our case.

If there is a worry, this should be a `log.error()`, or at least a `log.warn()`.

src/main/java/net/rptools/lib/image/ImageUtil.java
```diff
@@ -47,64 +42,135 @@
...
+          return img;
+        }
+      }
+    } else {
+      Rectangle b = token.getBounds(grid.getZone());
+      try {
+        return ImageUtil.scaleBufferedImage(
+            img, (int) Math.ceil(b.width * zoom), (int) Math.ceil(b.height * zoom));
+      } catch (Exception e) {
+        log.debug(e.getLocalizedMessage(), e);
```
kwvanderlinde: Same as above, either don't handle the error or make it `log.error()`.

src/main/java/net/rptools/lib/image/ImageUtil.java
```diff
@@ -389,20 +407,41 @@ public static BufferedImage createOutline(BufferedImage sourceImage, Color color
...
   }
 
+  /**
+   * Returns the provided image flipped according to the token's flip-states
+   *
+   * @param image The image to be processed
+   * @param token The token containing the flip-states
+   * @return A modified image, or the original image if no processing performed.
+   */
+  public static BufferedImage flipTokenImage(BufferedImage image, Token token) {
```
kwvanderlinde: This method isn't used.

src/main/java/net/rptools/lib/image/ImageUtil.java
```diff
@@ -415,21 +454,203 @@ public static BufferedImage flip(BufferedImage image, int direction) {
...
+      throw new IOException("Image not found: " + image);
+    }
+    inStream = new BufferedInputStream(inStream);
+    while ((bite = inStream.read()) >= 0) {
+      dataStream.write(bite);
+    }
+    return bytesToImage(dataStream.toByteArray(), image);
+  }
+
+  public static double getIsoFigureHeightOffset(Token token, Rectangle2D footprintBounds) {
```
kwvanderlinde, line 517: `getIsoFigureHeightOffset()` and `getIsoFigureScaleFactor()` seem like they would be more natural in the `Token` class as they don't really have to do with image manipulation even though they are fed into token image rendering.

kwvanderlinde (changes requested):

src/main/java/net/rptools/lib/MathUtil.java
bubblobill: Agreed. I originally created it to house mathematical methods that I found lurking in other classes that I could never find when I wanted them. Then I gave up trying to find existing methods and just started adding methods as they became useful. Some of which I stopped using as my code developed, as you discovered.

src/main/java/net/rptools/lib/MathUtil.java
bubblobill: Yes there is evidence, probably buried in StackExchange. Although it is probably easier to just use FastMath, and as you noticed, my need for it disappeared.

src/main/java/net/rptools/lib/MathUtil.java
bubblobill: \<whine>But I don't like casting and like generics... \</whine>

src/main/java/net/rptools/lib/MathUtil.java
bubblobill: good thinking 99

src/main/java/net/rptools/lib/MathUtil.java
bubblobill: wilco

src/main/java/net/rptools/lib/MathUtil.java
bubblobill: wilco

src/main/java/net/rptools/lib/MathUtil.java
bubblobill: spoilsport

src/main/java/net/rptools/lib/MathUtil.java
bubblobill: Where were you when I was researching for a fast and reliable Math.trunc?

src/main/java/net/rptools/maptool/model/Token.java
bubblobill: yeah... found that out the hard way.

src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
bubblobill: roger

src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
bubblobill: roger

src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
bubblobill: mebbe. it's been so long I will have to revisit and work out why I did it this way.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/tokenRender/TokenRenderer.java
bubblobill: Probably ignoring a condition somewhere like token shape

src/main/java/net/rptools/lib/image/ImageUtil.java
bubblobill: IIRC, it was bugging out on images with extreme aspect ratios

src/main/java/net/rptools/lib/image/ImageUtil.java
bubblobill: you and your semantic correctness

src/main/java/net/rptools/lib/image/ImageUtil.java
bubblobill: disappointing after all the time I spend writing methods that I find they already exist somewhere else.

src/main/java/net/rptools/lib/image/ImageUtil.java
bubblobill: Their only value is for working out where to draw the image and were extracted from the guts of zoneRenderer. Personally I think Token is already bloated and want it to only have the x/y of the image centre/footprint origin and separate all the bits for rendering. They probably belong in Grid or Footprint. I'll look into their utility before considering a relocation.

bubblobill:
Implemented the majority of requested changes, but too much time has passed and now there are conflicts I don't know how to resolve.

src/main/java/net/rptools/lib/image/ImageUtil.java
bubblobill: They are used in separate places which is why I brought them together into the utility class.

kwvanderlinde:
> Implemented the majority of requested changes, but too much time has passed and now there are conflicts I don't know how to resolve.

I can help you out with this. Does your `TokenLayout` branch have all your latest changes? If so, I can pull it, merge with `develop` and resolve these conflicts.

src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
bubblobill: Had a look to find out why I made it separate. 
`setTokenLayoutProps()` has a fixed arg list `setTokenLayoutProps(scale, xOffset, yOffset, token, map)`.
dealing with a bunch of optional arguments would have been messy `(scale, xOffset, yOffset, scaleX, scaleY, rotation, token, map)` and made it even harder to add future changes I have considered like skew and perspective adjustmenets.

Thus I opted to replace it with a function that performs the same actions but can accommodate the new and possible future changes by simply taking `(props_map, token, map)`.

The name needs help, I think I will rename it to `setLayoutProps`

bubblobill:
> > Implemented the majority of requested changes, but too much time has passed and now there are conflicts I don't know how to resolve.
> 
> I can help you out with this. Does your `TokenLayout` branch have all your latest changes? If so, I can pull it, merge with `develop` and resolve these conflicts.

Implementing the changes broke the layout panel. The pairing of spinners and sliders connected to wheel events no longer works, and I now have disproportionate movement when dragging the token image.

Methinks your requested changes are questionable :)

src/main/java/net/rptools/maptool/client/ui/zone/renderer/tokenRender/TokenRenderer.java
bubblobill: This is part of the existing code net.rptools.maptool.client.ui.zone.renderer.ZoneRenderer#showBlockedMoves that does not yet use TokenRenderer, i.e. it wasn't me.

bubblobill:
> > Implemented the majority of requested changes, but too much time has passed and now there are conflicts I don't know how to resolve.
> 
> I can help you out with this. Does your `TokenLayout` branch have all your latest changes? If so, I can pull it, merge with `develop` and resolve these conflicts.

Think I have fixed everything. TokenLayout branch now has a (theoretical) publishable version.

kwvanderlinde:
> > > Implemented the majority of requested changes, but too much time has passed and now there are conflicts I don't know how to resolve.
> > 
> > 
> > I can help you out with this. Does your `TokenLayout` branch have all your latest changes? If so, I can pull it, merge with `develop` and resolve these conflicts.
> 
> Think I have fixed everything. TokenLayout branch now has a (theoretical) publishable version.

I've resolved the merge conflicts. Had to do it with the GitHub editor though, so didn't have access to spotless. @bubblobill Mind running `spotlessApply`?

In the meantime I'll review the updates.

bubblobill:
> I've resolved the merge conflicts. Had to do it with the GitHub editor though, so didn't have access to spotless. @bubblobill Mind running `spotlessApply`?

done

src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
```diff
@@ -113,7 +113,8 @@ private TokenPropertyFunctions() {
         "getNotes",
         "setNotes",
         "getTokenLayoutProps",
-        "setTokenLayoutProps",
+        "setLayoutProps",
+        "setExtendedTokenLayoutProps",
```
kwvanderlinde: We'll need `setTokenLayoutProps` to remain for backwards compatibility.

I assume you meant to rename `setExtendedTokenLayoutProps` to `setLayoutProps` instead?

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
```diff
@@ -0,0 +1,1341 @@
...
+            getZoomSlider(),
+            null,
+            percentSpinnerToSlider,
+            percentSliderToSpinner);
+    /* limit zoom to minimum 34%, maximum 300%. Spinner changes are doubles, Slider changes are integers. */
+    zoomPair.addVetoableChangeListener(
+        evt -> {
+          log.info(evt);
+          if (evt.getPropertyName().toLowerCase().contains("value")
+              && evt.getNewValue().getClass().isAssignableFrom(Double.class)
```
kwvanderlinde: This `isAssignableFrom()` check is backwards. But it can also be cleared up by replacement with `instanceof`:
```
          if (evt.getPropertyName().toLowerCase().contains("value")
              && evt.getNewValue() instanceof Number newValue
              && (newValue.doubleValue() < 0.34 || newValue.doubleValue() > 3d)) {
            ...
          }
```

Similar thing for `scalePair` above.

kwvanderlinde (changes requested):
The code itself looks pretty much good from a couple minor things.

But there is some weirdness still going on with the layout panel related to the sliders and spinners:
1. Dragging any of the sliders do not cause the panel itself to update. I have to right-click or otherwise interact with the panel to get it to update. Sometimes I get torn rendering (partiial repaint) instead of a complete update.
2. If I drag the zoom slider all the way down, exceptions are shown in a dialog saying `width and height must be positive`. So presumably this value needs to be clamped somewhere above 0 before being used.
3. Similar thing with bringing the scale slider all the way down. An exception is logged (not shown to the user), and the image snaps back to full size.

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
```diff
@@ -0,0 +1,1341 @@
...
+    okButton = b;
+  }
+
+  public void setParentRoot(JRootPane rp) {
+    parentRoot = rp;
+  }
+
+  private final String helpText = assembleHelpText();
+
+  /* Component Getters */
```
cwisniew: The type of comment that is not really needed as it provides no information about what the code is doing, and is likely to be inaccurate when other people edit the code / use IDE to create new getter

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
```diff
@@ -0,0 +1,1341 @@
...
+    return sizeCombo;
+  }
+
+  /* Labels */
+  public JLabel getScaleLabel() {
+    if (scaleLabel == null) scaleLabel = parent.getLabel("scaleLabel");
+    return scaleLabel;
+  }
+
+  /* Spinners */
```
cwisniew: The type of comment that is not really needed, and is likely to be inaccurate when other people edit the code / use IDE to create new getter, I will stop repeating this comment but it applies for all instances in this class

src/main/java/net/rptools/maptool/client/swing/SpinnerSliderPaired.java
```diff
@@ -0,0 +1,555 @@
...
+            + getMaximum()
+            + ", val="
+            + getValue()
+            + ", stepSize="
+            + getStepSize()
+            + '}';
+      }
+    }
+
+    public NumericSliderModel sliderModelDelegate = new NumericSliderModel(0, 0, 0, 100);
```
cwisniew: The only public fields should be constants

src/main/java/net/rptools/maptool/client/swing/SpinnerSliderPaired.java
```diff
@@ -0,0 +1,555 @@
...
+            + super.getExtent()
+            + ", isAdjusting="
+            + super.getValueIsAdjusting()
+            + '}';
+      }
+    }
+
+    // spotless:on
+    // @formatter:on
+    @Override
```
cwisniew: I can't see where spotless was turned off... In any case if it was it shouldn't have been :)

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/EditTokenDialog.java
```diff
@@ -211,7 +231,7 @@ public void closeDialog() {
     validateLibTokenURIAccess(getNameField().getName());
     var combo = getStatSheetCombo();
     combo.removeAllItems();
-    // Default Entry
+    /* Default Entry */
```
cwisniew, line 234: um... why change from // to /* for single line comments? The same goes for all the instances below

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
```diff
@@ -0,0 +1,1341 @@
...
+      return getTokenSizeScale();
+    } else if (scaleAxis == 1) {
+      return getTokenScaleX();
+    } else {
+      return getTokenScaleY();
+    }
+  }
+
+  /* used to determine which scale is being edited: 0 is XY, 1 is X, 2 is Y */
+  private int scaleAxis = 0;
```
cwisniew, line 339: This would be clearer as an enum

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
```diff
@@ -0,0 +1,1341 @@
...
+        .addFocusListener(focusListener);
+    ((JSpinner.NumberEditor) getRotationSpinner().getEditor())
+        .getTextField()
+        .addFocusListener(focusListener);
+    ((JSpinner.NumberEditor) getScaleSpinner().getEditor())
+        .getTextField()
+        .addFocusListener(focusListener);
+  }
+
+  public void initButtons() {
```
cwisniew: Only make it public if you are calling it from other classes (and then classes outside of the package / inheritance hierarchy). Same comment for all the Other init*() methods.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/TokenLocation.java
```diff
@@ -20,7 +20,7 @@
 import net.rptools.lib.CodeTimer;
 import net.rptools.maptool.model.Token;
 
-class TokenLocation {
+public class TokenLocation {
```
cwisniew: Does this need to be made public? (I may have missed where you are creating/using it outside of the package). Someone make it package scope intentionally

src/main/java/net/rptools/maptool/util/GraphicsUtil.java
```diff
@@ -449,4 +445,60 @@ protected void paintComponent(Graphics g) {
     f.add(p);
     f.setVisible(true);
   }
+
+  public static Shape createGridShape(String gridType, double size) {
```
cwisniew: This should all be delegated to the Grid classes

src/main/java/net/rptools/maptool/client/swing/SpinnerSliderPaired.java
```diff
@@ -0,0 +1,555 @@
...
+    private Number getModelNumber(boolean asInt) {
+      if (asInt) {
+        return spinnerToSlider.apply(spinnerModelDelegate.getNumber());
+      } else {
+        return spinnerModelDelegate.getNumber();
+      }
+    }
+
+    // Delegate classes with additional setters and normal setters redirected
+    public NumericSpinnerModel spinnerModelDelegate = new NumericSpinnerModel(0d, 0d, 100d, 1d);
```
cwisniew: Non constant fields should not be public

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
```diff
@@ -0,0 +1,1341 @@
...
+  Grid grid = MapTool.getFrame().getCurrentZoneRenderer().getZone().getGrid();
+  RenderBits renderBits = new RenderBits();
+  boolean isoGrid = grid.isIsometric();
+  boolean noGrid = GridFactory.getGridType(grid).equals(GridFactory.NONE);
+  boolean squareGrid = GridFactory.getGridType(grid).equals(GridFactory.SQUARE);
+  boolean hexGrid = false;
+  boolean isIsoFigure = false;
+  int gridSize = grid.getSize();
+  double cellHeight = grid.getCellHeight();
+  double cellWidth = grid.getCellWidth();
```
cwisniew: Are the access levels here (and elsewhere in the class) correct? Did you mean for them to be available to all classes in the same package? or should they be private?

src/main/java/net/rptools/maptool/model/Token.java
```diff
@@ -2217,6 +2128,7 @@ public String toString() {
     return "Token: " + id;
   }
 
+  // Token Layout - Getters/Setters
```
cwisniew: Not useful and will become inaccurate quickly

src/main/java/net/rptools/maptool/model/Token.java
```diff
@@ -2227,27 +2139,127 @@ public Point getAnchor() {
...
+  /**
+   * Facing is in the map space where 0 degrees is along the X axis to the right and proceeding CCW
+   * for positive angles.
+   *
+   * <p>Round/Square tokens that have no facing set, return null. Top Down tokens default to -90.
+   *
+   * @return null or angle in degrees
+   */
+  public int getFacing() {
+    // -90° is natural alignment. TODO This should really be a per grid setting
```
cwisniew: Please don't add TODO comments into the code, if its really a TODO add an issue

src/main/java/net/rptools/maptool/model/Token.java
```diff
@@ -2269,8 +2281,8 @@ public TokenTransferData toTransferData() {
     td.put(TokenTransferData.ASSET_ID, imageAssetMap.get(null));
     td.put(TokenTransferData.Z, z);
     td.put(TokenTransferData.SNAP_TO_SCALE, snapToScale);
-    td.put(TokenTransferData.WIDTH, scaleX);
-    td.put(TokenTransferData.HEIGHT, scaleY);
+    td.put(TokenTransferData.WIDTH, scaleX); // this makes no sense
```
cwisniew: You are correct; adding this comment makes no sense at all, If you think its a bug test with the correct value to see result or add an issue

src/main/java/net/rptools/maptool/model/Token.java
```diff
@@ -2269,8 +2281,8 @@ public TokenTransferData toTransferData() {
...
+    td.put(TokenTransferData.HEIGHT, scaleY); // ditto
```
cwisniew: ditto :)

cwisniew (changes requested):

src/main/java/net/rptools/maptool/client/functions/TokenPropertyFunctions.java
bubblobill: > I assume you meant to rename setExtendedTokenLayoutProps to setLayoutProps instead?

Correct. It seems one cannot truly trust find/replace

bubblobill: https://github.com/RPTools/maptool/pull/5104/commits/17efdde4e4b48fae17d02489568974fbd66bb254

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
bubblobill: from the comment `/* limit scale to minimum 10%. Spinner changes are doubles, Slider changes are integers. */`, so a Number check won't work. This is actually checking the source of the event; if it is a Double it comes from a spinner, if it is an Integer it comes from a slider.
Since sliders and spinners update each other this will be fired multiple times so we only want to veto one of them.

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/EditTokenDialog.java
bubblobill: Because someone sent me on a mission to find commented out code, so I changed most single line comments to block comments to make it easier to search the code-base for `//`

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
kwvanderlinde: :facepalm: Apparently I can't read.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/TokenLocation.java
bubblobill: The decomposition of ZoneRenderer is naturally going to lead to urban sprawl. It makes sense (to me) to package things further to keep related stuff together. 
```
zone -> renderer -> tokenRendering  -> TokenRenderer
                                                             -> FacingArrowRenderer
                                                             -> HaloRendering
                             -> lighting ->
                             -> stuff
```
If it isn't public, descendant packages can't read it. 
TokenRenderer uses a private instance for class use. With no public TokenLocation it gets nerfed. 
[Hence this](https://discord.com/channels/296230822262865920/494598206626463755/1318088889457573969)

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
bubblobill: Most of them were shared at some point as I slowly split out the helper functions from the panel class.
So not in the least bit correct any more.

src/main/java/net/rptools/maptool/client/swing/SpinnerSliderPaired.java
bubblobill: See this tongue, see how it is all vibrating, surrounded by eructations, and accompanied by an displeasing sound? That is all for you.

bubblobill:
> The code itself looks pretty much good from a couple minor things.
> 
> But there is some weirdness still going on with the layout panel related to the sliders and spinners:
> 
> 1. Dragging any of the sliders do not cause the panel itself to update. 
I appear to have failed to implement a change listener on the slider.

... Sometimes I get torn rendering (partiial repaint) instead of a complete update.
Odd, repaint should be performed on the control and the display panel. Hopefully the shiny new change listener will cause the region to be marked dirty and fix this.

> 2. If I drag the zoom slider all the way down, exceptions are shown in a dialog saying `width and height must be positive`. So presumably this value needs to be clamped somewhere above 0 before being used.
Remember asking me why I had error handling on the image scaling op?

> 3. Similar thing with bringing the scale slider all the way down. An exception is logged (not shown to the user), and the image snaps back to full size.
The full size thing is me just returning the original image when catching the error. The vetoed change should be catching this before it goes to rescaling.

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
bubblobill: Gotcha, remove road signs to confuse the enemy.

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
bubblobill: This is inherited from ImageUtil.flip(BufferedImage image, int direction) which I renamed to flipCartesian.
Should the enum go there, or because it is fundamentally a property of a token's deisplay characteristics, should it live on token?
Because, you know, someone told me we can't change Token right now.

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
bubblobill: Fixed

src/main/java/net/rptools/maptool/util/GraphicsUtil.java
bubblobill: I disagree. 

1. The shapes generated by the Grid classes are woefully inadequate for close-up work because they use integer coordinates. It was one of my gripes when working up a GridRenderer, Integers do not make for pretty lines.
2. Without an instance of the Grid, you cannot access the grid shape. Creating a Grid in order to draw it's cell shape in a panel is overkill. IIRC The footprint editor does this with an instance of every Grid type.
3. I would concede it perhaps being a static method in Grid if anything. This would allow grids to summon a shape instead of having to rewrite the cell geometry creation code for each class.

cwisniew: > I disagree.
> 
> 1. The shapes generated by the Grid classes are woefully inadequate for close-up work because they use integer coordinates. It was one of my gripes when working up a GridRenderer, Integers do not make for pretty lines.

Who said anything about them having to be integers just because they are part of the grid class?

> 2. Without an instance of the Grid, you cannot access the grid shape. Creating a Grid in order to draw it's cell shape in a panel is overkill. IIRC The footprint editor does this with an instance of every Grid type.
> 3. I would concede it perhaps being a static method in Grid if anything. This would allow grids to summon a shape instead of having to rewrite the cell geometry creation code for each class.

If you don't want to put them in the grid classes you could I guess create other classes where there is one for every grid type and a factory creates them, but is that more work then in grid? Certianly we shouldn't be creating a utility class that has a single function to do this that needs to be modified any time grids are added or modified.

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
cwisniew: > This is inherited from ImageUtil.flip(BufferedImage image, int direction) which I renamed to flipCartesian. Should the enum go there, or because it is fundamentally a property of a token's deisplay characteristics, should it live on token? Because, you know, someone told me we can't change Token right now.

I mean you have already made other changes to the Token class :)

src/main/java/net/rptools/maptool/client/ui/zone/renderer/TokenLocation.java
cwisniew: > Hence this

But it doesn't as they are two different things and would not use the same class if it were public

src/main/java/net/rptools/maptool/client/swing/SpinnerSliderPaired.java
cwisniew: > See this tongue, see how it is all vibrating, surrounded by eructations, and accompanied by an displeasing sound? That is all for you.

Just let me know if I should be purchasing a perspex shield

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
bubblobill: That was before you said stop. When you did I thought this PR was dead in the water.

src/main/java/net/rptools/maptool/util/GraphicsUtil.java
bubblobill: Moved it to Grid.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/TokenLocation.java
bubblobill: Renamed to TokenPosition

src/main/java/net/rptools/maptool/client/ui/token/dialog/edit/TokenLayoutPanelHelper.java
kwvanderlinde: It's fine to modify logic in `Token` and other model classes. As long as there aren't changes to the fields and we're not storing different concrete types in the fields, then saving will not be affected.

cwisniew: > It's fine to modify logic in `Token` and other model classes. As long as there aren't changes to the fields and we're not storing different concrete types in the fields, then saving will not be affected.

There are new fields for facing, scale, and rotation, so anything modified will possibly not appear the same when loading into an old campaign, so this change is already excluded from 1.18, so it can't hurt to make the other change its already going to be delayed to 1.19

Unless these shouldn't be persisted, in which case we should make them transient

bubblobill: Moved to AppConstants as an Enum

PR #5107, opened 2024-12-16
Fix error loading HeroLab data with no images

### LINKED ISSUE #4945: [Bug]: heroImageAssets causes campaign to fail to load
### Describe the Bug

Untitled
[maptool.log](https://github.com/user-attachments/files/17096703/maptool.log)

When loading a campaign on a new computer without having the por files in the same location it results in the campaign being unable to load.

regression from 1.8.3 where this did not happen.

### To Reproduce

Load campaign containing por files after moving them

### Expected Behaviour

Load without images, not crash

### Screenshots

_No response_

### MapTool Info

1.15

### Desktop

_No response_

### Additional Context

_No response_

cwisniew:
Fixes #4945 

### Description of the Change
Fixes the case where HeroLab data fails to load in a campaign without image assets.
Also fixes up formatting

### Possible Drawbacks
Should be none

### Release Notes
- Fixes the case where HeroLab data fails to load in a campaign without image assets.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5107)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/HeroLabData.java
+++ b/src/main/java/net/rptools/maptool/model/HeroLabData.java
@@ -138,7 +138,6 @@ public class HeroLabData {
     XML xmlObj = new XMLDocument(getStatBlock_xml());
     results = String.join(delim, xmlObj.xpath(xPathExpression));
 
-
     return results;
   }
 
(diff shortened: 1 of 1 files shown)


bubblobill (approved):
Approved

PR #5126, opened 2024-12-27
Support Ephemeral ports

### LINKED ISSUE #5125: [Feature]: Support starting MapTool on an ephemeral port instead of configuring in advance
### Describe the Problem

When using UPnP or hosting a LAN game or connecting to other players through a VPN it is not necessary to use a fixed port, as there is another name used to identify the server.

My GURPS group rotate GMs, one of the other GMs is in my household so we have to coordinate to create our servers with a different port and have different port-forwarding configured. If an ephemeral port were mapped with UPnP we wouldn't need to worry about coordinating.

### The Solution you'd like

UPnP is fixed for my router (see https://github.com/RPTools/maptool/pull/5124 for the challenges in that) and passing port 0 into the server port field is supported.

_No response_

### Additional Context

_No response_

fishface60:

Closes https://github.com/RPTools/maptool/issues/5125

Depends on https://github.com/RPTools/maptool/pull/5143

### Description of the Change

This relaxes the validation of the server port field to allow 0, changes code that used the configured port for sending to the registry or advertising with service discovery to use the value from the bound socket, and ensures they are configured after the socket has been bound so the port can be correctly determined.

### Possible Drawbacks

If a user specifies it as 0 without using UPnP then connections will probably fail and those error messages are less informative than having picked an invalid port.

### Release Notes

* The server port may be specified as 0 to use a random port that may be mapped with UPnP and configured with the rptools.net server registry or advertised on LAN.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5126)

### DIFF
--- a/clientserver/src/main/java/net/rptools/clientserver/simple/server/AbstractServer.java
+++ b/clientserver/src/main/java/net/rptools/clientserver/simple/server/AbstractServer.java
@@ -20,7 +20,7 @@ import net.rptools.clientserver.simple.connection.Connection;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
-public abstract class AbstractServer implements Server {
+public abstract class AbstractServer {
 
   private static final Logger log = LogManager.getLogger(AbstractServer.class);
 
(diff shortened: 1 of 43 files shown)


cwisniew (approved):

cwisniew:

PR #5143, opened 2025-01-19
Retain serviceIdentifier in MapToolServer objects

### LINKED ISSUE #5142: [Refactoring]: Retain Server ID in MapToolServer class
### Describe the problem

The MapToolServer class constructor passes the ID to the ServiceAnnouncer, but the ServiceAnnouncer doesn't have an API to retrieve it, so if another part of the code needs to identify what the ID that's being announced is, such as potentially:

1. Macros
2. The Connection Info dialog could display it
3. My proposed URI schema feature https://github.com/RPTools/maptool/issues/5075 would have a LAN variant for connecting via discovery.

currently they would have to know that it happens to be the name of the GM of the server.

MapToolServer retains this information and exposes it via a getter.

### Expected Benefits

There would be an appropriate API I can use to implement this.
Potentially this opens up the possibility of an implementation that sends out a service discovery request to the LAN and if it detects something with the same name adapts its server ID for something unique.

### Additional Context

_No response_

fishface60:

fixes https://github.com/RPTools/maptool/issues/5142
depended on by https://github.com/RPTools/maptool/pull/5126
depended on by https://github.com/RPTools/maptool/pull/5076

### Description of the Change

This retains the service ID that's announced by the ServiceIdentifier in the MapToolServer class, so that it can be returned elsewhere.

### Possible Drawbacks

`serviceIdentifier` is a bit opaque compared to LAN ID, but service is the nomenclaiture of the ServiceAnnouncer and `Identifier` has more consistent capitalisation rules than `ID`, as older code expects acronyms to remain capitalised and would be `serviceID`, but the rust style is also popular in the codebase and would be `serviceId`.

### Documentation Notes

Not applicable

### Release Notes

Not applicable

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5143)

### DIFF
--- a/src/main/java/net/rptools/maptool/server/MapToolServer.java
+++ b/src/main/java/net/rptools/maptool/server/MapToolServer.java
@@ -21,6 +21,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Random;
 import java.util.concurrent.atomic.AtomicBoolean;
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.swing.SwingUtilities;
 import net.rptools.clientserver.ConnectionFactory;
@@ -67,6 +68,7 @@ public class MapToolServer {
     Stopped
   }
 
+  @Nonnull private final String serviceIdentifier;
   private final Server server;
   private final MessageHandler messageHandler;
   private final Router router;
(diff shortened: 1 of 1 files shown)


cwisniew (approved):

cwisniew:

PR #5151, opened 2025-01-26
Use a variable sized thread pool in AssetManager

### LINKED ISSUE #4962: [Bug]: Add-ons with javascript assets can cause clients on the same machine connect to hang indefinitely or exception when loaded
### Describe the Bug

When connecting a client on the same machine I can reliably get MapTool to indefinitely wait for an asset to load.

When creating a new campaign without the add-on loaded, connecting the client, and then loading the add-on sometimes an exception is shown.

### To Reproduce

1. Download [assets.zip](https://github.com/user-attachments/files/17196159/assets.zip)
2. Start MapTool
3. Import the properties, add the add-on and drag the token onto the map.
4. Start a server
5. Start MapTool again
6. Connect MapTool to the local server as a player
7. Observe the hang

Alternatively:

2. Start MapTool
3. Import the properties and drag the token onto the map
4. Start a server
5. Start MapTool again
6. Connect MapTool to the local server as a player
7. Add the add-on
8. Repeat the process until the following error is logged to console and a pop-up is shown.

```
19:42:36.391 (MapTool.java:298) [SocketConnection.ReceiveThread] ERROR - Error receiving game data update.
java.util.concurrent.ExecutionException: net.rptools.maptool.model.gamedata.InvalidDataOperation: Data Namespace addon: does not exist for type com.gitlab.richardmaw.mt-gurps.
        at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396) ~[?:?]
        at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2073) ~[?:?]
        at net.rptools.maptool.model.gamedata.GameDataImporter.importData(GameDataImporter.java:97) ~[main/:?]
        at net.rptools.maptool.client.ClientMessageHandler.handle(ClientMessageHandler.java:449) [main/:?]
        at net.rptools.maptool.client.ClientMessageHandler.handleMessage(ClientMessageHandler.java:151) [main/:?]
        at net.rptools.clientserver.simple.connection.AbstractConnection.dispatchMessage(AbstractConnection.java:109) [clientserver.jar:?]
        at net.rptools.clientserver.simple.connection.AbstractConnection.dispatchCompressedMessage(AbstractConnection.java:115) [clientserver.jar:?]
        at net.rptools.clientserver.simple.connection.SocketConnection$ReceiveThread.run(SocketConnection.java:164) [clientserver.jar:?]
Caused by: net.rptools.maptool.model.gamedata.InvalidDataOperation: Data Namespace addon: does not exist for type com.gitlab.richardmaw.mt-gurps.
        at net.rptools.maptool.model.gamedata.InvalidDataOperation.createNamespaceDoesNotExist(InvalidDataOperation.java:110) ~[main/:?]
        at net.rptools.maptool.model.gamedata.MemoryDataStore.setData(MemoryDataStore.java:188) ~[main/:?]
        at net.rptools.maptool.model.gamedata.MemoryDataStore.lambda$setProperty$9(MemoryDataStore.java:212) ~[main/:?]
        at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768) ~[?:?]
        at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1760) ~[?:?]
        at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387) ~[?:?]
        at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312) ~[?:?]
        at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1843) ~[?:?]
        at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808) ~[?:?]
        at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188) ~[?:?]
19:42:36.408 (I18N.java:181) [SocketConnection.ReceiveThread] DEBUG - Cannot find key 'Error receiving game data update.<br/>java.util.concurrent.ExecutionException: net.rptools.maptool.model.gamedata.InvalidDataOperation: Data Namespace addon: does not exist for type com.gitlab.richardmaw.mt-gurps.' in properties file.
```

### Expected Behaviour

Running a second MapTool instance locally to test player client behaviour neither hangs nor displays error messages.

### Screenshots

_No response_

### MapTool Info

First observed in v1.14.3 but present in develop

### Desktop

Ubuntu 24.04.1 LTS

### Additional Context

Current hypothesis for the exception is that the asset raced the task for creating the add-on and the asset was loaded before the add-on was created.

fishface60:
This fix is targeted against 1.17, for the version targeting 1.16 see https://github.com/RPTools/maptool/pull/5153

fixes https://github.com/RPTools/maptool/issues/4962

### Description of the Change

AssetManager has a thread pool for loading assets that prevents the network handler thread from blocking when it receives a message that requires an asset load.

It is a single thread, which effectively serialises loading. 

Unfortunately this means that if the appropriate action after loading an asset is to load a subsequent asset (such as AddOns loading the onInit script in order to run them when the AddOn has been loaded) then this deadlocks.

If the asset is not already cached then instead of the load callbacks being called in the loader thread the assetLoader class' thread pool handles the callbacks, so it should be safe to use a larger thread pool to load assets.

This patch switches the single thread pool with a reference to the `ForkJoinPool.commonPool()` which is used by CompletableFutures as the "default" thread pool.

### Possible Drawbacks

Supporting parallel loading of assets could potentially have observable changes that cause other bugs, but there are circumstances already that allow assets to load in parallel, so these code paths aren't untested.

### Release Notes

* Fixed a deadlock when joining a server if it included an AddOn that had an onInit script and the AddOn was already cached.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5151)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/AssetManager.java
+++ b/src/main/java/net/rptools/maptool/model/AssetManager.java
@@ -39,6 +39,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ForkJoinPool;
 import net.rptools.lib.FileUtil;
 import net.rptools.lib.MD5Key;
 import net.rptools.maptool.client.AppUtil;
@@ -90,7 +91,7 @@ public class AssetManager {
   /** Used to load assets from storage */
   private static AssetLoader assetLoader = new AssetLoader();
 
-  private static ExecutorService assetLoaderThreadPool = Executors.newFixedThreadPool(1);
+  private static ExecutorService assetLoaderThreadPool = ForkJoinPool.commonPool();
   private static ExecutorService assetWriterThreadPool = Executors.newFixedThreadPool(1);
 
   static {


cwisniew:
I don't think removing the thread pool in getAssetAsynchronously() is the correct way to tackle this. The method is called in the network code and the map rendering code (as well as other GUI code), pausing to load files in these locations can cause problems

fishface60:
> I don't think removing the thread pool in getAssetAsynchronously() is the correct way to tackle this. The method is called in the network code and the map rendering code (as well as other GUI code), pausing to load files in these locations can cause problems

Ah, I see, I missed that getAsset reads the asset from disk since it only uses that to check whether the asset exists, and it's pretty fast to read files on Linux with an NVMe drive so I didn't notice it doing anything.

I'll change this to a previous approach of replacing that thread pool with a [ForkJoinPool.commonPool()](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html#commonPool--)

fishface60:
On second thoughts, that's an observable behaviour change that could have repercussions if suddenly all load callbacks are being executed in parallel, which could cause bugs in a patch release.

Fundamentally the problem _is_ that it needs to run the asset load of the script inside the asset load of the addon, but I might be able to make it work like a recursive lock and if it detects it's already being called from a thread in that pool then load the asset from the thread instead of queueing a new task.

fishface60:
> On second thoughts, that's an observable behaviour change that could have repercussions if suddenly all load callbacks are being executed in parallel, which could cause bugs in a patch release.
> 
> Fundamentally the problem _is_ that it needs to run the asset load of the script inside the asset load of the addon, but I might be able to make it work like a recursive lock and if it detects it's already being called from a thread in that pool then load the asset from the thread instead of queueing a new task.

This approach doesn't work because it's not the loader thread itself that gets blocked calling getAssetAsynchronous.
The AddOnLibrary readFile function uses `CompletableFuture.supplyAsync` which spawns a new thread in the common thread pool.

cwisniew:
@fishface60 , I am slightly confused by your last two comments, is this PR ready for review or not? :)

fishface60:
Those comments were about previous approaches instead of the current one sorry.
But I think you already merged the version into 1.16 and then merged that into develop so I think this PR is unnecessary now.

PR #5227, opened 2025-02-11
Support strformat literals without passing args

### LINKED ISSUE #5226: [Bug]: strformat format literals don't work when not passing parameters
### Describe the Bug

The `strformat` macro function's documented format literals don't work when no arguments are passed.

### To Reproduce

Enter `<pre>{strformat("foo%nbar")}</pre>` in the chat window

### Expected Behaviour

It renders foo and bar on separate lines.

### Screenshots

_No response_

### MapTool Info

_No response_

### Desktop

_No response_

### Additional Context

Bug reported in https://discord.com/channels/296230822262865920/296657960720007169/1338966317662998601

fishface60:

fixes https://github.com/RPTools/maptool/issues/5226

### Description of the Change

`strformat` has an optimisation to skip calling `String.format` if there are no format args. This change removes the optimisation.

### Possible Drawbacks

* Someone might be using `strformat` in an undocumented manner and depends on % formats not being processed when no arguments are passed.
* `strformat` with only `%{variable}` interpolations is now slower.

### Release Notes

* Fixed strformat formatting literals such as %n being ignored if no arguments were passed. strformat can now be used to create string literals with embedded newlines.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5227)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/StringFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/StringFunctions.java
@@ -230,11 +230,7 @@ public class StringFunctions extends AbstractFunction {
       }
       if (functionName.equalsIgnoreCase("strformat")) {
         int size = parameters.size();
-        if (size > 1) {
-          return format(parameters.get(0).toString(), resolver, parameters.subList(1, size));
-        } else {
-          return format(parameters.get(0).toString(), resolver, null);
-        }
+        return format(parameters.get(0).toString(), resolver, parameters.subList(1, size));
       }
       if (functionName.equalsIgnoreCase("matches")) {
         if (parameters.size() < 2) {
@@ -442,10 +438,6 @@ public class StringFunctions extends AbstractFunction {
       }
     }
     m.appendTail(sb);
-
-    if (args == null) {
-      return sb.toString();
-    }
 
     Object[] argArray = args.toArray();
 


src/main/java/net/rptools/maptool/client/functions/StringFunctions.java
```diff
@@ -443,11 +443,7 @@ public String format(String string, VariableResolver resolver, List<Object> args
     }
     m.appendTail(sb);
 
-    if (args == null) {
-      return sb.toString();
-    }
-
-    Object[] argArray = args.toArray();
+    Object[] argArray = args != null ? args.toArray() : new Object[] {};
```
kwvanderlinde: This can be cleaned up by changing the call site of `format()` instead. The casework on `size` there is now redundant as we can always pass along the tail of the parameters. And then the casework here would not be needed either.

kwvanderlinde (approved):

cwisniew:
@fishface60 this is causing issues with existing macros and frameworks, it would be best if we rolled this back and created an alternate function that implemented the "correct" behaviour to maintain compatibility. Do you want to do this or do you want someone else to pick it up?

FYI @kwvanderlinde

kwvanderlinde:
> @fishface60 this is causing issues with existing macros and frameworks, it would be best if we rolled this back and created an alternate function that implemented the "correct" behaviour to maintain compatibility. Do you want to do this or do you want someone else to pick it up?
> 
> FYI @kwvanderlinde

Didn't see this until now. I put up a revert PR just so it's taken care of (#5343). I'll leave implementation of an alternative to @fishface60 or anyone else that feels like picking it up.

fishface60:
Sorry about that, didn't think anyone would be relying on it leaving formats unevaluated.
I'll pick up implementing an alternative next chance I get.

Issue #5164, opened 2025-02-02
[Feature]: Secure socket connections to the server with SSL


fishface60:
### Describe the Problem

When using the Registry without WebRTC, LAN or direct TCP connections to MapTool servers the connection is not encrypted and the authenticity of the MapTool server is not verified.

This feature can be considered related to https://github.com/RPTools/maptool/issues/5063 in that they are both required to ensure 

This can be considered a precondition of features like https://github.com/RPTools/maptool/issues/5163 where it's not the integrity of the game's data model that's being protected, but the user's system, and direct file access is significantly more exploitable so needs to be protected against MITM attacks.

### The Solution you'd like

1. MapTool supports connecting over SSL for servers that sysadmins have very carefully configured SSL tunnels to.

2. MapTool generates self-signed certificates from the existing RSA key.
    The RSA key and self-signed certificate are provided through a [KeyManager][]
    passed to [SSLContext.init][].
    At this point users have to install the server's self-signed certificate into their OS cert store themselves.

3. MapTool provides an interface for clients to copy pre-shared self-signed certificates that were sent by secure back-channels into a MapTool-specific certificate store. That is used by a custom [TrustManager][] passed to [SSLContext.init][].

4. Servers send their certificate to the MapTool registry and clients add the certificate to their certificate store automatically so all that groups need to do to use SSL is for the host to tick a box in the UI.

5. Extend the LAN announcement packets to include SSL certificate so it's just Direct TCP connections that need to have the certificates pre-shared.

[KeyManager]: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/javax/net/ssl/X509KeyManager.html
[SSLContext.init]: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/javax/net/ssl/SSLContext.html#init(javax.net.ssl.KeyManager[],javax.net.ssl.TrustManager[],java.security.SecureRandom)
[TrustManager]: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/javax/net/ssl/X509TrustManager.html

### Alternatives that you've considered.

Concentrate on https://github.com/RPTools/maptool/issues/5063 but that abandons users who don't use WebRTC to less secure connections and inability to offer certain features.

PR #5228, opened 2025-02-11
Minimum viable SSL connection

fishface60:

Works towards https://github.com/RPTools/maptool/issues/5164

### Description of the Change

Most of the change is documenting how to set up a CA from MapTool's RSA keys, generate certificates and create the SSL tunnel.

The code change is adding support for creating an SSLSocket instead of a regular Socket and the URIs and Dialog to opt into using SSL.

### Possible Drawbacks

Having a "Use SSL" checkbox to connect to a MapTool server behind a SSL port without corresponding config to create an SSL server is confusing.

### Release Notes

* Added experimental support for connecting to MapTool servers manually secured behind SSL/TLS. See `doc/SSL.md` for documentation on how to create and configure certificates while the manual steps have yet to be automated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5228)

### DIFF
--- a/clientserver/src/main/java/net/rptools/clientserver/simple/connection/SocketConnection.java
+++ b/clientserver/src/main/java/net/rptools/clientserver/simple/connection/SocketConnection.java
@@ -17,6 +17,8 @@ package net.rptools.clientserver.simple.connection;
 import java.io.*;
 import java.net.Socket;
 import java.net.SocketTimeoutException;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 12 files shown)


cwisniew:
I assume the final intent of this change is to move everything into MT and not require any manual set up, or socat or openssl commands to be run or installed on the users machines?

fishface60:
On Wed, 12 Feb 2025, 01:17 Craig Wisniewski, ***@***.***>
wrote:

> I assume the final intent of this change is to move everything into MT and
> not require any manual set up, or socat or openssl commands to be run or
> installed on the users machines?
>
Yep, the plan is that this CA stuff becomes automated and users are
prompted to trust the CA certificate on first use and verified through some
secure back-channel like the PIN is.

I suppose I could have named the PR better.
I think I called it Minimum Viable because if you really wanted to a GM
could set things up manually and it wouldn't be a huge hassle for players
to use.

The big document is partly "here's an explanation of how it works" but
right now it's more useful for being able to test it at all. After
everything is finished it's probably only there for historical interest and
could be mostly or entirely removed.

>

PR #5231, opened 2025-02-12
Defer initialization of addons until after campaign load

### LINKED ISSUE #5178: [Bug]: Add-on onInit fails if it requires data from e.g. a table
### Describe the Bug

A campaign with an add-on embedded will generate an error on the add-on's `onInit`, if the event requires campaign resources, like switching to a lobby map, reading images from a table etc. 

According to Skullman3194 (on the Discord), it appears that MT attempts to execute the add-on code before it has actually load the campaign resources. 

If I defer the `onInit` code that requires these resources, i.e. use the `execLink(link, 1, "self")` function, my initialization code works.  

### To Reproduce

1. Create a campaign. 
2. Create a second map. 
3. Create an add-on which attempts to switch to second map in `onInit`. 
4. Save campaign. 
5. Restart MapTool and load campaign. 
6. Observe failure. 

### Expected Behaviour

I expect the add-on to be able to load campaign resources in the initialization code, just like a library token. 
Without having to resort to deferring code. 

### Screenshots

_No response_

### MapTool Info

1.15.2 - clean install

### Desktop

Windows 11

### Additional Context

I will ASAP construct an example campaign to demonstrate the issue.

fishface60:

Fixes https://github.com/RPTools/maptool/issues/5178

### Description of the Change

This changes the register/reregister functions to conditionally initialize and adds a method to the Library Manager to initialize all addons, so that when an addon is added to a running campaign it is initialized immediately, but when initializing from connecting to a server or loading from a file addon initialization is deferred until after the campaign is loaded.

### Possible Drawbacks

There may be a cleaner way to handle this than changes all over the place.

### Release Notes

* Deferred running Add-on onInit/onFirstInit scripts until after the campaign is loaded so that their changes aren't trampled during campaign load and they can depend on campaign state.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5231)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppActions.java
+++ b/src/main/java/net/rptools/maptool/client/AppActions.java
@@ -2635,6 +2635,7 @@ public class AppActions {
 
         MapTool.serverCommand().setCampaign(campaign.campaign);
 
+        // NOTE: Addons are initialized in here
         MapTool.setCampaign(campaign.campaign, campaign.currentZoneId);
         ZoneRenderer current = MapTool.getFrame().getCurrentZoneRenderer();
         if (current != null) {
--- a/src/main/java/net/rptools/maptool/client/ClientMessageHandler.java
+++ b/src/main/java/net/rptools/maptool/client/ClientMessageHandler.java
@@ -498,7 +498,7 @@ public class ClientMessageHandler implements MessageHandler {
             Asset asset = AssetManager.getAsset(a);
             try {
               var addOnLibrary = new AddOnLibraryImporter().importFromAsset(asset);
-              new LibraryManager().reregisterAddOnLibrary(addOnLibrary);
+              new LibraryManager().reregisterAddOnLibrary(addOnLibrary, true);
             } catch (IOException e) {
               SwingUtilities.invokeLater(
                   () ->
@@ -658,6 +658,7 @@ public class ClientMessageHandler implements MessageHandler {
     EventQueue.invokeLater(
         () -> {
           Campaign campaign = Campaign.fromDto(msg.getCampaign());
+          // NOTE: Addons are initialized in here
           MapTool.setCampaign(campaign, null);
 
           // Hide the "Connecting" overlay
--- a/src/main/java/net/rptools/maptool/client/MapTool.java
+++ b/src/main/java/net/rptools/maptool/client/MapTool.java
@@ -957,6 +957,7 @@ public class MapTool {
     MapTool.getFrame().getCampaignPanel().reset();
     MapTool.getFrame().getGmPanel().reset();
     UserDefinedMacroFunctions.getInstance().handleCampaignLoadMacroEvent();
+    new LibraryManager().initializeAddOnLibraries();
   }
 
   public static AssetTransferManager getAssetTransferManager() {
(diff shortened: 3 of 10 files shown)


cwisniew:
A better approach is to create a list of "pending" AddOns; if the campaign hasn't been loaded yet, add the registered library to the pending list; otherwise, register it. Once the campaign has been loaded, process the pending add-ons from the list and register them.

This means that we wouldn't end up with partially processed add-ons (like this change would introduce), and any new things added to add-ons would automatically be fixed if they needed to wait for the campaign to be loaded. Also, people modifying the code later would not need to figure out whether they must pass true or false to the register library function. 

Of course, some locking would need to occur for the variable campaign loaded/list (well, I guess you can use a `ConcurrentLinkedQueue` for the list and not lock that). You would also need two new methods in the `LibraryManager`, one to set the Campaign as not loaded (called first thing in the `MapTool.setCampaign()`) and one to set it as loaded (called last thing in `MapTool.setCampaign()`

fishface60:
`CampaignManager.clearCampaignData()`  looks like a good place to flag that add-on loading should be deferred.

Having dug a bit further, I don't think inside `MapTool.setCampaign` itself is the best place to initialize deferred add-ons, since it's sometimes used to set temporary campaigns and is sometimes used to update the map, but we only want to initialize the add-ons when the campaign is finally loaded and not re-initialize if we don't need to.

There's currently code to handle campaign macro init in `MapTool.setCampaign` but that doesn't matter for temporary campaigns because they don't have any and setting the new campaign will load new campaign macro init.
I don't think add-ons get this luxury, and I'm not sure if the existing code is calling onCampaignLoad at times I wasn't expecting.

This is **hard**.

cwisniew:
> `CampaignManager.clearCampaignData()` looks like a good place to flag that add-on loading should be deferred.
> 
> Having dug a bit further, I don't think inside `MapTool.setCampaign` itself is the best place to initialize deferred add-ons, since it's sometimes used to set temporary campaigns and is sometimes used to update the map, but we only want to initialize the add-ons when the campaign is finally loaded and not re-initialize if we don't need to.
> 
> There's currently code to handle campaign macro init in `MapTool.setCampaign` but that doesn't matter for temporary campaigns because they don't have any and setting the new campaign will load new campaign macro init. I don't think add-ons get this luxury, and I'm not sure if the existing code is calling onCampaignLoad at times I wasn't expecting.
> 
> This is **hard**.


What I would suggest is to do this work 
1. Create a new `CampaignLoaded` event in  `net/rptools/maptool/client/events`
2. in `MapTool.setCampaign()` 
	a. create a copy of the reference to the existing campaign `var oldCampaign = client.getCampaign()`
 	b. At the end of the method, compare the campaign to the old one, if they do not match, send out the `CampaignLoaded` event
3. In the `LibraryManager`, subscribe to the `CampaignLoaded` event and use this to empty the queued-up add-ons (from the previous comment). 
4. In `CampaignManager.clearCampaignData()`, call the method that sets the campaign flag to queue add-ons. (I wouldn't do this as an event as its very much sequence dependant where as the final one where the campaign is loaded isnt)


This should mean that it only gets called once per campaign load, it doesn't matter if its a temporary campaign or not it will just "process" the empty queue in that case. Technically it doesn't matter if it gets called more than once per campaign as the queue will be empty on all but the first call (but best to try avoid doing it :) )

kwvanderlinde:
I'm closing this PR as there's hasn't been any activity for a few months. When there's updates to share, feel free to reopen it or open a new PR.

PR #5266, opened 2025-02-17
Overlay Updates 

### LINKED ISSUE #5234: [Feature]: Allow GMs to toggle off locked overlays
### Describe the Problem

When an overlay is locked, players can't hide the overlay using the _Window_ > _Overlays_ menu. This also applies to GMs, which means a GM can accidentally saddle themselves with an overlay they can't easily hide. It also means libraries and other GMs can do it too.

### The Solution you'd like

The menu entries under _Window > Overlay_ would be enabled on GM clients even if the `locked` flag is true.

Writing libraries in a way that is careful to set `locked=0` on GM clients.

### Additional Context

_No response_

### LINKED ISSUE #5235: [Bug]: "Refreshing" an overlay does not update the locked flag
### Describe the Bug

When an overlay is already open, rerunning the same `overlay()` call with a different `locked` value will not change whether the overlay is locked. Instead, the overlay must be closed and reopened to change the `locked` value.

### To Reproduce

1. Add this overlay code to a macro and execute it:
    ```
    [r, overlay("test", "locked=1"): {
      <html>
        <head>
        </head>
        <body>
          Hello there
        </body>
      </html>
    }]
    ```
2. Check that the overlay is locked by run this in chat: `[r: isOverlayLocked("test")]`.
    - Should return `1`.
3. Modify the macro in (1) so `locked=0`, and execute it again.
4. Check that the overlay is locked by run this in chat: `[r: isOverlayLocked("test")]`.
    - Will still return `1`.

### Expected Behaviour

The overlay should be locked if the last `overlay()` call for it set `locked=1`. The overlay should not be locked if the last `overlay()` call for it set `locked=0`.

If the `locked` option is not set in the most recent call, it could either:
1. keep the same state; or
2. apply the default (`locked=0`).

### Screenshots

_No response_

### MapTool Info

1.16.0-rc.1

### Desktop

Linux Mint 22.1

### Additional Context

_No response_

### LINKED ISSUE #5239: [Bug]: Able to close internal stat sheet overlay with closeOverlay()
### Describe the Bug

Although the internal stat sheet overlay cannot be seen in the UI, is not exposed by `getInfo("client")`, and cannot be closed with `closeOverlay("MapTool-Internal:Internal Map Under Pointer Overlay")`, it can be closed with `closeOverlay("*")`.

### To Reproduce

1. Make sure stat sheets are enabled, and have a token that shows its stat sheet when hovered.
2. In chat, prepare the following (but don't send it yet): `[r: closeOverlay("*")]`
3. With keyboard focus still in chat, hover the mouse of the token to show its stat sheet.
4. While the stat sheet is showing and keyboard focus is still in chat, hit Enter to run the command.
    - At this point, the token's stat sheet will disappear.

### Expected Behaviour

The internal stat sheet overlay should not be affected by `closeOverlay()`.

### Screenshots

_No response_

### MapTool Info

1.15.2, 1.16.0.rc-1

### Desktop

Linux Mint 22.1

### Additional Context

_No response_

Jmr3366:
#5234 GMs ignore locked feature
#5235 lock can now be set/reset through `overlay`
#5239 Added better checks for INTERNAL_FRAME_PREFIX

### Possible Drawbacks
Any user initiated frame,dialog,overlay functions where INTERNAL_FRAME_PREFIX is the prefix, will now return _standard reserved word error_
### Release Notes

Examples:
```[overlay("MyOverlay", "zorder=4;"): { <html></html> }]```
Will remove `locked` if it existed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5266)

### CODE REVIEW

File: src/main/java/net/rptools/maptool/client/functions/MacroDialogFunctions.java [hunk 1]
```diff
@@ -75,11 +76,12 @@ public Object childEvaluate(
       Parser parser, VariableResolver resolver, String functionName, List<Object> parameters)
       throws ParserException {
     // Macros can not interact with internal frames/dialogs/overlays
-    if (parameters.size() > 0 && HTMLFrameFactory.isInternalOnly(parameters.get(0).toString())) {
+    if (parameters.size() > 0
+        && (HTMLFrameFactory.isInternalOnly(parameters.get(0).toString())
+            || parameters.get(0).toString().startsWith(AppConstants.INTERNAL_FRAME_PREFIX))) {
```
kwvanderlinde: This `startsWith()` condition is redundant. `HTMLFrameFactory.isInternalOnly()` already does this (more, actually) via its `substring()`/`equalsIgnoreCase()` check.

Although looking at that I'm not sure why `isInternalOnly()` uses a case-insensitive check. Maybe we should just replace the `isInternalOnly()` implementation with a simpler `startsWith()` check?

kwvanderlinde: And in relation to #5239, we would actually need to restrict the meaning of `"*"` specifically in `closeOverlay()` so that it does not close internal overlays.

src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLOverlayPanel.java
```diff
@@ -257,12 +257,18 @@ public void showOverlay(String name, int zOrder, boolean locked, String html, Ob
...
-              overlayManager.setZOrder(zOrder);
-              overlays.add(overlayManager);
+            } else {
+              if (zOrder != overlayManager.getZOrder()) {
+                // Resorts by removing and adding back the overlay
+                overlays.remove(overlayManager);
+                overlayManager.setZOrder(zOrder);
+                overlays.add(overlayManager);
+              }
+              if (locked != overlayManager.getLocked()) {
```
kwvanderlinde: No need for the check first. Might as well just call `setLocked()`

src/main/java/net/rptools/maptool/client/functions/MacroDialogFunctions.java
Jmr3366: > This `startsWith()` condition is redundant. `HTMLFrameFactory.isInternalOnly()` already does this (more, actually) via its `substring()`/`equalsIgnoreCase()` check.
> 
> Although looking at that I'm not sure why `isInternalOnly()` uses a case-insensitive check. Maybe we should just replace the `isInternalOnly()` implementation with a simpler `startsWith()` check?

I added the extra check because in my testing, I was able to create an overlay starting with the reserved word.  I can not duplicate this.  I must have had a typo in my test case.  I think it would be best to revert it back and I will remove what I added here.

kwvanderlinde:
@Jmr3366 Just to confirm, #5239 is out of scope for this PR now, right? If so, I'm happy with how it looks.

Jmr3366:
@kwvanderlinde I missed putting that back in from the initial commit when I was troubleshooting things as it changes how `removeAllOverlays` functions; Let me know and I'll back it out.

src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLOverlayPanel.java
```diff
@@ -217,25 +216,19 @@ private void removeOverlay(HTMLOverlayManager overlay) {
...
-            overlay.flush();
-          }
-          overlays.clear();
-          setVisible(false);
-        });
+    for (HTMLOverlayManager overlay : overlays) {
+      if (!overlay.getName().startsWith(AppConstants.INTERNAL_FRAME_PREFIX)) {
+        removeOverlay(overlay.getName());
+      }
+    }
```
kwvanderlinde: I think this needs to stay within `Platform.runLater()`, or at least the call to `removeOverlay()` needs to be.

kwvanderlinde:
> @kwvanderlinde I missed putting that back in from the initial commit when I was troubleshooting things as it changes how `removeAllOverlays` functions; Let me know and I'll back it out.

No, looks good to keep it in. Just one note on threading, otherwise I'm happy with how it is.

Jmr3366:
I reverted back the `removeAllOverlays`. That function is now only called in one place. I removed it from being used in the `closeOverlay("*")` use case (as I understand it, wouldn't be needed there and for me ,logically speaking, I didn't see reason to treat "*" different to a named overlay).

EDIT: Working through some things I missed . . .  :(

Jmr3366:
@kwvanderlinde I _hope_ this is final . . . :)

kwvanderlinde (approved):
lgtm

PR #5354, opened 2025-04-07
add libgdx based renderer  (#328)

thelsing:
* add gdx renderer

#1781

### Description of the Change

This adds a new optional renderer for maps based on libgdx.

It supports almost everything the swing renderer supports. Currently only the new tokenlabels are missing. Also included is support for animated gifs as token images. Support for video is also included but you can't add videos to maptool currently. 

Labels on the map are zoomed like every other part. This is different from the swing renderer. 

I simplified the rendering pipeline a bit and may have broken something in the gdx renderer. I also didn't add any caches for tokenlocations for now. So there is quite some potentential for optimization. 

The gdx renderer only renders the map. So the swing renderer still has to run to fill locations for token and to call the drawing code for the tool overlays. I added "skipdrawing" code to the swing renderer to avoid drawing when it's not necessary. This is quite ugly but I'm not sure how to proceed. 

For easy development and comparison I added a libgdx-icon to the toolbar that switches between the two renderers. 

The gdx-backend I use is a jogl-backend that was developed by the jogl guys but has been removed from libgdx. I forked it and updated it to latest libgdx at my github account. 

I made quite some stuff in our model classes public because the renderer renders the objects now instead of the model objects drawing itself.

I'll try to work on all requested changes, but I would welcome some help. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5354)


src/main/java/net/rptools/maptool/client/ui/zone/gdx/GdxRenderer.java
```diff
@@ -0,0 +1,2310 @@
...
+    timer.start("tokenlist-5");
+
+    // handle flipping
+    image.setFlip(token.isFlippedX(), token.isFlippedY());
+    timer.stop("tokenlist-5");
+
+    image.setOriginCenter();
+
+    timer.start("tokenlist-5a");
+    if (token.isFlippedIso()) {
```
kwvanderlinde: Needs to be `token.getIsFlippedIso()` now.

kwvanderlinde (changes requested):

kwvanderlinde:
Aside from the one requested change, this PR looks good to go.

src/main/java/net/rptools/maptool/client/ui/zone/gdx/GdxRenderer.java
thelsing: Done.

kwvanderlinde (approved):

bubblobill:
This is causing build failures for me due to a missing dependency `implementation "net.java.jinput:jinput-platform:2.0.7"`

PR #5357, opened 2025-04-09
Separate auras from lights

### LINKED ISSUE #4598: [Feature]: Separate auras from lights
### Describe the Problem

Auras are not lights. They don't illuminate anything. Their lumens value is meaningless. They support visibility control that is not good for lights (`GM`/`OWNER`). They don't combine like lights.

And yet they try to pose as lights. They squat in the _Lights_ tab and token context menu alongside lights. They are accessed by lighting macro functions. They share an internal representation. And they rob the word "light" of any meaning, making it difficult to talk about lights-not-auras without inventing unnatural terminology.

The only things auras and lights have in common are their similar syntax and their respect of VBL.

All this leads to lights and auras being unable to evolve independently. This is not just a matter of MT behaviour, it's also true that any discussion about lights will inevitably be sidetracked by talk of auras and whatever non-lighty stuff they should do.

### The Solution you'd like

A new `Auras` tab is added to the campaign properties dialog where auras can be defined. It will mostly support the existing aura syntax, with a couple tweaks to respect their reality as not lights:
- The `aura` keyword won't be needed, since the whole tab is only auras.
- No lumens values will be supported.

The `Lights` tab syntax will also be updated to respect their reality as not auras, forbidding the `aura`, `GM`, and `OWNER` keywords.

The right-click menu for tokens would have separate menus for lights and auras and the ability to clear each separately from the other.

Keep fighting the good fight in explaining why auras don't act as lights despite being called lights. And keep struggling to talk about lights without accidentally including auras.

### Additional Context

One downside is with macro functions. We already don't distinguish light and auras in functions like `setLight()`. This is set in stone at this point, so that confusion will continue. I'm partial to adding separate functions for each, which would become the preferred ones to use, but this is a minor point compared to the UI changes request above.

kwvanderlinde:

Closes #4598

### Description of the Change

This is a UI change to configure and display auras separately from lights in the few places the two overlap:
1. **Campaign Properties dialog:** auras are now defined in a separate _Auras_ tab.
2. **Edit Token dialog:** unique auras have a separate input field from unique lights.
3. **Token popup menu:** auras have a separate submenu from lights where they can be toggled or cleared. There is no more "Clear All" entry for clearing both lights and auras, but there are stilil separate "Clear Lights" and "Clear Auras" entries.

As a UI-only change, this does not separate auras in the serialized campaign format, nor are existing macro functions affected. Instead, auras and lights are separated on demand when populated the above UI elements, and merged back together when committing new configurations. Because light sources are keyed by a hidden `GUID` rather than name, this means it is now possible to have an aura with the same name as a light. This is fine in terms of UI functionality, but can make the [light functions](https://wiki.rptools.info/index.php/Category:Light_Function) a bit unwieldy since their name arguments can match both a light and an aura. For this reason, it isn't recommended to name auras the same as lights until an improved lighting API is provided.

On the topic of macro functions, since we aren't changing the model, existing macros continue to work the same. The light macro functions can still detect, toggle, and clear auras as well as lights, just as they have always done.

With lights and auras being configurable separately, there is no longer a unified syntax between them. Instead they sport simiilar-but-distinct syntaxes that restrict configurations to their needs:
- Lights reject the `GM` and `OWNER` flags
- Auras ignore lumens values,  and don't accept the `aura` keyword anymore.

The syntax changes required updates to the help text so that it is clear what is accepted now and what is not, and what the differences are between lights and auras. While working on that, I found several peculiarities that I also changed:
- The link to the wiki is now a clickable hyperlink instead of plain text.
- Literal keywords are now written verbatim and formatted as code, rather than being represented to by a translatable strings (the explanation is still translatable). This applies to `scale`, `GM`, `OWNER`, `distance`, `arc`, `offset`, and `width`, as well as `ignores-vbl` which was missing a translation key.
- The shape keywords themselves are no longer translatable (`circle`, `square`, etc), but we do stiil allow a translatable explanatory string.
- Example values and default values in the tables are consistently formatted as code.
- Add some clarity on which parts of sight/light/aura definitions are meant to be repeated.
- Only include actual syntax in code blocks (i.e., no list numbering). This allows the examples to be copied and used as-is. Correspondingly, definition lists are used to describe each light rather than relying on numbering.
- Represent code blocks as `<pre>` tags rather than `<code>` with embedded `<br>`, and reduce the font size to normal so they aren't rivaling headings in size.
- Represent headings as `<h1>`/`<h2>` tags rather than custom enlarged text.
- Allow the help text to scroll horizontally if needed.
- Add a "Name" row to the table to explain that key part of the syntax.
- Add translatable explanations for light examples.

There is actually one small change to the model, but it won't affect saved campaign formats or any other functionality. The `CampaignProperties.lightSourcesMap` is no longer returned directly to any callers, and so cannot be modified directly as has been done for the Campaign Properties dialog. Instead, we now consistently do a deep copy of this map when passing it around. This is done in `readResolve()` (to make sure we know what kind of `Map<>` is used), in the copy constructor, and in `getLightSourcesMap()`. A new `setLightSourcesMap()` method allows callers to write back any changes that are needed.

### Possible Drawbacks

Existing lighting examples are now invalid syntax.

### Documentation Notes

This is what (a portion of) the light help text used to look like:
image

It now looks like:
image

And for auras:
image

In the Edit Token dialog, we now have a separate box for unique auras:
image

The token popup menu has a separate submenu just for auras:
image

### Release Notes

- Separated aura configuration and toggling from lights in the UI (macros are unaffected).

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5357)


bubblobill:
No. You did not just make changes to the Token Properties dialog.
Say it isn't so.

kwvanderlinde:
> No. You did not just make changes to the Token Properties dialog. Say it isn't so.

Notice how I left this in draft :smile: 

I don't want to step on your work.

bubblobill:
> > No. You did not just make changes to the Token Properties dialog. Say it isn't so.
> 
> Notice how I left this in draft 😄
> 
> I don't want to step on your work.

Last time you did it was a nightmare.

Another thought, The context menu is getting seriously oversized. This is it on my notebook monitor 
image
If you are going to be adding a row, we might want to combine the two facing elements into a submenu and Show Path and Revert Last Move into a Path entry

bubblobill:
#5360 first to separate the layout tab, then you are free

kwvanderlinde:
I made #5374 for cleaning up the token menu.

kwvanderlinde:
Ready for review now.

cwisniew (approved):

cwisniew (approved):

cwisniew (approved):

cwisniew (approved):

PR #5360, opened 2025-04-13
Token layout tab reorganised

### LINKED ISSUE #5096: [Feature]: Improved and enhanced token layout options

### LINKED ISSUE #5097: [Feature]: Move token layout panel to its own tab in edit token dialogue

bubblobill:
resolves #5097

### Description of the Change
Cut down version of the changes in #5104

This PR makes no changes to the token class and does not implement independant scaling by axes or rotation of the token image.

### Possible Drawbacks
It is lame

### Documentation Notes
Moved token layout panel to its own tab

### Release Notes
Moved token layout panel to its own tab

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5360)


src/main/java/net/rptools/maptool/model/Token.java
```diff
@@ -232,32 +236,29 @@ public enum Update {
...
+  private int lastY;
+  private Path<? extends AbstractPoint> lastPath;
 
   private int anchorX;
   private int anchorY;
-
   private double sizeScale = 1;
+  private double scaleX = 1;
+  private double scaleY = 1;
+  private double imageRotation = 0;
```
kwvanderlinde: All the `imageRotation` plumbing (field, accessors, macro functions) will need to be removed as well.

bubblobill: Dammit.. thought i had caught them all.

bubblobill: Apparently not even close. 
I think all references to imageRotation have now been removed.

kwvanderlinde (changes requested):
I like how it looks. There's just one bug to chase down yet before merging.

When I set offsets, the selection box for the token is offset by the expected amount, but the token image is offset by twice that amount. E.g., if I place a token here on a 100x100 grid:
image

Then set the offsets to -50, -50, this is what happens
image

And this is what should happen, and which is what happens on 1.16.2:
image

bubblobill:
> Then set the offsets to -50, -50, this is what happens 
image

Which is mostly right because this ->

> And this is what should happen, and which is what happens on 1.16.2:
 image

is wrong. What you should be seeing is this

image

The token layout should primarily be used for correcting the image placement within the cell, i.e. its apparent position. example, a little zoom and some offset makes this guy fill the space and leaves his arm out in the cold.:
image

Changing the actual position of the token is not related to the layout.

Changing the actual position of the token relative to its footprint should be handled in footprints. Part of a whole bundle of follow-on improvements to go with implementing the new footprint stuff.

Check the new commit which leaves the selection marquee where it belongs.

kwvanderlinde (approved):

Issue #4600, opened 2023-12-31
[Refactoring]:  Code maintenance - UI text strings


bubblobill:
### Found a problem with translation, incorrect text, or some other dodgy text?
Please add a comment to this issue.

cwisniew:
Maybe just add the tasks as tasks in the "Improvements you would like to see" portion (using `- [ ] Task description`) when you want to address something to make it easier to track when it's done. It's also easy to click on these and turn them into issues if further discussion is required, I have no problems with these sub issues not including all of the template info.

Baaaaaz:
`Token Type` is not the same as `Token Property Type`, however the Edit > Campaign Properties > Token Properties UI tab content seems to have a heading of `Token Type` for things that relate to `Token Property Type`.  Action requested is to resolve the misnaming to `Token Property Type` (and perhaps its position by moving it to the left to align with the list box of Token Property Types).

![image](https://github.com/RPTools/maptool/assets/131240480/1168de97-1e22-4fec-9a90-b29604d8fa9e)

bubblobill:
![image](https://github.com/RPTools/maptool/assets/45483160/987ab746-be42-4ca9-b18f-251fbb7ccccd)

Maklgu:
![Image](https://github.com/user-attachments/assets/14ca657e-8505-4985-8399-b4bf51cc5c16)
![Image](https://github.com/user-attachments/assets/82d2e64d-9c31-4f7c-9a80-884551833e52)
These text strings should be replaced by a `Could not load footprint definitions for grid type {0}.`

Maklgu:
![Image](https://github.com/user-attachments/assets/543504d6-531c-48f8-b051-d73208fba651)
This should be replaced by a `<html>Displays names below tokens.<br>Visible NPC: {0} on {1}<br>Visible PC: {2} on {3}<br>Not Visible: {4} on {5}.</html>` or `<html>Displays names below tokens. By default:<br>Visible NPC: White on Blue<br>Visible PC: Black on Grey<br>Not Visible: White on Dark Grey.</html>`
Reason: color settings for token names overlay can be changed through the Accessibility tab of MapTool preferences.

Maklgu:
![Image](https://github.com/user-attachments/assets/e096cec6-032c-4b5f-bc83-0ad640d293d6)

![Image](https://github.com/user-attachments/assets/e3b3eb61-bca2-4032-be3e-74759a668971)

[...]

PR #5366, opened 2025-04-20
Consolidated getFootprints to aid translation.

### LINKED ISSUE #5365: Consolidate grid footprint errors into one string

bubblobill:
fixes #5365 
specifically addressing [this comment](https://github.com/RPTools/maptool/issues/4600#issuecomment-2765637370)

### Description of the Change
Brought similar error messages together under one string with the grid type as an argument.
As the message is used with showError, I added overloads for showError and generateMessage to accept additional parameters.
Since I was consolidating code, I moved all getFootprints from the various grid classes and put them under Grid.

### Possible Drawbacks
I may break things unwittingly.

### Documentation Notes
N/A

### Release Notes
N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5366)


src/main/java/net/rptools/maptool/model/Grid.java
```diff
@@ -176,6 +184,11 @@ private int normalizeFacing(int facing) {
    */
   public abstract Point2D.Double getCellCenter(CellPoint cell);
 
+  protected OffsetTranslator getOffsetTranslator() {
+    return null;
+  }
+  ;
```
kwvanderlinde: Stray semicolon

src/main/java/net/rptools/maptool/model/Grid.java
```diff
@@ -211,7 +224,24 @@ public TokenFootprint getFootprint(GUID guid) {
...
 
-  public abstract List<TokenFootprint> getFootprints();
+  public List<TokenFootprint> getFootprints() {
+    if (footprintList == null) {
+      String type = GridFactory.getGridType(this);
+      try {
+        String path = FOOTPRINT_XML_PATHS.get(type);
+        if (type.equalsIgnoreCase(GridFactory.HEX_HORI)
+            || type.equalsIgnoreCase(GridFactory.HEX_VERT)
+            || type.equalsIgnoreCase(GridFactory.ISOMETRIC_HEX)) {
```
kwvanderlinde: Would be cleaner & more extensible to call `getOffsetTranslator()` and check if the result is `null` rather than doing type checks.

src/main/java/net/rptools/maptool/client/MapTool.java
```diff
@@ -207,13 +207,33 @@ public static Dimension getThumbnailSize() {
...
+  /**
+   * Same as {@Link #generateMessage(String, Throwable)} but allows for additional arguments for
+   * formatting the string.
+   *
+   * @param msgKey the string to use when calling {@link I18N#getText(String)}
+   * @param t the exception to be processed
+   * @param params arguments to pass when fetching string.
+   * @return the <code>String</code> result
+   */
+  private static String generateMessage(String msgKey, Throwable t, Object... params) {
```
kwvanderlinde: Use `Object[] params` instead of `Object... params`. Same thing under the hood, but we aren't calling it in a way that needs varags. Also makes the `null` check below look a lot less wierd.

kwvanderlinde (changes requested):

src/main/java/net/rptools/maptool/model/Grid.java
bubblobill: fixed

src/main/java/net/rptools/maptool/model/Grid.java
bubblobill: done

src/main/java/net/rptools/maptool/client/MapTool.java
bubblobill: implemented

kwvanderlinde (approved):

PR #5375, opened 2025-04-22
Allow MT Line Parser and JS Fetch to resolve lib: uri for macros

### LINKED ISSUE #5336: [Bug]: Inconsistent behaviour between differing link syntax
### Describe the Bug

Different syntax for links to macros can produce different results.
e.g. there are situation where `lib://Ackbar/macro/dataProvider` and `"macro:dataProvider@lib:Ackbar` do not function identically.
Similarly, relative links that might be expected to work produce non-sensical errors.

### To Reproduce

Open a campaign and add the token attached below (lib:Ackbar) to it.
In the selected window, Click on the `html.dialog5` button.
Open the macro `/js/my.js` for editing.
At line 52, change the line `myRest.dataURI = dpURI2;` 
the values dpURI,  dpURI3, and  dpURI4 change the URI being used for *fetch* between the following values:
```
const dpURI = "lib://Ackbar/macro/dataProvider";
const dpURI2 = "macro:dataProvider@lib:Ackbar";
const dpURI3 = "dataProvider";
const dpURI4 = "./dataProvider";
```
Note the difference in the chat window between outputs using dpURI2 and the others.
None of the resulting error message components are of diagnostic use.
```
Getting data from macro "dataProvider"
135: RangeError: Status must be between 200 and 599
Response@[native code]
@
send@[native code]
send@
@
@[native code]
```

### Expected Behaviour

I would expect
1. `lib://Ackbar/macro/dataProvider` and `macro:dataProvider@lib:Ackbar` to be synonymous.
2. Error messages that advise me the link is invalid or provide an error code I can research elsewhere.

Ideally, relative links would behave accordingly in suitable environments, but that is not as bad as two syntaxes for the same thing being a breaking point.

### Screenshots

_No response_

### MapTool Info

1.15.2

### Desktop

Win 10

### Additional Context

[lib_Ackbar.rptok.zip](https://github.com/user-attachments/files/19526235/lib_Ackbar.rptok.zip)

### LINKED ISSUE #5347: [Bug]: The createAsset macro function can not create assets from add on URI
### Describe the Bug

The `createAsset` macro function can not accept a lib:// URI 

### To Reproduce

Try to create an asset based on a lib:// URI

### Expected Behaviour

You should be able to create an asset from a lib URI

### Screenshots

_No response_

### MapTool Info

MapTool 1.16.x

### Desktop

All

### Additional Context

_No response_

cwisniew:
Fixes #5336 
Fixes #5347 

### Description of the Change
The ability to use the lib:// URI format in fetch() in JavaScript, while documented on the Wiki, didn't actually exist. This change enables its use (as well as relative URI). As a consequence, it will also be possible to use the lib:// URI in other places that previously only allowed macro@lib:Token, so it should be possible to use it everywhere now. This required a fair amount of refactoring and code cleanup to work correctly. 

This change also introduces new macro functions to take full advantage of this 

`return = call(macro, [arg1 ... [argN]])`
Calls a macro with the specified arguments

`call.deferred(macro, [arg1 ... [argN]])`
Deferred call of a macro with the specified arguments (always returns "")

`call.a(macro, JSONArray)`
Calls a macro with the specified arguments in the JSON Array, `arg(0) = jsonArray[0]`, `arg(1) = jsonArray[1]` etc

`call.deferred.a(macro, JSONArray`
Deferred version of the above

`jsonArrayReturn = call.foreach(macro,  [arg1 ... [argN]])`
Calls the macro once for each argument, passing that argument to the macro

`jsonArrayReturn = call.foreach.a(macro, JSONArray)`
Calls the macro once for each element in the JSONArray, passing that element to the macro

Unlike `[macro(): ]` and UDF calls, these functions can bypass the requirement of the parser to convert some things into a string before parsing the call where it is converted back again. So these should be more performant than using those methods, especially for large JSON objects or Strings. Likewise, the `call.foreach` / `call.*.a` versions avoid redundant parsing of the arguments on every loop iteration, so they should be significantly faster than `[foreach():]`. 

This change also includes 
* a check that the document is loaded before running any JS from runJSFunction (note this is not foolproof as the document may not have started loading yet if runJSFunction is called directly after opening a dialogue/frame/overlay), but it still should be included as a sanity check). 
* The ability to create assets from lib:// URI (more tidying up of lose ends where lib:// URI was not taken into account)

The JavaScript files involved have also been formatted as the indentation was all over the place (putting it mildly)

### Possible Drawbacks
This will require extensive testing to ensure that there are no subtle differences that could break users' macros.
Its not possible to apply the same tweaks for `[macro():]` and UDFs of bypassing the extra parsing steps due to the way they function.

### Documentation Notes
JavaScript fetch() from html5 dialog/frame/overlay now works with lib:// URI syntax (including relative paths)
Added more efficient ways to call macro functions
* `return = call(macro, [arg1 ... [argN]])`
* `dcall(macro, [arg1 ... [argN]])`
* `call.a(macro, JSONArray)`
* `dcall.a(macro, JSONArray`
* `jsonArrayReturn = call.foreach(macro,  [arg1 ... [argN]])`
* `jsonArrayReturn = call.foreach.a(macro, JSONArray)`

### Release Notes
Fixed inconsistencies when using lib:// to refer to macros, added new functions which are more efficient to call macros

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5375)


cwisniew:
Ready for review

src/main/java/net/rptools/maptool/client/functions/MacroLinkFunction.java
kwvanderlinde, line 1: In `createMacroText()`, will also need to update `MapTool.getParser().getMacroSource();` to `MapTool.getParser().getMacroSource().getLocation();`

src/main/java/net/rptools/maptool/client/MapToolMacroContext.java
```diff
@@ -51,7 +52,8 @@ public MapToolMacroContext(String name, String source, boolean trusted) {
    * @param trusted Is the macro trusted or not.
    * @param macroButtonIndex The index of the button that ran this command.
    */
-  public MapToolMacroContext(String name, String source, boolean trusted, int macroButtonIndex) {
+  public MapToolMacroContext(
+      String name, MacroLocation source, boolean trusted, int macroButtonIndex) {
```
kwvanderlinde: Can `source` ever be `null` here, or can we also mark it `@Nonnull` as in the other constructor?

src/main/java/net/rptools/maptool/model/library/LibraryManager.java
```diff
@@ -260,10 +261,14 @@ public Optional<Library> getLibraryForMTScriptCall(
       String namespace, MapToolMacroContext context) {
     String ns = namespace;
     if ("@this".equalsIgnoreCase(namespace)) {
-      if (context == null || context.getSource() == null || context.getSource().isEmpty()) {
+      var source = context.getSource();
+      if (context == null
```
kwvanderlinde: Should check `context  == null` a line earlier, otherwise we could see NPEs.

src/main/java/net/rptools/maptool/model/library/token/LibraryToken.java
```diff
@@ -305,8 +305,16 @@ public CompletableFuture<Optional<MTScriptMacroInfo>> getMTScriptMacroInfo(Strin
...
         .runOnSwingThread(
             () -> {
+              // Cater for the different ways the macro name can be specified
+              String expectedMacroName = macroName;
+              if (macroName.toLowerCase().startsWith("lib://")) {
+                expectedMacroName = macroName.replaceFirst("(?i)lib:\\/\\/.*\\/macro\\/", "");
+              } else if (macroName.toLowerCase().startsWith("lib:")) {
+                expectedMacroName = macroName.substring(4);
+              }
+              ;
```
kwvanderlinde: Stray semicolon

src/main/java/net/rptools/maptool/client/functions/CallFunction.java
```diff
@@ -0,0 +1,171 @@
...
+  private final JsonArrayFunctions jsonArrayFunctions = jsonMacroFunctions.getJsonArrayFunctions();
+
+  /** Create the Object and register the paramater count and function name. */
+  public CallFunction() {
+    super(
+        1,
+        UNLIMITED_PARAMETERS,
+        "call",
+        "call.a",
+        "dcall",
```
kwvanderlinde: `call.deferred` would be more consistent and discoverable compared to `dcall`.

kwvanderlinde (changes requested):

src/main/java/net/rptools/maptool/client/functions/MacroLinkFunction.java
cwisniew: Updated this

src/main/java/net/rptools/maptool/client/MapToolMacroContext.java
cwisniew: The constructors now match

src/main/java/net/rptools/maptool/model/library/token/LibraryToken.java
cwisniew: I called the semicolon catcher, they have rounded it up and removed it

src/main/java/net/rptools/maptool/client/functions/CallFunction.java
cwisniew: Changed, I am forwarding all hate mail I get for making the function name too long to you ;)

kwvanderlinde (approved):

PR #5390, opened 2025-05-01
Replace grgit with git-version in build script

### LINKED ISSUE #5389: [Feature]: Replace grgit in the build as its no longer maintained
### Describe the Problem

grgit is no longer maintianed and doesn't support several git features (like worktrees) which will cause the build to fail if they are being used

### The Solution you'd like

Replace grgit with git-version plugin in the gradle build scripts

_No response_

### Additional Context

_No response_

cwisniew:
Resolves #5389

### Description of the Change
Replaces grgit in the gradle build with the git-version plugin

### Possible Drawbacks
Should be none

### Documentation Notes
Replaces grgit in the gradle build with the git-version plugin, enabling some more advanced features of git to be usable

### Release Notes
Replaces grgit in the gradle build with the git-version plugin, enabling some more advanced features of git to be usable

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5390)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ plugins {
     id "base"
     id "java"
 
-    alias(libs.plugins.grgit)
+    alias(libs.plugins.git.version)
     alias(libs.plugins.javafx)
     alias(libs.plugins.jpackage.runtime)
     alias(libs.plugins.protobuf)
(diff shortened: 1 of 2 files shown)


kwvanderlinde (approved):

bubblobill:
Or, as a possible downside, you could break the IDEs of contributors.

cwisniew:
> Or, as a possible downside, you could break the IDEs of contributors.

How so? I tested it and it works fine with Intellij. Does it work ok when you run gradlew from the command line?

PR #5406, opened 2025-05-06
Generate walls from UVTT line-of-sight data

### LINKED ISSUE #5182: [Feature]: Add option to generate walls instead of masks when importing UVTT
### Describe the Problem

The line-of-sight portions of a UVTT are more naturally represented using walls instead of 2-dimensional masks. In complicated situations, e.g., auto-generated caves, there is also a noticable delay when importing the UVTT due to the costly unions that must be done, and this would be avoided by generating walls instead.

### The Solution you'd like

Add a preference to allow walls to be generated instead of masks when importing UVTT. There would be three options to choose from:
1. Always generate walls
2. Always generate BL
3. Ask each time

If (3) is selected, a dialog will be shown the user before the file is processed so that the user can control the behaviour on a map-by-map basis.

When walls are requested, they would be generated as follows:
- `"line_of_sight"` data would result in walls that block vision and movement in both directions.
- `"objects_line_of_sight"` data would result in rings of directional walls that block movement and block vision leaving the object.
- `"portals"` would be treated as windows. If the `"closed"` flag for a portal is `true`, the resulting wall would block both movement and vision.  Otherwise, the wall will block movement but not vision.

We could force the use of walls instead of giving a preference. But given how new walls are, and the fact that some might want to be able to reproduce previous imports, I think leaving in a way to use masks is a good idea, at least for now.

### Additional Context

_No response_

kwvanderlinde:

Resolves #5182

### Description of the Change

When importing a UVTT file, the user now has a choice of whether to generate VBL+MBL as always, or to use the newer walls. By default, the user is prompted for a choice each time, but this choice can optionally be remembered for future imports. Alternatively, the choice can be set in the Preferences dialog.

When generating walls, open portals are represented as walls, albeit walls the don't block movement, sights, lights, or auras. This differs from the VBL+MBL import where open portals result in nothing, but is consistent in terms of not blocking anything. Since a wall is created, the user can easily modify the blocking properties of any portals to their liking after the import is completed, in case this default behaviour does not suffice.

### Possible Drawbacks

Users may be confused at first with the extra prompt.

### Documentation Notes

Either walls or *BL can be used to represent the line of sight data from a Universal VTT file. When importing a UVTT file, you will be asked which one you want to use:
image

If you don't want to be asked every time, select the "Remember my choice" option:
image

Alternatively, you can set your preferred option via Preferences:
image

### Release Notes

- Added the option for walls to be generated during UVTT import instead of VBL+MBL

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5406)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppPreferences.java
+++ b/src/main/java/net/rptools/maptool/client/AppPreferences.java
@@ -193,6 +193,9 @@ public class AppPreferences {
   public static final Preference<MapSortType> mapSortType =
       EnumType.create(MapSortType.class, "sortByGMName", MapSortType.GMNAME);
 
+  public static final Preference<UvttLosImportType> uvttLosImportType =
+      EnumType.create(UvttLosImportType.class, "uvttLosImportType", UvttLosImportType.Prompt);
+
   public static final Preference<Boolean> useSoftFogEdges = BooleanType.create("useSoftFog", true);
 
   public static final Preference<Boolean> newMapsHaveFow =
@@ -506,6 +509,23 @@ public class AppPreferences {
 
     MapSortType() {
       displayName = I18N.getString("mapSortType." + name());
+    }
+
+    @Override
+    public String toString() {
+      return displayName;
+    }
+  }
+
+  public enum UvttLosImportType {
+    Walls("uvttLosImportType.walls"),
+    Masks("uvttLosImportType.masks"),
+    Prompt("uvttLosImportType.prompt");
+
+    private final String displayName;
+
+    UvttLosImportType(String key) {
+      displayName = I18N.getString(key);
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


FullBleed:
Do you foresee there being more uvtt type settings in the future?  If so, it might be good to give uvtt settings their own tab so they can be safely ignored--or easily found--as needed.  Other instructions/expectations for how MT handles uvtt files could also be included in such a tab.

cwisniew:
> Do you foresee there being more uvtt type settings in the future? If so, it might be good to give uvtt settings their own tab so they can be safely ignored--or easily found--as needed. Other instructions/expectations for how MT handles uvtt files could also be included in such a tab.

UVTT is limited to what it can contain, I can't see many more options other than create lights in the near future, it would be more likely that in the future there would be other import formats -- not that this is likely either -- and it could be grouped with them in another tab if it ever happens.

cwisniew (approved):

Baaaaaz:
@kwvanderlinde Typo in the dialog title?  Should read "Universal" not "Universla".
https://github.com/search?q=repo%3ARPTools%2Fmaptool%20universla&type=code

PR #5411, opened 2025-05-06
Add bounds to numeric preferences

### LINKED ISSUE #5393: [Feature]: Preferences - Include a valid range in definition
### Describe the Problem

Preferences are often limited in the range of values they can represent but not in the range they can hold.
e.g. an int may represent a percentage or 0-255; a double a value between 0 and 1.
e.g. grid size can be any int but is actually bound by constants defined elsewhere in the code.

UI controls for numbers would be easier to implement if the bounds were accessible in the preference itself.

### The Solution you'd like

For numerical types, include a min and max bound and a second constructor, e.g.
```
private Preference(String key, T defaultValue, Type<T> type) {
  this(key, () -> defaultValue, type);
}

private Preference(String key, Supplier<T> defaultValue, Type<T> type) {
  this(key, defaultValue, null, null, type);
}

private Preference(String key, T defaultValue, T minValue, T maxValue, Type<T> type) {
  this(key, () -> defaultValue, () -> minValue, () -> maxValue, type);
}

private Preference(
    String key, Supplier<T> defaultValue,  @Nullable Supplier<T> minValue, @Nullable Supplier<T> maxValue, Type<T> type) {
  this.key = key;
  this.defaultValue = defaultValue;
  this.minValue = minValue;
  this.maxValue = maxValue;
  this.type = type;
}
```

_No response_

### Additional Context

_No response_

kwvanderlinde:

Resolves #5393

### Description of the Change

Preferences of type `Integer` or `Double` are now considered "numeric preferences" and are represented as `NumericPreference`, a child class of `Preference`. Numeric preferences support a minimum and maximum value that defines their valid range. When storing a value for a numeric preference, the value is first clamped to fit within the valid range. When loading a value, the loaded value is also clamped to fit within the value range.

The bounds of a numeric preference are accessible via `NumericPreference#getMinValue()` and `NumericPreference#getMaxValue()`, so that they can be easily consumed in UI code per the original motivation.

A few bits of code could be removed:
- `ByteType` became redundant since it is the same as `IntegerType` with values constrained to the range [0, 255].
- Validators aren't needed anymore. They were only used to enforce a positive `frameRateCap`, but that is more clearly represented using explicit bounds instead of a generic `Predicate`.
- `FloatType` was used only for one preference (`statusBarSpeed`), and that preference's values were ultimately only used in a `double` context. So that preference now uses `DoubleType` and `FloatType` is no more.

### Possible Drawbacks

None

### Documentation Notes

N/A

### Release Notes

- Added support for explicit bounds on numeric preferences

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5411)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppPreferences.java
+++ b/src/main/java/net/rptools/maptool/client/AppPreferences.java
@@ -22,7 +22,6 @@ import java.io.File;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Consumer;
-import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.prefs.Preferences;
 import javax.annotation.Nullable;
@@ -60,10 +59,10 @@ public class AppPreferences {
   public static final Preference<Boolean> saveReminder =
       BooleanType.create("autoSaveReminder", true);
 
-  public static final Preference<Integer> autoSaveIncrement =
+  public static final NumericPreference<Integer> autoSaveIncrement =
       IntegerType.create("autoSaveIncrement", 5);
 
-  public static final Preference<Integer> chatAutoSaveTimeInMinutes =
+  public static final NumericPreference<Integer> chatAutoSaveTimeInMinutes =
       IntegerType.create("chatAutosaveTime", 0);
 
   public static final Preference<String> chatFilenameFormat =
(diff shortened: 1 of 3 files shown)


bubblobill:
1. Yay!
2. Damn you and your distaste for floats : ) - Gone and broke my uncomitted work.

kwvanderlinde:
> 2. Damn you and your distaste for floats : ) - Gone and broke my uncomitted work.

How badly?

bubblobill:
Well, mostly Graphics2D stuff like BasicStroke whinges when I don't feed it a float, and a Double won't just cast to a float so I have to Number it and get the float value. 

Don't be annoyed when the a getAsFloat method magically appears in the prefs class. 

So far it just means changing a dozen or so prefs I have set up in various projects.

kwvanderlinde:
> Well, mostly Graphics2D stuff like BasicStroke whinges when I don't feed it a float, and a Double won't just cast to a float so I have to Number it and get the float value.

Right, `Double` doesn't cast to `float` but it does have the `floatValue()` method. So no need to explicitly bring `Number` into it. E.g.,:
```
float value = AppPreferences.scrollStatusSpeed.get().floatValue();
```

> Don't be annoyed when the a getAsFloat method magically appears in the prefs class.

See above. Wouldn't gain anything compared to just using `floatValue()`.

cwisniew (approved):

Issue #4966, opened 2024-10-01
[Feature]: Foreshortened token rendering on isometric grids


kwvanderlinde:
### Describe the Problem

Non-figure tokens look out of place on isometric grids because they don't have any perspective applied.

### The Solution you'd like

Top-down, circle and square tokens would be rotated and scaled on isometric grids to enhance the illusion of depth.

### Additional Context

It should look something like this:
![foreshortened-token](https://github.com/user-attachments/assets/1b8b234e-2fb3-45f4-85c9-0cdec853a389)

FullBleed:
That looks better.

It would be even better if it could be transformed into looking like a coin standing on edge.

bubblobill:
oh snap.
see my comment on your [facing issue](https://github.com/RPTools/maptool/issues/4965#issuecomment-2387708677)

It is a straightforward series of affine transforms -> shear, scale, and rotate

bubblobill:
> It would be even better if it could be transformed into looking like a coin standing on edge.

Next you'll be wanting them on little plastic stands

![image](https://github.com/user-attachments/assets/c6839dc8-5bea-44ba-bfeb-aa48151aa494)

FullBleed:
Well, I don't think it really *needs* a stand... but they definitely need to spin around like a coin when the tokens get hit.

In all seriousness... I do like the standing render better for circle and square tokens.

Topdowns, however, would have to layflat... Can a "raised" effect be added for them?  Something to give it thickness?  Something about the transform to align with iso makes it uncannily melt into the map.

kwvanderlinde:
Seeing that standing coin is actually really cool. It's like a figure token without all the work :sweat_smile: 

But there is also utility in having tokens lie flat. E.g., using square tiles to build a map on an iso grid. So - thinking out loud here - what if we made stamps lie flat, and gave a preference to have Tokens stand on edge (on by default)?

Agree with @FullBleed that standing doesn't make sense for top-down tokens. Not sure we can really do much to improve the situation for them vs lying flat - it may just be a matter of top-down tokens not being a good fit for iso grids. Similar to how figure tokens on non-iso grids are questionable.

kwvanderlinde:
In a recurring theme, I found out this functionality technically exists already. Right-click a token > _Flip_ > _Isometric Plane_. For now let's leave aside the fact that the terminology is off - no actual flipping occurs.

***

[...]

PR #5431, opened 2025-05-13
Rework Edit Token dialog to suit smaller screens and promote better organization

### LINKED ISSUE #2641: Edit Token Frame is too big for 1024x768 screens
**Describe the bug**
With the MT header and Windows Toolbar there isn't enough room to click the Okay or Cancel button on a laptop with 1024x768 resolution. Afaik, the Frame is 735px tall with no way to tab to the buttons.

**To Reproduce**
Set your screen to a 768 tall resolution and Edit a Token. Put in some notes and try to hit Okay.

**Expected behavior**
I expect the textarea collapse more or perhaps have alternate okay/cancel button at top of frame.

**MapTool Info**
- Version: 1.8.4

### LINKED ISSUE #5376: [Bug]: Token Editor exceeds a max screen height
### Describe the Bug

The height of Token Editor window exceeds the one for screen because of newly added options to the 'Config' tab. Users with screen height less than 900 px may encounter this problem.
Image
The content of the other tabs is resized to fit the new height, so all tabs now should be scrolled manually with a scroll bar (mouse scroll doesn't work here) to see buttons at the bottom. User cannot see the tab switcher and these buttons at the same time.

Here are some examples of tabs with a non-completely visible content.
|Notes/GM Notes tabs|Properties tab|
|---|---|
|Image|Image|
|Image|Image|

### To Reproduce

1. Open Token Editor dialog.
2. Decrease the height of the opened window.

### Expected Behaviour

1. Mouse scroll scrolls the content of the Token Editor dialog pages.
2. The tab switcher bar doesn't scroll with the content of pages and is fixed at the top.
3. Scrolling Notes and GM Notes tabs scroll their text content, not the page itself, and the buttons at the bottom are fixed at the bottom of the dialog, so their position is independent of the content of the 'Config' page. 

### Screenshots

_No response_

### MapTool Info

MapTool Nightly 20250422

### Desktop

_No response_

### Additional Context

_No response_

kwvanderlinde:

Fixes #2641
Fixes #5376

### Description of the Change

The **Edit Token** dialog has had a total makeover to bring related elements together and reduce wasted space.  Both vertical and horizontal space was saved so that small screens could be better supported by this dialog. Functionaliy has been kept the same, just some stuff needs to be found somewhere else now.

The new dialog was tested on small resolutions to make sure that it fits. These are the requirements:
- If using **Aah (Very Large Print)**, a resolution of 1024x768 is needed to avoid scrollbars in the dialog.
- If using **Aah (Large Print)**, a resolution of 800x600 can be used without scrollbars, though the dialog takes up the entire screen.
- If using other themes, the dialog fits comfortably on the screen.

Here is an overview of the changes. Screenshots demonstrating the changes can be found in the **Documentation Notes**.
- Moved tab bar to sit on the left instead of at the top
- Moved top matter (name, label, etc.) intead a new _Basic_ tab, and moved the ownership and speech lists there as well rather than having them sit in their own tabs.
- Moved the token image, portrait, handout into a new _Visuals_ tab, and also moved the token opacity, image table, and flip options there from the _Config_ tab.
- Merge the _GM Notes_ tab into the _Notes_ tab, which itself now has separate nested tabs for _Player Notes_ vs _GM Notes_.
- Moved the stat sheet configs and property type selection out of the _Config_ tab and into the _Properties_ tab.
- Moved the sight, unique light sources, and unique auras options out of the _Config_ tab and into the new _Sights, Lights & Auras_ tab.
- Move the **Is VIsible Over FoW** option out of the _Topology_ tab and into the __Config_ tab with the other visibility options. Then optimize the layout of the _Topology tab.
- Rename the _States_ tab to _States & Bars_ and make the bars panel a sibling to the states panel instead of a child.
- Merge the _Lib:Token properties_ tab into the _Config_ tab.
- Reflow the _Config_ tab to use less horizontal space.
- In the _Layout_ tab, stack the zoom slider above the scale slider to reduce horizontal space requirements.
- The _Hero Lab_ tab was left as-is.

### Possible Drawbacks

It will certainly be an adjustment getting used to the new layout.

### Documentation Notes

These screenshots were taken at their default size using the _Cyan Light_ theme. The system is Linux Mint Cinnamon with a screen resolution of 2560x1440.

_Basic_ tab
image

_Visuals_ tab
image

_Notes_ tab
image

_Properties_ tab
image

_Sights, Lights & Auras_ tab
image

_Topology_ tab
image

_States & Bars_ tab
image

_Config_ tab
image

_Layout_ tab
image

### Release Notes

- Reworked the Edit Token dialog to be better organized and to fit smaller screens.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5431)


cwisniew:
@bubblobill since you turn into a grumpy old man whenever this dialog is mentioned, do you want to review it :)

bubblobill:
I'm good with that since I fully expect to butcher it in the future.

cwisniew (approved):

FullBleed:
Since "Lib:Token Properties" is grayed out when a token is not a lib token, can the "Image Token" box be grayed out when a token isn't a "Figure" token?  That's always been a misleading feature as it has no use for all non-iso tokens/maps (which is the vast majority.)

bubblobill:
@FullBleed But if a token has an image table and the grid changes to isometric, should we not automatically change the token to a figure?

FullBleed:
> @FullBleed But if a token has an image table and the grid changes to isometric, should we not automatically change the token to a figure?

Despite liking isometric maps, I don't use them so I'm sure someone who does will have a stronger opinion.  But it sounds like that presumes that only figure tokens could then be on isometric maps?  I think I'd rather MT be smart about how it renders different types of tokens on Iso maps and not force only one type.  i.e. If I drop a circle token on an iso map that it would treat it like laying a coin on a table and render it accordingly... I actually thought that feature was done recently... *shrug*

At any rate, I'm not sure how that relates to what I was commenting on though...  As far as I know, the "Image Table" is a feature related *exclusively* to Figure Tokens.  So if a token isn't a Figure token, then I think the "Image Table" box should be disabled/grayed out so people don't think it does something.  As is, it seems to indicate that a circle (or token other than figure) can have an associated image table that you could use to changes images... but that is not the case.

kwvanderlinde:
> Since "Lib:Token Properties" is grayed out when a token is not a lib token, can the "Image Token" box be grayed out when a token isn't a "Figure" token? That's always been a misleading feature as it has no use for all non-iso tokens/maps (which is the vast majority.)

Not here, because this PR is explicitly not meant to change functionality. But also not in principle, as I expand upon below.

> Despite liking isometric maps, I don't use them so I'm sure someone who does will have a stronger opinion. But it sounds like that presumes that only figure tokens could then be on isometric maps? I think I'd rather MT be smart about how it renders different types of tokens on Iso maps and not force only one type. i.e. If I drop a circle token on an iso map that it would treat it like laying a coin on a table and render it accordingly... I actually thought that feature was done recently... _shrug_

Not done, but the FR does exist (#4966), complete with a rant about lack of smarts :slightly_smiling_face:

> At any rate, I'm not sure how that relates to what I was commenting on though... As far as I know, the "Image Table" is a feature related _exclusively_ to Figure Tokens. So if a token isn't a Figure token, then I think the "Image Table" box should be disabled/grayed out so people don't think it does something. As is, it seems to indicate that a circle (or token other than figure) can have an associated image table that you could use to changes images... but that is not the case.

No, it's not exclusive to Figure tokens. IMO it should only be used to render figure tokens, and that's clearly what image tables are meant for. But the reality is that it is possible to use it with any token on any grid type. The wiki even [explains how to do it.](https://wiki.rptools.info/index.php/Introduction_to_Tables#Token_Image_Tables)

Expanding on @bubblobill's point (although I'll avoid references to hypothetical functionality), just because a token is a Figure token right now doesn't mean it won't be in five minutes. If the grid is changed to iso, or the token is moved to an iso map, it is natural that I want to change the token to Figure if it has an image table. So it also makes sense to allow a user to edit the image table properties even if the token isn't currently a Figure token.

FullBleed:
> No, it's not exclusive to Figure tokens. IMO it should only be used to render figure tokens, and that's clearly what image tables are meant for. But the reality is that it is possible to use it with any token on any grid type. The wiki even [explains how to do it.](https://wiki.rptools.info/index.php/Introduction_to_Tables#Token_Image_Tables)

Well, given that, I'd agree it should be visible all the time.  I'd still feel differently if it was a Figure only feature.

At any rate I learned something new about token image tables.  ;)

Issue #5147, opened 2025-01-24
[Feature]: All windows should have their exit and save buttons visible


peterkelly70:
### Describe the Problem

Some of us have bad eyes and need large fonts, but having the exit or save buttons off the bottom of the screen is a real pain. It’s very anti-accessibility to render a window that you can’t exit because your fonts are large. I suggest that all internal frames have scrollbars and that the size of container elements should never be bigger than the available screen real estate. Internal frames should scroll, while exterior ones should always fit the screen.

### The Solution you'd like

I’d like to see all internal frames be scrollable, and all external containers render all their buttons so they can be clicked.

### Alternatives that you've considered.

None

### Additional Context

![Image](https://github.com/user-attachments/assets/c1e8c2fe-c72a-45df-87ad-dd88dc445ebf)

FullBleed:
What OS are you on?

If Windows, what accessibility settings are you using?  Are you using "Scaling" and/or "Accessibility/Text Size"?

I'm in 4k using scaling 200% and the Accessibility/Text Size at 125% and I don't get that kind of compaction and loss of save/close buttons you're getting.  There are some places where some text gets clipped in the headers though, but nothing like what you're showing.

Also, have you tried this to see if it helps: https://www.rptools.net/2019/03/windows-10-dpi-scaling-strikes-again/

bubblobill:
Dear Blind @peterkelly70,
(To the magical and masterful musical stylings of "Ballad of the Hydra Slayers")
Had this problem, yeah, yeah;
It was annoying, yeah, yeah;
So I fixed it, yeah, yeah;
etc.

#5101 #5102 (as part of #5096) addresses *most* of these issues. It's not the desirable fix of having them always visible, but now they can at least be scrolled into view. Making MT accessible requires some major rewrites of many UI elements that have accreted over the years. This puts most dialogs into a scrollpane as an interim fix.
This behaviour can be previewed in the 1.16 beta releases.

If you think this fixes it,
You can close the issue, yeah, yeah.

Edit: Also, Theme Aah(Large Print) & Aah(Very Large Print) may prove better for you. They have a better chance of the UI Manager adjusting to cope.

PR #5432, opened 2025-05-13
Generic dialog fixes

### LINKED ISSUE #4107: [Feature]: Limit dialogue boxes to screen boundaries
### Describe the Problem

Dialogue boxes can exceed the size of the screen, this usually pushes the Okay/Cancel buttons off the screen (see example below).
1. Editing content changes the controls focus so Enter/Escape may not function to close the dialogue.
2. Most users do not know how to move them (when they can be moved) to make the buttons visible. 
3. Acessibility measures like large-print make this more common, making it harder for those who need the most assistance.
### The Solution you'd like

Limit the max-height/width to the screen bounds and Insert a scroll pane to hold the content.

Re-education camps

### Additional Context

image

### LINKED ISSUE #5423: [Bug]: Stray maximize button in some dialogs
### Describe the Bug

Since 1.16, some dialogs have maximize/restore buttons added to them. On Windows, these are in the titlebar as expected, but on Linux they overlap with the dialog content.

This affects the **New Token** dialog, **Edit Token** dialog, **Add Resource to Library...** dialog, and probably a few others.

### To Reproduce

1. Open MapTool on Linux running X11
2. Open Preferences, and enable **Show Dialog on New Token**.
3. Drag a token from the Library to the map. The **New Token** dialog should open.
4. Notice the maximize button overlapping the **Name** field.

### Expected Behaviour

The maximize button should be part of the titlebar, or else not be there.

### Screenshots

Here is the **New Token** dialog. Notice the square in the Name field - that's the maximize button.
Image

For the **Edit Token** dialog it's less intrusive since it normally doesn't overlap controls, but it can if the dialog is resized horizontally:
Image

### MapTool Info

1.16.2, latest develop (eeddfa5f40)

### Desktop

Linux Mint 22.1

### Additional Context

_No response_

bubblobill:
fixes #4107, #5376, #5423
advances #5147 
### Description of the Change
Maximise button removed
Changes to force fit dialog on available screen-space.

### Possible Drawbacks
Things work differently for different people.
Reduced accessibility.

### Documentation Notes
General dialog improvements to prevent them exceeding screen size.

### Release Notes
General dialog fixes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5432)


bubblobill:
@kwvanderlinde 
It isn't perfect yet, but it puts us back in a good starting position.
Main problem is it doesn't quite do layout correctly for sizing some smaller dialogs, i.e. The scroll-pane shows a scroll-bar when there is ample space to expand. I was mostly concerned with fixing the oversize going off the screen issue.

src/main/java/net/rptools/maptool/client/swing/SwingUtil.java
```diff
@@ -129,12 +117,48 @@ public static boolean isShiftDown(int modifiers) {
...
+    // how big can it be?
+    GraphicsConfiguration gc = outerWindow.getGraphicsConfiguration();
+    Insets insets = outerWindow.getToolkit().getScreenInsets(gc);
+    Rectangle bounds = gc.getDevice().getDefaultConfiguration().getBounds();
+    System.out.println(bounds);
+    System.out.println(insets);
+    bounds =
+        new Rectangle(
+            insets.left,
+            insets.top,
```
kwvanderlinde: It's not used, but for future sanity this should be `bounds.x + insets.left` and `bounds.y + insets.top`.

src/main/java/net/rptools/maptool/client/swing/SwingUtil.java
```diff
@@ -129,12 +117,48 @@ public static boolean isShiftDown(int modifiers) {
...
+    }
+    // centre it
     int x = outerWindow.getLocation().x + (outerSize.width - innerSize.width) / 2;
     int y = outerWindow.getLocation().y + (outerSize.height - innerSize.height) / 2;
 
+    // just to make sure it doesn't cover UI elements
+    x = Math.max(x, insets.left);
+    y = Math.max(y, insets.top);
+    x =
+        x + innerSize.width > insets.left + bounds.width
```
kwvanderlinde: This condition isn't correct. `x` is the position on screen, not relative to `outerWindow`. You need something like
```
x - outerWindow.getLocation().x > insets.left + bounds.width
```

Otherwise, completely well-sized dialogs get thrown way out to the left unless the MT window is at (0, 0).

src/main/java/net/rptools/maptool/client/swing/SwingUtil.java
```diff
@@ -129,12 +117,48 @@ public static boolean isShiftDown(int modifiers) {
...
+            ? insets.left + bounds.width - x + innerSize.width
```
kwvanderlinde: I'm not sure what is meant to be done here in practical terms. But `x` is a position, so it certainly shouldn't be subtracted except from another position. Since `x` can be very large (multiple monitors), this often gives a negative result, i.e., it tries to throw the dialog off the left side of the screen.

src/main/java/net/rptools/maptool/client/swing/SwingUtil.java
```diff
@@ -129,12 +117,48 @@ public static boolean isShiftDown(int modifiers) {
    * @param outerWindow window to be centered over
    */
   public static void centerOver(Window innerWindow, Window outerWindow) {
+    // how big can it be?
+    GraphicsConfiguration gc = outerWindow.getGraphicsConfiguration();
+    Insets insets = outerWindow.getToolkit().getScreenInsets(gc);
+    Rectangle bounds = gc.getDevice().getDefaultConfiguration().getBounds();
+    System.out.println(bounds);
```
kwvanderlinde: Please remove the println logging.

src/main/java/net/rptools/maptool/client/swing/GenericDialog.java
```diff
@@ -96,6 +71,44 @@ public void closeDialog() {
...
+    GraphicsConfiguration gc = getOwner().getGraphicsConfiguration();
+    Insets insets = getOwner().getToolkit().getScreenInsets(gc);
+    Rectangle bounds = gc.getDevice().getDefaultConfiguration().getBounds();
+    return new Dimension(
+        bounds.width - insets.left - insets.right, bounds.height - insets.top - insets.bottom);
+  }
+
+  @Override
+  public Dimension getPreferredSize() {
+    Dimension contentPref = getContentPane().getPreferredSize();
```
kwvanderlinde: I believe this should use `super.getPreferredSize()` otherwise it doesn't account for the titlebar. And that means I have to scroll to see the OK and Cancel buttons even for dialogs that have lots of space and so shouldn't need to scroll.

src/main/java/net/rptools/maptool/client/swing/GenericDialog.java
```diff
@@ -96,6 +71,44 @@ public void closeDialog() {
...
+    System.out.println(getContentPane().getInsets());
+    System.out.println(contentPref);
```
kwvanderlinde: Remove the printlns

kwvanderlinde (changes requested):

src/main/java/net/rptools/maptool/client/swing/SwingUtil.java
bubblobill: Ahh... whoops

src/main/java/net/rptools/maptool/client/swing/GenericDialog.java
bubblobill: It was, then it wasn't, now it is again.

src/main/java/net/rptools/maptool/client/swing/SwingUtil.java
bubblobill: Oh yes it should be. If that position pushes the part off the window off screen it needs a thwacking.

src/main/java/net/rptools/maptool/client/swing/SwingUtil.java
bubblobill: It is, but it makes no sense. I rewrote it to make more sense.

src/main/java/net/rptools/maptool/client/swing/SwingUtil.java
```diff
@@ -129,12 +117,49 @@ public static boolean isShiftDown(int modifiers) {
...
     int y = outerWindow.getLocation().y + (outerSize.height - innerSize.height) / 2;
 
+    // just to make sure it doesn't cover UI elements, i.e. encroach on Screen Insets
+    // Left-hand side
+    x = (int) Math.max(x, usableScreenArea.getMinX());
+    // top side
+    y = (int) Math.max(y, usableScreenArea.getMinY());
+    // Right-hand side
+    if (x + innerSize.width > usableScreenArea.getMaxX()) {
+      x = (int) (usableScreenArea.getMaxX() - innerSize.width - x);
```
kwvanderlinde: This calculation always gives a negative result.

Instead, place the dialog so that its right edge matches the right edge of the screen:
```java
x = (int) usableScreenArea.getMaxX() - innerSize.width;
```

src/main/java/net/rptools/maptool/client/swing/SwingUtil.java
```diff
@@ -129,12 +117,49 @@ public static boolean isShiftDown(int modifiers) {
...
+    Insets insets = outerWindow.getToolkit().getScreenInsets(gc);
+    Rectangle usableScreenArea = gc.getDevice().getDefaultConfiguration().getBounds();
+    usableScreenArea =
+        new Rectangle(
+            usableScreenArea.x + insets.left,
+            usableScreenArea.y + insets.top,
+            usableScreenArea.width - insets.left - insets.right,
+            usableScreenArea.height - insets.top - insets.bottom);
+
+    Dimension maxSize = usableScreenArea.getSize();
```
kwvanderlinde, line 132: Can use `usableScreenArea` everywhere instead of `maxSize` (one less thing to keep track of).

src/main/java/net/rptools/maptool/client/swing/SwingUtil.java
```diff
@@ -129,12 +117,49 @@ public static boolean isShiftDown(int modifiers) {
    * @param outerWindow window to be centered over
    */
   public static void centerOver(Window innerWindow, Window outerWindow) {
+    // how big can it be?
+    // screen area minus OS bits like taskbars
+    GraphicsConfiguration gc = outerWindow.getGraphicsConfiguration();
+    Insets insets = outerWindow.getToolkit().getScreenInsets(gc);
+    Rectangle usableScreenArea = gc.getDevice().getDefaultConfiguration().getBounds();
```
kwvanderlinde, line 124: I don't have a good alternative, but I just wanted to call out the downsides of using the screen bounds here. The `usableScreenArea` isn't really be the available screen space, it is the bounds of the particular screen that the `outerWindow` (MT frame) happens to count as its current monitor.

This has two downsides:
1. It's an ambiguous concept if MT is spread over more than one monitor.
2. Even if there's lots of room for the dialog, the below checks might think it is out of space in one direction or another if the edge of the screen happens to be nearby.

As I said, I don't have an alternative, nor do I think it is worth trying to avoid these relatively minor issues.

kwvanderlinde (changes requested):

src/main/java/net/rptools/maptool/client/swing/SwingUtil.java
bubblobill: Lowest common denominator. This is a Dialogue Box. As such it needs to meet the expectations of the paradigm as well as a whole bunch of design guidelines that are hard to find. 

- This is part of `centerOver` so the dialogue is spawned with the intention of an obvious contextual relationship with the outer window. Usually achieved by being as close as possible to being in the middle of the owner, but.,,
- A dialogue should never cross a hard screen device boundary on creation. If the owner spans multiple devices then you just have to go with whichever ones you're given.
- The whole of the content of the dialogue should be visible. Where this is not possible,
- The whole of the content must be accessible.

In the past, MT has managed to spawn dialogue windows with title bars off the top of the screen, and the button bar off the bottom of the screen leaving users buggered. That is why I started this whole debacle, when I discovered that operating systems do not allow dialog boxes to be resizable, by design. This means they have no minimize/maximise functionality so you can't just use a key combination to fit it on the monitor. You need to rediscover how to move windows using cursor keys which is a pretty big ask for most users.

TBH, the solution to your problem is pretty easy. You just need to solve the concurrency problems with keeping the token editor open and we can stick it in an actual window that remembers its location in logical screen space.

src/main/java/net/rptools/maptool/client/swing/SwingUtil.java
bubblobill: ja, done

src/main/java/net/rptools/maptool/client/swing/SwingUtil.java
bubblobill: No. This is a case of the maths teacher telling you to show your working.
It's why I rewrote the whole thing with big long variable names and stuff so that it is obvious what is happening. Yes, it is functionally pointless, but it says `maxSize` which is hard to misconstrue.

src/main/java/net/rptools/maptool/client/swing/SwingUtil.java
kwvanderlinde: > TBH, the solution to your problem is pretty easy. You just need to solve the concurrency problems with keeping the token editor open and we can stick it in an actual window that remembers its location in logical screen space.

I'll get right on that 😅 Definitely the long-term answer.

kwvanderlinde (approved):

PR #5435, opened 2025-05-13
Fix preferences dialog width depending on confg file path

### LINKED ISSUE #5350: [Bug]: Preference dialog is stretched horizontally when running the zipped app-image
### Describe the Bug

Using the new app-image zips from 1.17, if the zip is extracted in a location with a long path name, then the **Preferences** dialog becomes stretch horizontally.

### To Reproduce

1. In the file explorer, make a deeply nested folder or give a folder an extremely long name.
2. Download the `*.zip` file for your platform and extract it into the folder created in (1).
3. Run the extracted MapTool executable.
4. Click on _Edit_ > _Preferences_ in the main menu.
5. Notice the **Preferences** dialog is really wide. In particular, the _Themes_ tab shows a horizontally stretched image.

### Expected Behaviour

The preferences dialog should not scale according to the length of the installation path.

### Screenshots

Image

### MapTool Info

1.17.0-rc.1

### Desktop

Linux Mint 22.1

### Additional Context

The culprit appears to be the Startup tab, where the installation directory is writtten out on a single line in tthe **Startup Preferences Information** text.

Image

kwvanderlinde:

Fixes #5350

### Description of the Change

This PR factors out the path information in the _Startup_ tab of the **Preferences Dialog**. Now the config file is displayed separately from the "manual copy" text, which also means it is always displayed whether or not that text is shown. When the "manual copy" text is needed, it is now also separate from the ***Startup Preference Information***, and uses a warning icon instead of a heading to get the user's attention.

In the _Themes_ tab, the theme example image is now scaled to fit instead of stretched.

For comparison, the _Startup_ tab in the old **Preferences** dialog looked like this:
image
But now it looks like this:
image

### Possible Drawbacks

This changes how the `startup.preferences.info` and `startup.preferences.info.manualCopy` i18n keys are used. It should not be difficult for translators to update, but without an update they will both look a little odd.

### Documentation Notes

N/A

### Release Notes

- Fixed the preferences dialog to not get large when the config file path is long.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5435)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/preferencesdialog/PreferencesDialog.java
+++ b/src/main/java/net/rptools/maptool/client/ui/preferencesdialog/PreferencesDialog.java
@@ -904,9 +904,7 @@ public class PreferencesDialog extends JDialog {
     // jpackage config files can't be written to. Show a warning to the user describing the
     // situation.
     if (appCfgFile != null) {
-      var warning = I18N.getText("startup.preferences.info.manualCopy");
-      configFileWarningLabel.setText(
-          String.format("<html>%s</html>", warning.replace("\n", "<br>")));
+      configFileWarningLabel.setText(I18N.getText("startup.preferences.info.manualCopy"));
       configFileWarningLabel.setVisible(true);
     } else {
       configFileWarningLabel.setText(null);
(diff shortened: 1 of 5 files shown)


cwisniew (approved):

cwisniew (approved):

bubblobill:
Just to be sure, the Startup tab shows as enabled right?

kwvanderlinde:
> Just to be sure, the Startup tab shows as enabled right?

Yes. It's disabled during development, but is enabled for installs.

Not sure why, makes it a bit of a pain to test.

cwisniew:
> > Just to be sure, the Startup tab shows as enabled right?
> 
> Yes. It's disabled during development, but is enabled for installs.
> 
> Not sure why, makes it a bit of a pain to test.

I think its because the file where the startup information comes from doesn't exist when you do gradle run.
You would probably have to do gradle package and then run the resultant package

PR #5443, opened 2025-05-15
TokenRenderer/Zone Renderer - consistent token flippage on dragging

bubblobill:
addressses #5097 
specifically [this](https://github.com/RPTools/maptool/issues/5097#issuecomment-2881782695)

### Description of the Change
TokenRenderer: 
-flush renderImageMap on grid change.
-cache table images
-changed render function arguments 
ImageUtil:
- removed iso requirement for image tables
- added function to transform provided token image
FacingArrowRenderer: added grid change listener
ZoneRenderer:
- added iso grid checks to token iso flip checks
- a lot of housekeeping.

### Possible Drawbacks
@kwvanderlinde will find even more broken things.

### Documentation Notes
n/a

### Release Notes

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5443)


src/main/java/net/rptools/maptool/client/ui/zone/renderer/tokenRender/FacingArrowRenderer.java
```diff
@@ -75,74 +77,79 @@ public void paintArrow(
...
-    tokenG.draw(facingArrow);
+      tokenG.fill(facingArrow);
+      tokenG.setColor(borderColour);
+      tokenG.draw(facingArrow);
 
-    tokenG.setTransform(oldAT);
+      tokenG.setTransform(oldAT);
+    } catch (Exception e) {
+      log.debug(e);
+    }
```
kwvanderlinde, line 151: I assume you were chasing down an error. Should the `try`/`catch` stay? If so, this should be a `log.error()` with a message alongside the exception.

kwvanderlinde (commented):
Looks good overall. Just the one nitpick from me.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/tokenRender/FacingArrowRenderer.java
bubblobill: Take out the try/catch and watch your MT explode on the launch-pad.

Prior to there being a grid, or footprints, or any of the other niceties you expect to exist in a ZoneRenderer, tokens get asked to be drawn with or without their facing arrows approxiamately 1 milliard times, each one of which will throw an exception. Until I get a better idea of the flow, this is here to make sure it is not the cause of any upsets. Once confident it will change to ignored, and potentially rendered useless and removed.

bubblobill: > I assume you were chasing down an error.

Incorrect. If I'm chasing down an error, I use log.info.
Things only graduate to log.debug when I'm sure they are mostly solved but may be worth noting.

kwvanderlinde: > > I assume you were chasing down an error.
> 
> Incorrect. If I'm chasing down an error, I use log.info. Things only graduate to log.debug when I'm sure they are mostly solved but may be worth noting.

Then it's better to raise the level to `log.warn` or `log.error` rather than reduce it to `log.debug`. That way if it happens in the wild it will actually show up in logs and can help with diagnostics.

bubblobill: changed to `log.error("Failed to paint facing arrow.");`

kwvanderlinde (approved):

PR #5444, opened 2025-05-16
New markdownToHtml() function optional parameter for user selection of flexmark extensions

### LINKED ISSUE #4840: [Feature]: Enable setting of HTML attributes via markdown and markdownToHtml()
### Describe the Problem

When using `markdownToHtml()` function I'm not able to set ids, classes, or other attributes in markdown script to be incorporated into the processed HTML output.

### The Solution you'd like

Add the extension https://github.com/vsch/flexmark-java/tree/master/flexmark-ext-attributes to the to `markdownToHtml()` function and enable all its goodness.

Writing HTML in the markdown script in order to set the attributes, which defeats the purpose of using markdown in the first place.

### Additional Context

_No response_

Baaaaaz:
[4840 [Feature]: Enable setting of HTML attributes via markdown and markdownToHtml()](https://github.com/RPTools/maptool/issues/4840)

You can link a pull request to an issue with one of
closes #4840

### Description of the Change
* New flexmark extensions added: aside, attributes, ins, strikethrough, subscript, superscript
* New `markdownToHtml()` mtscript function optional parameter added to allow user to specify which extensions to use for processing markdown as a comma separated list, or alternatively provide "*" to use all of them.
* moved `options.set` for extensions where relevant to after the optional extensions are added.
* implemented an enum and iterators to improve future option management

### Possible Drawbacks
Hard to fathom flexmark's codebase, but have tried to follow flexmark's own example of [optional extensions](https://github.com/vsch/flexmark-java/blob/bcfe84a3ab6d23d04adce3e5a0bae45c6b791d14/flexmark-java-samples/src/com/vladsch/flexmark/java/samples/BasicProfileSample.java#L42), but not sure if there is an actual issue if an extension is added twice but I think it uses 'keys' so maybe not.  My testing has revealed no issues though.

### Documentation Notes

markdownToHtml

**Usage** Add:
*markdownToHTML(string, type, optionalExtensions)*

**Parameters** Add:
* *optionalExtensions - A comma separated string list of flexmark markdown extensions to be applied in addition to the specified markdown format, including: `"aside", "attributes", "definitions", "ins", "strikethrough", "subscript", "superscript", "tasklist", "tables", "toc"`.  Alternatively, use `"*"` to include all optional extensions.*

**Examples** Add:
```
[r: markdownToHTML(getNotes(), "GITHUB", "toc")]  
[r: markdownToHTML(getNotes(), "MARKDOWN", "ins, toc")]  
[r: markdownToHTML(getNotes(), "GITHUB_DOC", "*")]
```

### Release Notes
New `markdownToHtml()` mtscript function optional parameter for user selection of flexmark extensions

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5444)


src/main/java/net/rptools/maptool/client/functions/MarkDownFunctions.java
```diff
@@ -47,46 +55,102 @@ public class MarkDownFunctions extends AbstractFunction {
...
+        }
+        if (optionalExtensionsList.contains("TASKLIST")) {
+          extensions.add(TaskListExtension.create());
+        }
+        if (optionalExtensionsList.contains("TOC")) {
+          extensions.add(TocExtension.create());
+        }
+      }
+    }
+
```
cwisniew, line 146: This would be better handled with an enum so the code doesn't need to be added in two places if there is a new option.

cwisniew (changes requested):

src/main/java/net/rptools/maptool/client/functions/MarkDownFunctions.java
```diff
@@ -47,46 +55,102 @@ public class MarkDownFunctions extends AbstractFunction {
...
+        }
+        if (optionalExtensionsList.contains("ATTRIBUTES")) {
+          extensions.add(AttributesExtension.create());
+        }
+        if (optionalExtensionsList.contains("DEFINITION")) {
+          extensions.add(DefinitionExtension.create());
+        }
+        if (optionalExtensionsList.contains("INS")) {
+          extensions.add(InsExtension.create());
+        }
```
cwisniew: This would be better handled with an enum so the code doesn't need to be added in two places if there is a new option.

cwisniew (changes requested):

src/main/java/net/rptools/maptool/client/functions/MarkDownFunctions.java
Baaaaaz: @cwisniew Thanks for the review.  I think I have made the requested changes... but I am both a java and git newbie.  Not sure what I have done or was doing with the two merge commits below or how that effects this PR?!  Please can you take another look or advise what I should have done or how to unpick git it if it is a problem?  Thanks.

cwisniew: Yep looks good, thanks

cwisniew (approved):

PR #5451, opened 2025-05-17
Fix for remote client image corruption

### LINKED ISSUE #5403: [Bug]: Token/Map assets showing "received invalid image" error
### Describe the Bug

After installation of 1.16.2, players (but not the GM) when attempting to load a map will receive and error noting "received an invalid image" and list of what seems to be an asset id.  In my particular case this was two tokens created at approximately the same time (may have been done either pre- or post-installation of the new version of Maptool; cannot be certain).  Was necessary to recreate the tokens (go back and pull the art asset into TokenTool) to make the problem go away.
Accompanying will be screenshots of the error reports, and copies of the two tokens producing the errors.

Image
Image
Image
Image

### To Reproduce

There seems no reliable way to duplicate the bug, though two other users on the Discord have spoke about similar situations and error reports post 1.16.2

### Expected Behaviour

Normal map loading behavior.

### Screenshots

See above.

### MapTool Info

Version 1.16.2

### Desktop

Windows 11 Version	10.0.26100 Build 26100

### Additional Context

_No response_

fishface60:

fixes https://github.com/RPTools/maptool/issues/5403

### Description of the Change

This moves the check for whether there is already a request for the asset inside the requestAsset method to remove a race condition where AssetManager would synchronize checking whether there was a request and then request it, which fails when multiple threads are checking whether the asset is cached in parallel, all get the result that there are no pending requests, and then submit multiple requests when there should be only one.

### Possible Drawbacks

This is potentially not a complete solution.

It'll stop the parallel requests which result in corrupt asset requests, but it's possible sequential requests for the same assets might still happen if there's so many tokens it's still enumerating them by the time the asset has been loaded.
The locking might be better further up the call stack, but this fixed it for me.

### Release Notes

* Fixed "Received an invalid image" errors when there are multiple tokens on a map sharing an image larger than 5 Kilobytes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5451)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/AssetLoader.java
+++ b/src/main/java/net/rptools/maptool/model/AssetLoader.java
@@ -86,6 +86,14 @@ public class AssetLoader {
     repositoryStateMap.clear();
   }
 
+  /**
+   * Check whether there is currently a pending request for the given asset ID.
+   *
+   * <p>NOTE: This may be immediately incorrect by the time this function returns so decisions on
+   * whether to request an asset must not depend on the result.
+   *
+   * @return true if there is an incomplete request.
+   */
   public synchronized boolean isIdRequested(MD5Key id) {
     return requestedIdSet.contains(id);
   }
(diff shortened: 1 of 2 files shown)


cwisniew (approved):

FullBleed:
Have you stress-tested this?  I recently began creating a large, randomly generated map where thousands of a single asset are deployed (size, facing, location drop are randomized).  It sounds like you have concerns that such a thing might be problematic.

fishface60:
> Have you stress-tested this? I recently began creating a large, randomly generated map where thousands of a single asset are deployed (size, facing, location drop are randomized). It sounds like you have concerns that such a thing might be problematic.

Seems to work.


https://github.com/user-attachments/assets/9561caae-0e0e-429b-9116-c1a1a2f4fc40

PR #5461, opened 2025-05-21
Fix for API not working when navigating to new page in frame5/dialog5…

### LINKED ISSUE #5460: [Bug]: When navigating to a new page in Frame5/Dialog5/Overlay the API is not available
### Describe the Bug

When navigating to a new page in frame5/dialog5/overlay the MapTool API and other supporting features are not available in the new page.

### To Reproduce

Create a lib:token with URI support, or an addon and have 2 pages where one has a link to navigate to the other. In the first page the API will be available and links to external sites will open an OS browser. When navigating to the new page the API will not be available and links to external sites will attempt to open in the frame/dialog/overlay but fail and just display a blank page.

### Expected Behaviour

Any page navigated to should have all the features that are available to the initially loaded page.

### Screenshots

_No response_

### MapTool Info

MapTool 1.16

### Desktop

All

### Additional Context

_No response_

cwisniew:
Fixes #5460

### Description of the Change

This change makes the features available to all links navigated to in the frame5/dialog5/overlay that the initial page gets and you would expect the subsequent pages to have available. The core change is the introduction of the HTMLContent class to unify and enhance how HTML is injected and managed in the UI. This addresses issues where API features were not functioning correctly when navigating to a new page within frame5/dialog5/overlay as well as moving code related to displaying this content into the one place rather than several other places in the code

Key Changes
- Introduced the new HTMLContent class to manage HTML content and injection of the Java bridge and base URL.
- Updated HTML frame, dialog, and overlay managers to use HTMLContent instead of raw strings, which allows you to pass either a HTML string or a URL to be displayed
- Refactored methods and interfaces to pass around HTMLContent objects, improving extensibility.
- Removed the now-obsolete HTMLUtil class.
- Improved error handling and localization for HTML loading failures.
- Adjusted macro function handling to support URLs and string content better
- Moved the CSP meta into an enum rather than string, this doesn't do much other than allow easier documentation in the code as to what each part is
- Some other general clean up of some comments / unused variables in the code files I touched.
- Method to get the data for an Asset as an input stream
- All content injected into the HTML page to be displayed is now done by parsing the HTML and inserting the nodes (via JSoup) where as before it was all just preappended to the HTML string 

### Possible Drawbacks

Its far from perfect, but to get the webview to work requires some hacky things, this change at least tries to move the majority of those hacky things into one class so others don't need to know about them (apart from the HTMLWebViewManager class which still needs some knowledge as well).

### Documentation Notes
- Fixed bug where there was no API available after navigating to a new page in a frame5/dialog5/overlay

### Release Notes
- Fixed bug where there was no API available after navigating to a new page in a frame5/dialog5/overlay

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5461)


src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLContent.java
```diff
@@ -0,0 +1,418 @@
...
+ *
+ * <p>This class is can also used to inject the Java bridge and the base URL into the HTML content
+ * if needed.
+ *
+ * <p>This class is immutable and thread-safe.
+ */
+public class HTMLContent {
+
+  /** Magic value that window.status must take to initiate the bridge. */
+  public static final String BRIDGE_VALUE = "MY_INITIALIZING_VALUE";
```
kwvanderlinde: This isn't used.

src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLContent.java
```diff
@@ -0,0 +1,418 @@
...
+   *
+   * @param url the URL of the HTML content
+   * @return the HTMLContent object with the Java Bridge ad base URL injected.
+   */
+  public HTMLContent injectJavaBridgeAndBaseUrl(@Nullable URL url) {
+    if (javaBridgeInjected && baseUrlInjected) {
+      return this; // both already injected so return the same object
+    }
+    var newHtml = htmlString;
+    var document = Jsoup.parse(newHtml);
```
kwvanderlinde, line 285: Chance for NPE here as `htmlString` could be `null` at this point.

kwvanderlinde (commented):
Nitpicks only

src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLContent.java
cwisniew: removed

src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLContent.java
cwisniew: Now throws an IllegalStateException if it is null

kwvanderlinde (approved):

PR #5498, opened 2025-05-26
refactor and and vue hack

### LINKED ISSUE #5496: [Feature]: Vue Apps display blank page in frame5/dialog5/overlay
### Describe the Problem

Vue apps created with Vite (and also possibly those created with other methods) are not displayed properly in Frame5/Dialog5/Overlay as there is some sort of weird interaction happening where the div tag with the app id is replaced with nothing rather than the vue app. This can be fixed by altering the build process of the vue app to place the script it generates at the bottom of the body after the div rather than in the head.

### The Solution you'd like

Add a way to inform MapTool that its attempting to load a Vue app so it can preprocess the HTML and move the script to the bottom of the body without needing to alter the Vue build process.

Probably by checking for the query string mtvuefix=true

Making everyone change their build process for Vue apps

### Additional Context

I had originally considered just moving all scripts to the bottom regardless as a solution but decided against it as it may have unintended consequences for existing frameworks

cwisniew:
resolves #5496

### Description of the Change
Specifying the query `mtvuefix=true` on a  lib:// URL will now cause MapTool to move any `script` tags in the HTML being loaded to the bottom of the body. This fixes the issue with Vue pages being blank. This works for links on a page as well as the initial page to be displayed.

I also refactored `HTMLContent` and how HTML pages are fetched/proccessed slightly. I promise this is the last refactor of this area... for 1.18 :)

### Possible Drawbacks
There should be none as it only comes into effect when `mtvuefix=true` is specified and its unlikely anyone is currently doing this.

### Documentation Notes
Specifying `mtvuefix=true` as part of the query string for a lib:// URL will perform a workaround to fix a Vue app if it is loading as a blank page.

### Release Notes
Added workaround for Vue apps that load as a blank page.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5498)


kwvanderlinde (commented):
As I mentioned in [this comment](https://github.com/RPTools/maptool/issues/5496#issuecomment-2915111398), I don't think the `VueHack` is warranted.

The rest of the changes look good to me.

kwvanderlinde (approved):

PR #5515, opened 2025-05-28
New MTScript get table entries functions

### LINKED ISSUE #4859: [Feature]: Add new MT Table functions to improve access to, and searching of, all table entries
### Describe the Problem

1. Unable to search a MT table by the contents of the Value column.  e.g. if I store an Image (asset) in a table entry along with a name (in the corresponding Value column) I cannot easily nor reliably find this image again solely by the name in MT script.

2. Not assured to get all table entries comprehensively from a table via MT Script (well without first relying on careful data entry, scripting, gap/deduplication handling, and some other assumptions).

### The Solution you'd like

To make MT Tables a bit more user friendly for storing and retrieving data I'd like:

A. a new function: `getMatchingTableEntries(name, pattern)`.  Similar to `getTableEntry()` but instead of taking a `roll` argument, takes a regex `pattern` to enable searching the contents of the Value column in the specified table, returning a json ARRAY of OBJECTS  of raw table data of entries where the Value column's contents matches the pattern, or an empty ARRAY [] where there are no matches.

B. a new function: `tableMatchingImages(name, pattern [, size])`.  Similar to `tableImage(name, row [, size])` combined with point A above, but returns a JSON array of image asset values.

C. a new function: `getTableEntries(name)` to return a json ARRAY of OBJECTS of raw table data for all table entries for the specified table.  i.e. similar to `getTableEntry()` but returns all tables entries, or an empty ARRAY if the table has no entries.  Being able to get all entry data from a table natively may also enable other great things such as scripted re-sorting of entries, generating smaller filtered tables, creating merged tables, error checking, picking from a relevant selection of images when creating a token, etc.

Arguably B could be achieved via A with one or two extra MT scripting steps.
Arguably C could be achieved via A if the pattern could be set to match all values.

1. Creating a separate searchable index of row keys, storing it somewhere, and then searching this index to find the roll number to get the table entry from the table.  e.g. Aliasmask's imageDB stores this index in the same table at roll = 0.  However, creating and maintaining such an index when adding or removing table rows is problematic as it is not natively integrated with the table itself and sometimes one has to make assumptions such as row roll values incrementation and you also have to enter table data to match that assumption.

2. Determining the table range (via `getTableRoll()` and then iterating through each row using `getTableEntry()` (assuming that the table roll range is representative of the actual row rolls however - which it may not be as it can be set independently of the actual entry roll ranges), and then finding any value matches on each iteration but also accounting for any gaps or duplicate results.

3. Waiting for... [https://github.com/RPTools/maptool/issues/103](https://github.com/RPTools/maptool/issues/103)

### Additional Context

Proposed function names based on existing `getTableEntry()` and `getMatchingProperties()`.

Requested changes should not break existing table functionality but extends their usefulness and perhaps makes them more intuitive for storing and retrieving text/images, and perhaps lowers any entry barrier for new campaign/framework developers by not forcing them down the json route from day 1.  Also provides another option for seasoned campaign/framework developers.

Baaaaaz:
closes #4859

### Description of the Change
* ~~Added new LookupEntry method `toJson()`~~ (REVERTED)
* Added new three MTScript functions for tables: 
  * getTableEntries
  * getTableEntriesCount
  * getTableEntrybyIndex
* New `convertLookupEntrytoJson()` function 

### Possible Drawbacks
I'm a Java newbie and have not done things in the best way.

### Documentation Notes
Add pages for three new MTScript [table](https://wiki.rptools.info/index.php/Category:Table_Function) get functions which expand the utility of MapTool tables:

**`getTableEntries(name, pattern, limit, offset)`** -> Returns a json array of json objects containing the table entry details filtered by the arguments (where given).
 * `name` - the name of the table to get entries from
 * `pattern` - (optional) filter table entries where their value match the regex pattern, or provide "" to not filter.
 * `limit` - (optional) the maximum number of filtered entries to be returned, or provide 0 to not limit.
 * `offset` - (optional) skip a certain number of filtered entries to be returned, or provide 0 to not offset.

**`getTableEntriesCount(name, pattern)`** -> Returns a count of the number of tables entries
 * `name` - the name of the table to analyse
 * `pattern` - (optional) filter table entries where their value match the regex pattern, or provide "" to not filter.

**`getTableEntrybyIndex(name, index)`** -> Returns a json object containing the table entry details.
 * `name` - the name of the table to get a table entry from
 * `index` - the index of the entry value to return

### Release Notes
-- Added new MTScript functions for getting table entries by pattern match or by index

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5515)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/LookupTableFunction.java
+++ b/src/main/java/net/rptools/maptool/client/functions/LookupTableFunction.java
@@ -311,7 +311,7 @@ public class LookupTableFunction extends AbstractFunction {
       if (rollInt < entry.getMin() || rollInt > entry.getMax()) {
         return ""; // entry was found but doesn't match
       }
-      return convertLookUpEntrytoJson(entry);
+      return convertLookUpEntryToJson(entry);
 
     } else if ("getTableEntryAtIndex".equalsIgnoreCase(function)) {
       /*
@@ -321,12 +321,12 @@ public class LookupTableFunction extends AbstractFunction {
        */
       FunctionUtil.checkNumberParam(function, params, 2, 2);
       String name = params.get(0).toString();
-      int index = ((BigDecimal) params.get(1)).intValue();
+      int index = FunctionUtil.paramAsInteger(function, params, 1, false);
       LookupTable lookupTable = getMaptoolTable(name, function);
       List<LookupEntry> entriesList = lookupTable.getEntryList();
       if (index >= 0 && index < entriesList.size()) {
         LookupEntry entry = entriesList.get(index);
-        return convertLookUpEntrytoJson(entry);
+        return convertLookUpEntryToJson(entry);
       } else {
         return "";
       }
@@ -342,8 +342,9 @@ public class LookupTableFunction extends AbstractFunction {
       FunctionUtil.checkNumberParam(function, params, 1, 4);
       String name = params.get(0).toString();
       String pattern = (params.size() > 1) ? params.get(1).toString() : "";
-      int limit = (params.size() > 2) ? ((BigDecimal) params.get(2)).intValue() : 0;
-      int offset = (params.size() > 3) ? ((BigDecimal) params.get(3)).intValue() : 0;
+      int limit = (params.size() > 2) ? FunctionUtil.paramAsInteger(function, params, 2, false) : 0;
+      int offset =
+          (params.size() > 3) ? FunctionUtil.paramAsInteger(function, params, 3, false) : 0;
 
       LookupTable lookupTable = getMaptoolTable(name, function);
       List<LookupEntry> entriesList = lookupTable.getEntryList();
@@ -351,7 +352,13 @@ public class LookupTableFunction extends AbstractFunction {
       // Filter by regex pattern, then apply offset, then apply limit
       if (!pattern.isEmpty()) {
         entriesList =
-            entriesList.stream().filter(entry -> entry.getValue().matches(pattern)).toList();
+            entriesList.stream()
+                .filter(
+                    entry -> {
+                      assert entry.getValue() != null;
+                      return entry.getValue().matches(pattern);
+                    })
+                .toList();
       }
       if (offset > 0) {
         entriesList = entriesList.stream().skip(offset).toList();
@@ -361,7 +368,7 @@ public class LookupTableFunction extends AbstractFunction {
       }
       JsonArray jarrEntries = new JsonArray();
       for (LookupEntry entry : entriesList) {
-        jarrEntries.add(convertLookUpEntrytoJson(entry));
+        jarrEntries.add(convertLookUpEntryToJson(entry));
       }
       return jarrEntries;
 
@@ -378,7 +385,13 @@ public class LookupTableFunction extends AbstractFunction {
       List<LookupEntry> entriesList = lookupTable.getEntryList();
       if (!pattern.isEmpty()) {
         entriesList =
-            entriesList.stream().filter(entry -> entry.getValue().matches(pattern)).toList();
+            entriesList.stream()
+                .filter(
+                    entry -> {
+                      assert entry.getValue() != null;
+                      return entry.getValue().matches(pattern);
+                    })
+                .toList();
       }
       return entriesList.size();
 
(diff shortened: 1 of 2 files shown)


src/main/java/net/rptools/maptool/model/LookupTable.java
```diff
@@ -447,6 +448,21 @@ public LookupEntryDto toDto() {
...
+      entryDetails.addProperty("picked", this.picked);
+
+      if (this.imageId != null) {
+        entryDetails.addProperty("assetid", "asset://" + this.imageId.toString());
+      } else {
+        entryDetails.addProperty("assetid", "");
+      }
+      return entryDetails;
+    }
   }
```
cwisniew: Please put the conversion to a JsonObject into `LookupTableFunction.java` so that macro specific code doesn't start spreading all over the code base  :)

cwisniew (dismissed):

src/main/java/net/rptools/maptool/model/LookupTable.java
Baaaaaz: That's funny as that was how I had done it originally!!

Baaaaaz:
@cwisniew Has this change missed the1.18 boat and is it possible for it to be added, or does it need to wait for 1.19?

cwisniew:
> @cwisniew Has this change missed the1.18 boat and is it possible for it to be added, or does it need to wait for 1.19?

I will do the first RC in about 10 days, that's when I will freeze new changes for 1.18

Baaaaaz:
I think I have committed the requested changes and re-requested a review, but this PR is still showing as "Changes requested"?  Do I need to do anything else here in github?

src/main/java/net/rptools/maptool/client/functions/LookupTableFunction.java
```diff
@@ -292,32 +295,93 @@ public Object childEvaluate(
...
-        entryDetails.addProperty("assetid", "asset://" + imageId.toString());
+    } else if ("getTableEntryAtIndex".equalsIgnoreCase(function)) {
+      /*
+       * Parameters:
+       * name - the name of the table to get a table entry from
+       * index - the index of the entry value to return
+       */
+      FunctionUtil.checkNumberParam(function, params, 2, 2);
+      String name = params.get(0).toString();
+      int index = ((BigDecimal) params.get(1)).intValue();
```
kwvanderlinde: There is a method `FunctionUtil.parseAsBigDecimal()` that will convert to `BigDecimal` and properly report cases where the parameter is not actually a number. E.g., in this case:
```
int index = FunctionUtil.paramAsBigDecimal(function, param, 1, false);
```

Same advice goes the the `limit` and `offset` parameters of `getTableEntries`.

kwvanderlinde:
> I think I have committed the requested changes and re-requested a review, but this PR is still showing as "Changes requested"? Do I need to do anything else here in github?

Just a quirk of GitHub. It will stay like that until the reviewer dismisses the request, but doesn't mean there is any more to do on your end.

src/main/java/net/rptools/maptool/client/functions/LookupTableFunction.java
Baaaaaz: I think I cribbed the `((BigDecimal) params.get(1)).intValue()` bit from from a dice function... but I need `index` as an `int`, which is better?
```
int index = FunctionUtil.paramAsBigDecimal(function, params, 1, false).intValue();
```
or
```
int index = FunctionUtil.paramAsInteger(function, params, 1, false);
```
I'm not sure why it needs to be checked as a BigDecimal anyway.  The former seems more forgiving for the user, for example accepting `1.2` as an argument and drops the decimal places.  While the latter is stricter it just seems tidier, so I'll go with this unless someone can explain why not!

Baaaaaz:
@kwvanderlinde thanks for the review.  I have made some changes, so ready for another review!

src/main/java/net/rptools/maptool/client/functions/LookupTableFunction.java
kwvanderlinde: You're right, `paramAsInteger()` is better here.

kwvanderlinde (approved):
Sorry this took so long :slightly_smiling_face:

FullBleed:
FYI, `getTableEntryByIndex` does not show up in getInfo.

FullBleed:
I see what's happening... `getTableEntryAtIndex` is what we see in getInfo.

PR #5524, opened 2025-05-31
Fix for ZoneRenderer.getTokenBounds()

### LINKED ISSUE #5508: [Bug]: Token rendering unbearably slow with many tokens on screen
### Describe the Bug

When there are many tokens on the map (several thousand), it can take many seconds to render a single frame in the latest dev builds. In 1.16 and the 1.17 RCs, the same scenario renders in a fraction the time (sluggish, but usable).

### To Reproduce

1. Open this campaign: [many-tokens-perf-test.cmpgn.renamed-to.zip](https://github.com/user-attachments/files/20465919/many-tokens-perf-test.cmpgn.renamed-to.zip)
2. Attempt to zoom or pan the map.
3. Wait a while until the frame shows up.

### Expected Behaviour

Performance for token rendering should be comparable to prior versions.

### Screenshots

_No response_

### MapTool Info

develop (90828db33)

### Desktop

Linux Mint 22.1

### Additional Context

For me, on 1.16.2, the map renders in just shy of 100ms. But on latest develop, it takes almost 9 seconds for a single frame.

kwvanderlinde:

Fixes [this bug](https://github.com/RPTools/maptool/issues/5508#issuecomment-2923370582) introduced by #5490.

### Description of the Change

`ZoneRenderer.getTokenBounds()` was recently accidentally changed from returning bounds in screen space to returning bounds in world space. This caused some functionality, like light icons, stack indicators, and the facing tool to misbehave.

This PR correct this by removing `ZoneRenderer.getTokenBounds()` in favour of callers being more explicit about their needs. If they need screen space, they are responsible for doing the transformation. If they require the bounds to be on screen, they have to explicitly check that.

For light source icons specifically, `ZoneViewModel` is now responsible for deciding their locations. Both `ZoneRenderer` (via `LightSourceIconOverlay`) and `GdxRenderer` use this now instead of duplicating this logic and basing it on `ZoneRenderer.getTokenBounds()`.

The similar method `getMarkerBounds()` was removed as its only use was not itself used for anything.

### Possible Drawbacks

None

### Documentation Notes

N/A

### Release Notes

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5524)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/tool/FacingTool.java
+++ b/src/main/java/net/rptools/maptool/client/tool/FacingTool.java
@@ -87,16 +87,23 @@ public class FacingTool extends DefaultTool {
   public void mouseMoved(MouseEvent e) {
     super.mouseMoved(e);
 
-    if (tokenUnderMouse == null || renderer.getTokenBounds(tokenUnderMouse) == null) {
+    if (tokenUnderMouse == null) {
       return;
     }
-    Rectangle bounds = renderer.getTokenBounds(tokenUnderMouse).getBounds();
 
-    int x = bounds.x + bounds.width / 2;
-    int y = bounds.y + bounds.height / 2;
+    var viewModel = renderer.getViewModel();
+    var position = viewModel.getTokenPositions().get(tokenUnderMouse.getId());
+    if (position == null) {
+      return;
+    }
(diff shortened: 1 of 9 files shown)


cwisniew (approved):

kwvanderlinde:
Fixed the merge conflicts

Issue #167, opened 2018-01-22
Feature Request: Change Grid appearance to a user side setting


Zuljita:
At present, changing a grid's color is set on server side, it would be nice if it could be set on client side to better support users who might need  higher contrast grids.

JamzTheMan:
I see no issue with this and seems like a good idea.

Does a player need a "per map setting" though as well or just a "global grid color override"?

It would probably be easier to throw a button on the top button bar with a grid color override on/off toggle. And then maybe a grid color preference in the Edit -> Preference menu?

On one side, it would be a pain if the GM was switching between 2 maps and you wanted 2 different colors. But on the other side, if you were, say, color blind, having one global colored grid you could activate as needed would be easier?

Syndaryl:
It's probably best on a per-map basis: A white grid will have poor contrast
on a pale background and a black grid will have poor contrast on a dark
background. An option to adjust grid line width would also help.

JamzTheMan:
Definitely more control, just painful that if you needed a high contrast grid color, you will have to do it over and over and over again.....

Another option or in addition to: If meta+G toggles grid, we could add a shift+meta+G to cycle through the colors...

Azhrei:
It might be enough to toggle between a user selected color and it's inverse (instead of cycling through them).  But yes, it's a pain that the GMs grid doesn't propagate to clients as a default.  But if users can change the color on a per map basis, where do we store that?  We could store it on the Player object, but that's shared and if the same player name is used to login multiple times, they'll all get the same color and that may not be desirable.  We can't store it in the Zone either, as that's also shared.  I guess we'll need a Map<Zone, Color> that's stored in AppPreferences, maybe?

JamzTheMan:
Yea, that's why I think a global grid color may just be a better way to go.
It's better than what they have now.
Do people change the grid color that often anyway?

Phergus:
I almost never change the grid color cuz I mostly don't use it.  Sometimes if I'm trying to get some map lined up with the grid and the map is really dark I'll change it but that's not very common.

I like the idea of cycling the grid through a preset palette or just toggling between a color and it's inverse on a hotkey.

Per map seems like a bit much given the current way that grids are done and no per-client zone properties but maybe it is something to target later with a reworking of that area.

Jaggeroth:
I think the grid colour should be a global setting but entirely local. Can't see a reason for anything more complex.

Syndaryl:
*Pops open campaign file*
[...]

PR #5581, opened 2025-06-12
Grid Renderer - Improved appearance of grids

### LINKED ISSUE #3674: [Feature]: Improve visibility of coordinates text over map areas similar in color
### Feature Request

It would make discerning coordinates much easier by eye.

**Context:**
I DM games online screen-sharing a player-mode client MapTool instance on a separate screen in fullscreen mode with grid and coordinates on.
This way, players either don't have to learn to use another app to play, or can play when they do not have access to a computer/laptop (e.g. on a phone). When they want their character to move, they tell me the coordinates and I move the tokens for them.
**Over map areas or backgrounds similar in color (e.g. using the built-in "Sand" texture) to the coordinates text it is hard to see the coordinates, especially when screen-sharing a high resolution window to a low resolution device (e.g. phone).**
Under these circumstances it happens a lot that players have trouble discerning the coordinates.

Resolution probably connected to #3670, #3671, #3672, #3673.

### The Solution you'd like

Ability to change color of coordinates text, and/or ability to change color and/or thickness of edge around characters in coordinates text.

It occurred to me to make the coordinates text color adaptive to the map area underneath it, but ...

1. ... using the inverse of the colors under the coordinates text characters would probably make it ugly and probably less readable than currently if the map area under the characters is finely detailed and/or colorful.
2. ... using some smart algorithm to choose a color for the characters that would be reasonably visible over their current background (and not ugly) would be probably very hard to come up with (if at all possible), and also unreasonably computationally intensive (especially, considering that it would probably have to be recomputed every time the map is moved or zoomed).

### Additional Context

_No response_

bubblobill:
progresses #3840, #167
resolves #3674

### Description of the Change
Finally realises the intent of my change for GridLineWidth to GridLineWeight.

Separates the rendering of the grid from the construction of the grid shape(s) to draw.
Square, Iso, and Hex Grid create the shape to draw and pass it back to GridRenderer for actual painting.
Square Grid - changed coordinates to have drop shadow on all sides to improve legibility.

GridRenderer 
- scales drawing stroke according to zone scale.
- bases stroke on size of grid (2% of size) for consistent appearance between grid sizes.
- paints grids with a translucent contrast shadow above 50% zoom.

Did not touch the LibGDX stuff other than a refactor.

### Possible Drawbacks
Should be none.

### Documentation Notes
Improved appearance of grids

### Release Notes

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5581)


src/main/java/net/rptools/maptool/client/ui/zone/renderer/GridRenderer.java
```diff
@@ -14,31 +14,93 @@
...
 
 public class GridRenderer {
-  private final Zone zone;
-  private final ZoneRenderer renderer;
+  private ZoneRenderer renderer;
+  private Zone zone;
+  private static float gridLineWeight;
+  private static float scale;
+  private static float baseWidth = 2f;
+  private static Color[] gridColours;
```
kwvanderlinde, line 33: When the current grid changes (either by modifying a map or loading a different map), the `gridColours` needs to be recalculated.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/GridRenderer.java
```diff
@@ -14,31 +14,93 @@
...
+  }
+
   public void renderGrid(Graphics2D g, PlayerView view) {
-    int gridSize = (int) (zone.getGrid().getSize() * renderer.getScale());
-    if (!AppState.isShowGrid() || gridSize < ZoneRendererConstants.MIN_GRID_SIZE) {
+    if (zone == null
+        || !AppState.isShowGrid()
+        || zone.getGrid().getSize() * renderer.getScale() < ZoneRendererConstants.MIN_GRID_SIZE) {
+      if (renderer != null) {
+        setRenderer(renderer);
```
kwvanderlinde: I don't think this call does anything.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/GridRenderer.java
```diff
@@ -14,31 +14,93 @@
...
 
-  public GridRenderer(ZoneRenderer renderer) {
+  GridRenderer(ZoneRenderer renderer) {
+    new MapToolEventBus().getMainEventBus().register(this);
+    setRenderer(renderer);
+  }
+
+  public void setRenderer(ZoneRenderer renderer) {
     this.renderer = renderer;
+    if (renderer.getZone() == null) {
```
kwvanderlinde: Renderers don't have null zones.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/GridRenderer.java
kwvanderlinde, line 1: What do you think about moving the various `Grid#draw(...)` implementations into `GridRenderer`? It would avoid the need to make half the class static and should simplify some of the state management.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/GridRenderer.java
bubblobill: Sure they do. Take it out and discover a whole new world.

bubblobill: No zone during ZR  construction.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/GridRenderer.java
bubblobill: There's a reason this is step one on the journey to prettier grids.

This gets gridRenderer actually painting the various grids.

I intend to construct the grids in a consistent fashion later whilst adding fruit.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/GridRenderer.java
kwvanderlinde:
```java
  public ZoneRenderer(Zone zone) {
    if (zone == null) {
      throw new IllegalArgumentException("Zone cannot be null");
    }
    this.zone = zone;
```
They shouldn't, not even during construction.

Regardless, you can also avoid it by passing the `Zone` to the constructor like many other renderers do and you can dispense with this and other null checks for the zone.

bubblobill: This is a minimal-change/initial steps PR. Keeping the previous bullet-proofing and stuff.
I will probably institute your RenderBuddy class down the road. For now, all these checks and stuff, make sure that ZRs without Zones, and Zones without Grids, etc. don't cause screwups. Also, it fails tests if I remove the zone null check.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/GridRenderer.java
bubblobill: Sure it does. If there is no Zone, set the renderer again. Rinse and repeat.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/GridRenderer.java
bubblobill: Sure, and the GridChanged event actually works. </sarcasm>

src/main/java/net/rptools/maptool/client/ui/zone/renderer/GridRenderer.java
kwvanderlinde: Ugh, this is the problem with handling nulls in such a dynamic way.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/GridRenderer.java
kwvanderlinde: Just checked the tests, they work fine without the null handling.

bubblobill: Well they were failing on something somewhere until I spammed null handling everywhere.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/GridRenderer.java
kwvanderlinde: Creating the `gridColours` array is not at all expensive. You might as well just do it every time in `renderGrid()` and don't worry about caching it.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/GridRenderer.java
```diff
@@ -14,31 +14,93 @@
...
+  private static float baseWidth = 2f;
+  private static Color[] gridColours;
 
-  public GridRenderer(ZoneRenderer renderer) {
+  GridRenderer(ZoneRenderer renderer) {
+    new MapToolEventBus().getMainEventBus().register(this);
+    setRenderer(renderer);
+  }
+
+  public void setRenderer(ZoneRenderer renderer) {
```
kwvanderlinde: Method should be `private`.

kwvanderlinde:
As long as the grid colour stuff is made to work, I'm happy with the PR. I'd encourage you to double-check the need for the null checks, but I'm not going to push that too hard.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/GridRenderer.java
bubblobill: Oh no. Can't do that.
Step 2. Calculate grid colours based on type of display. Line - dashed - dotted - multiply. Calculate negative colour using perception based calculations. Interpolate colours using easing functions. 
2 a) Make considerable more than 4 painting passes, or 
2 b) Create an image to paint instead, or
2 c) Find exotic things on the internet to try.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/GridRenderer.java
bubblobill: Nope. Because GridChanged is broken, this will be called in ZR onGridChanged. In fact, I'll do that now and add colour calculation to setRenderer().

kwvanderlinde: In that case, do something a little different: set `renderer` directly in the constructor and leave it at that. Then have a different method (`gridChanged()` or whatever you want to call it) that can udpate the the `GridRenderer` state based on the `renderer` that you already have.

That makes it a lot easier to follow the intent compared to possibly having arbitrary `ZoneRenderer` being injected at any time.

bubblobill: changes to grid colour via the grid edit tool should trigger gridchanged event, which triggers ZR to update renderer on grid dependant renderers.

kwvanderlinde: Please re-read what I wrote.

You don't need to support setting some arbitrary `ZoneRenderer` since `GridRenderer` _already has the `ZoneRenderer`._ This would suffice instead:
```java
  public void onGridChanged() {
    this.zone = renderer.getZone();
    setGridColours();
  }
```

bubblobill: Then it will have to wait until not 2:45am.

kwvanderlinde: Oh yeah, get some sleep :sweat_smile:

bubblobill: sorted

kwvanderlinde (approved):

PR #5592, opened 2025-06-13
Reduce workflow permissions and other fixes

### LINKED ISSUE #5591: [Maintenance]: Constrain workflows
### Describe the problem

Most of our workflow YAML configs do not have specific permissions and thus inherit some very expansive permissions.

Add `permissions` to each job (or workflow as a whole).

### Expected Benefits

Less chance of undesirable behaviour from workflows.

### Additional Context

_No response_

kwvanderlinde:

Fixes #5591

### Description of the Change

Three workflows now have explicit permissions:
- **Publish Assets** requires `content: write` in order to upload assets.
- **Build Verification** requires `content: read` in order to checkout code and run tests.
- **Nightly Release** requires `content: write` in order to create and delete releases. It also now uses the token defined by the workflow permissions, not the token stored in GH secrets.

The other workflow `*.yml` files already declare permissions.

I found some further "issues" while testing this (none of these have a material impact on workflows):
- `dependabot.yml` has some old ignore lists that are now removed.
- `nightly-publish.yml` now detects the latest stable tag properly, and can handle the case where there are no previous nightlies to delete.
- `nightly-publish.yml` no longers runs on forks, but only on the `RPTools/maptool` repository.

### Possible Drawbacks

Should be none. But if I made permissions too strict then some workflows could start failing.

### Documentation Notes

N/A

### Release Notes

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5592)


bubblobill (approved):
Let's see if it breaks.

PR #5595, opened 2025-06-17
Go back to using the NIGHTLY_BUILD_GH_TOKEN secret

kwvanderlinde:

Fixes nightly build after PR #5591

### Description of the Change

Github has a Creative™ design where jobs will only trigger subsequent jobs if they use Personal Access Tokens (PATs) for permissions, but not if they use `GITHUB_TOKEN`. This is true even if `GITHUB_TOKEN` is defined by the workflow's declared permissions rather than being the general permissions for the repository.

This prevents us from using explicit and constrained permissions in `nightly-publish.yml` as doing so will not allow the `publish.yml` workflow to trigger afterwards.

### Possible Drawbacks

None

### Documentation Notes

N/A

### Release Notes

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5595)

### DIFF
--- a/.github/workflows/nightly-publish.yml
+++ b/.github/workflows/nightly-publish.yml
@@ -1,5 +1,10 @@
 name: Nightly Release
 
+# Note that we use the NIGHTLY_BUILD_GH_TOKEN PAT in this workflow, and do not define permissions
+# for GITHUB_TOKEN. If we use GITHUB_TOKEN, GitHub will not trigger downstream workflows, in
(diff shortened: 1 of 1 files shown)


cwisniew (approved):
Sorry, I should have caught this, I ran into the same problem when creating the job and testing it on my repo.

You are right, though, GitHub's permission scheme is very "creative". Especially good at creating headaches.

PR #5598, opened 2025-06-18
Reduce the statefulness of HaloRenderer

### LINKED ISSUE #5597: [Bug]: 1.18 - Halos do not show up until the map is modified.
### Describe the Bug

The first time a token is given a halo after opening MapTool, the halo does not appear. If the map is subsequently modified, or the campaign is reloaded, then the halo appears.

### To Reproduce

1. Open MapTool with a blank campaign.
2. Add a token to the map.
3. Right-click the token and give it a white halo. Notice no halo appears.
4. Edit the map (don't actually need to change anything) and hit "OK". Now the white halo appears on the token.

### Expected Behaviour

Halos should always render for tokens that have them.

### Screenshots

_No response_

### MapTool Info

1.18.0-rc.1

### Desktop

Linux Mint 22.1

### Additional Context

_No response_

kwvanderlinde:

Fixes #5597

### Description of the Change

This removes some unnecessary state from `HaloRenderer`, reducing the possibility that the rendering logic does not match user expectations.

There's a few dimensions to this:
- The zone for the `HaloRenderer` is injected in the constructor so it doesn't have to later be looked from a `ZoneRenderer`. This ensures the renderer is always properly initialized with consistent state, and removes the need for a `setRenderer()` method separate from `gridChanged()`.
- The static `GRID_SHAPE_MAP` was a memory leak (never cleared) and was also not needed since `Grid` objects cannot migrate between zones or renderers, and therefore different `HaloRenderer` instances have no need to share this cache. Instead, the non-static `shapeMap` suffices as a cache in its own right that just needs to be cleared whenever the grid changes.
- The halo shape (now in the `cachedHaloShape` field) is calculated on-demand rather than when the grid changes. This ensures that if, for some reason, the zone does not have a grid at a certain point in time, then there is still a chance to calculate it at the next render.
- The `paintState` field was only there to pass data from `renderHalo()` to `paintLineHalo()`. This is now a method parameter instead of a field.
- `HaloRenderer` subscribes itself to `GridChanged` events rather than relying on the `ZoneRenderer` to tell it when the grid changes.

### Possible Drawbacks

None

### Documentation Notes

N/A

### Release Notes

- Fixed a bug where halos would not render until a grid was changed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5598)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/renderer/HaloRenderer.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/renderer/HaloRenderer.java
@@ -21,74 +21,77 @@ import java.awt.geom.Ellipse2D;
 import java.util.HashMap;
 import java.util.Map;
 import net.rptools.maptool.client.AppPreferences;
-import net.rptools.maptool.client.AppState;
 import net.rptools.maptool.client.ui.zone.ZoneViewModel;
 import net.rptools.maptool.model.*;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 
 public class HaloRenderer {
-  private static final Logger log = LogManager.getLogger(HaloRenderer.class);
   private final RenderHelper renderHelper;
-  private static final Map<Grid, Map<TokenFootprint, Shape>> GRID_SHAPE_MAP = new HashMap<>();
-  private Map<TokenFootprint, Shape> shapeMap;
-  private Grid grid;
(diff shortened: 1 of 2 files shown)


bubblobill:
Damn you. This was next on my hitlist.

bubblobill (approved):
Oh look, you did just what I was planning to do.
Go team.

Issue #5607, opened 2025-06-20
[Bug]: 1.18 Edit Token - Properties Tab - drop down selection requirement


Jmr3366:
### Describe the Bug

If more than one campaign property exists and you open the Edit Token dialog, The token properties _tab_ will not update until you make a selection in the properties drop down.

### To Reproduce

In 1.18.rc1, new campaign.
Create a new Token Type under Campaign Properties
Drag a token on the map
Select the new token type for this token
Click Ok.

Opening edit token dialog leaves the properties field _empty_

https://github.com/user-attachments/assets/ffc205c5-c1ed-4ac0-bb82-d0f4525d67a7

### Expected Behaviour

Edit Token dialog, Token Properties tab, should update with the properties of that Token Type.

### MapTool Info

1.18.rc1

### Desktop

Win 11

kwvanderlinde:
There is an error as well that is indirectly caused by this. It _should_ be avoided since #5609, but the code is still pretty weak and could use further improvement.

This error shows up when pushing "OK" after getting the blank properties table.

<pre>
java.lang.ClassCastException: class com.jidesoft.grid.PropertyTableModel cannot be cast to class net.rptools.maptool.client.ui.token.dialog.edit.EditTokenDialog$TokenPropertyTableModel (com.jidesoft.grid.PropertyTableModel and net.rptools.maptool.client.ui.token.dialog.edit.EditTokenDialog$TokenPropertyTableModel are in unnamed module of loader 'app')
	at net.rptools.maptool.client.ui.token.dialog.edit.EditTokenDialog.commit(EditTokenDialog.java:930)
	at net.rptools.maptool.client.ui.token.dialog.edit.EditTokenDialog.lambda$initOKButton$11(EditTokenDialog.java:786)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1972)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2314)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:407)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
	at java.desktop/javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:279)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6621)
[...]

PR #5619, opened 2025-06-21
1.18 - Avoid potential for cast exceptions for token property table

kwvanderlinde:

Improves #5607

### Description of the Change

Instead of just casting the property table's model, we now check that it is of the expected type and merely log any unexpected types.

To avoid the possibility of these unexpected types, the table is now initialized with an empty `TokenPropertyTableModel`, and when updated will always be assigned a `TokenPropertyTableModel` (never a `DefaultTableModel`).

### Possible Drawbacks

None

### Documentation Notes

N/A

### Release Notes

- Fixed an error where the Edit Token dialog would show an error after pressing "OK" if there are no properties displayed for the token.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5619)


bubblobill (approved):
approved

PR #5623, opened 2025-06-23
Generic dialogue rewrite

### LINKED ISSUE #5147: [Feature]: All windows should have their exit and save buttons visible

### LINKED ISSUE #5505: [Bug]: Preferences dialog cut off at right side of screen
### Describe the Bug

MapTool v1.17.0-rc2 does some sizing magic with the Edit/Preferences... Dialog.
This causes the right side of the dialog to get cut off.
It also causes the theme example to stretch weirdly.

### To Reproduce

1. Go To Edit
2. Click on Preferences...

### Expected Behaviour

Don't cut off parts of the dialog when resizing.
Respect aspect ratio's of static images used in the dialog.

### Screenshots

Image

### MapTool Info

shadow jar from tag 1.17.0-rc2

### Desktop

Mint Linux 22.1

### Additional Context

screen resolution tested: 1366x768

### LINKED ISSUE #5606: [Maintenance]: GenericDialog can be awkward to implement
### Describe the problem

It is hard to subclass GenericDialog because the constructor requires the content panel which often needs initialising first.

There is also a fair amount of duplication for things like Ok/Cancel buttons that could be generated as part of the GenericDialog class.

The ability to set and access the content post construction.
The ability to provide a consistent button bar for action buttons.

### Expected Benefits

Easier coding, more consistent code.

### Additional Context

_No response_

bubblobill:
closes #5606, #5505, #5147, .. and more I probably cannot find

### Description of the Change
Dialogues employ Jide Resizable.
Dialogue action buttons are now created as part of the dialog and not the content panel.
Buttons are added to the bottom of the pane in a scrollable button bar and should always be visible.
Buttons are laid out in order determined by OS (improved accessibility)
Button mnemonic keys are now separate i18n strings.
Content is placed within a scroll pane to ensure that oversize content can still be accessed.
Factory makes creating dialogues easier, and allows building in different steps, e.g.

```
private final GenericDialogFactory dialogFactory = GenericDialog.getFactory()
          .setDialogTitle(I18N.getText("action.addIconSelector"))
          .addButton(ButtonKind.CANCEL)
          .setCloseOperation(WindowConstants.DISPOSE_ON_CLOSE);
```
```
public AddResourceDialog() {
    super(new AddRessourcesDialogView().getRootComponent());
    setPreferredSize(new Dimension(550, 300));
    panelInit();
    dialogFactory
        .setContent(this)
        .addButton(
            ButtonKind.INSTALL,
            e -> {
              install = true;
              if (commit()) {
                dialogFactory.getDialog().closeDialog();
              }
            })
        .setDefaultButton(ButtonKind.INSTALL);
```

Still have to add missing I18n strings for buttons.

### Possible Drawbacks
Nightmare of a million dialogs might mean conversion isn't 100% yet.

### Documentation Notes
General dialog improvements

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5623)


bubblobill:
@kwvanderlinde pinging you as you tend to be my main dialog conflict source

src/main/resources/net/rptools/maptool/language/i18n.properties
```diff
@@ -163,7 +236,7 @@ Label.startup       = Startup
 Label.themes        = Themes
 Label.application   = Application
 Label.preferences   = Preferences
-Label.save          = Save
```
kwvanderlinde, line 166: Need to keep this as it is still used in the **Preferences** dialog.

src/main/java/net/rptools/maptool/client/MapTool.java
```diff
@@ -1335,6 +1335,7 @@ else if (AppPreferences.loadMruCampaignAtStart.get()) {
             AppActions.loadCampaign(campaignFile);
           }
         } catch (NoSuchElementException nse) {
+          AppPreferences.loadMruCampaignAtStart.set(false);
```
kwvanderlinde, line 1338: Just because there isn't a recent campaign this time, doesn't mean there won't be one next time. Let's not modify user preferences without their input.

src/main/resources/net/rptools/maptool/language/i18n.properties
```diff
@@ -70,11 +70,116 @@ AutoSaveManager.status.autoSaveComplete = Autosave complete.  Elapsed time (ms):
...
+Button.moveDown.mnemonic    = {down}
+Button.moveUp               = Move Up
+Button.moveUp.mnemonic      = {up}
+Button.new                  = New
+Button.new.mnemonic         = n
+Button.next                 = Next
+Button.next.mnemonic        = n
+Button.no                   = No
+Button.no.mnemonic          = n
+Button.ok                   = Okay
```
kwvanderlinde: Keep this as "OK" as that is the norm for UIs.

src/main/java/net/rptools/maptool/client/ui/startserverdialog/StartServerDialog.java
```diff
@@ -60,16 +61,22 @@ public class StartServerDialog extends AbeillePanel<StartServerDialogPreferences
...
     super(new StartServerDialogView().getRootComponent());
     panelInit();
+    dialogFactory
+        .setContent(this)
+        .addButton(
+            ButtonKind.NETWORKING_HELP,
+            e -> {
+              // We don't have a good, server-side way of testing anymore.
+              boolean ok = MapTool.confirm("msg.info.server.networkingHelp");
+              if (ok) MapTool.showDocument(I18N.getString("msg.info.server.forumNFAQ_URL"));
```
kwvanderlinde: Please add braces to this `if`.

src/main/java/net/rptools/maptool/client/ui/startserverdialog/StartServerDialog.java
```diff
@@ -226,17 +225,18 @@ protected void preModelBind() {
   }
 
   public void initOKButton() {
-    getOKButton()
+    dialogFactory
+        .getOKButton()
         .addActionListener(
             e -> {
-              if (getPortTextField().getText().length() == 0) {
+              if (getPortTextField().getText().isBlank()) {
```
kwvanderlinde: Let's avoid the change in behaviour and use `.isEmpty()` instead. Same goes for the other replacements uses of `isBlank()` in this PR.

src/main/java/net/rptools/maptool/client/swing/ButtonKind.java
```diff
@@ -0,0 +1,112 @@
...
+  final String i18nMnemonicKey;
+
+  ButtonKind(String name, String buttonPanelButtonType, String i18nKey, String i18nMnemonicKey) {
+    this.name = name;
+    this.buttonPanelButtonType = buttonPanelButtonType;
+    this.i18nKey = i18nKey;
+    this.i18nMnemonicKey = i18nMnemonicKey;
+  }
+
+  public int getMnemonicKey() {
```
kwvanderlinde: Might as well move this look up into the constructor so it's only calculated once and will be guaranteed to be consistent during the runtime.

src/main/java/net/rptools/maptool/client/swing/ButtonKind.java
```diff
@@ -0,0 +1,112 @@
...
+    String keyString = I18N.getText(this.i18nMnemonicKey).trim();
+    if (keyString.length() == 1) {
+      return KeyEvent.getExtendedKeyCodeForChar(keyString.charAt(0));
+    } else {
+      // Literal strings, should not be translated.
```
kwvanderlinde: Since these shouldn't be translated, let's hardcode them. Add an optional parameter to the constructor for the buttons that need it. If not provided, then try to look up a mnemonic from i18n.

kwvanderlinde (changes requested):

src/main/java/net/rptools/maptool/client/ui/startserverdialog/StartServerDialog.java
bubblobill: isEmpty returns false for white-space.

A value of "  " is not useful. 

Unless I have blank and empty backwards.

src/main/java/net/rptools/maptool/client/swing/ButtonKind.java
bubblobill: Good thinking.

src/main/resources/net/rptools/maptool/language/i18n.properties
bubblobill: It still exists. I just changed the key string to Button.save

kwvanderlinde: Alright, you got me. Either way, **Preferences** won't open anymore.

src/main/java/net/rptools/maptool/client/ui/startserverdialog/StartServerDialog.java
kwvanderlinde: You're right about their behaviours: `isEmpty()` is the same as `length() == 0` while `isBlank()` matches whitespace.

But that's my point. This change is tangential to the rest of the PR, and my gut says there's going to be some knock-on effects. I don't disagree that blanks for some of these fields are not particular useful, and aren't really handled well anyway.

src/main/resources/net/rptools/maptool/language/i18n.properties
bubblobill: It looks like the refactor didn't take. This is why I normally avoid renaming i18n keys like the plague. I'll see if I can make it obey me.

src/main/resources/net/rptools/maptool/language/i18n.properties
bubblobill: Whilst I disagree with you. You may find this amusing
image
Apparently the ButtonPanel class ignores me

src/main/java/net/rptools/maptool/client/ui/startserverdialog/StartServerDialog.java
bubblobill: done

src/main/java/net/rptools/maptool/client/MapTool.java
bubblobill: a) I invented "loadMruCampaignAtStart" so I should be able to do what I want with it.
b) This resulted from me trying to load a buggy campaign during development. What I actually want is to turn it off if the last time it tried it failed. I will make it a separate FR.

src/main/java/net/rptools/maptool/client/swing/ButtonKind.java
bubblobill: added a hard-coded map

src/main/java/net/rptools/maptool/client/ui/startserverdialog/StartServerDialog.java
bubblobill: changed to isEmpty

kwvanderlinde (approved):

PR #5644, opened 2025-06-26
Warn when resetting the initiative round counter

### LINKED ISSUE #5617: [Feature]: Reduce the risk of accidentally resetting initiative and the clearing round counter via the UI
### Describe the Problem

I'm always frustrated when... in the middle of running a combat I accidentally click the Initiative Panel's minibutton which resets all token initiatives and clears the round counter when I did not mean to do so.  It is so close to the other minibuttons and has an un-undoable affect.

### The Solution you'd like

Any of the following would help:
- Move it to the far right of the Initiative Panel toolbar (beyond the round counter even).
- Color it red.
- Add a confirmation dialog "Are you sure you want to reset the initiative? Yes/No".
- Add an undo feature.

_No response_

### Additional Context

Image

Baaaaaz:
closes #5617

### Description of the Change

1. Adds an AppPreference for warning about resetting the round counter (false by default - i.e. current behavior)
2. Adds an initiative panel popup menu item to allow setting the preference (visible to GMs only)
3. If enabled shows a confirmation dialog to a GM when clicking on the reset round counter button on the initiative panel
4. Moves the reset counter mini button to the right of the round counter (i.e. away from the initiative PREV/HOLD/NEXT  buttons)

### Possible Drawbacks
Slight change to the interface layout, but in the interests of reducing risks (i.e. for even when the confirmation is disabled)

### Documentation Notes

Initiative Panel Toolbar Reset Counter mini-button new placement:
image

Initiative Panel Menu Item:
image

Confirmation Dialog:
image

### Release Notes
- Moved button for resetting the initiative round counter and added optional reset confirmation step.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5644)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppPreferences.java
+++ b/src/main/java/net/rptools/maptool/client/AppPreferences.java
@@ -390,6 +390,9 @@ public class AppPreferences {
   public static final Preference<Boolean> showInitiativeGainedMessage =
       store.defineBoolean("showInitGainMessage", true);
 
+  public static final Preference<Boolean> initiativePanelWarnWhenResettingRoundCounter =
+      store.defineBoolean("initWarnWhenResettingRoundCounter", false);
+
   public static final Preference<Boolean> pathfindingEnabled =
       store.defineBoolean("useAstarPathfinding", true);
 
--- a/src/main/java/net/rptools/maptool/client/AppPreferences.java
+++ b/src/main/java/net/rptools/maptool/client/AppPreferences.java
@@ -391,7 +391,7 @@ public class AppPreferences {
       store.defineBoolean("showInitGainMessage", true);
 
   public static final Preference<Boolean> initiativePanelWarnWhenResettingRoundCounter =
-      store.defineBoolean("initWarnWhenResettingRoundCounter", false);
+      store.defineBoolean("initWarnWhenResettingRoundCounter", true);
 
   public static final Preference<Boolean> pathfindingEnabled =
       store.defineBoolean("useAstarPathfinding", true);
--- a/src/main/java/net/rptools/maptool/client/ui/tokenpanel/InitiativePanel.java
+++ b/src/main/java/net/rptools/maptool/client/ui/tokenpanel/InitiativePanel.java
@@ -882,16 +882,19 @@ public class InitiativePanel extends JPanel
             return;
           }
 
-          int reset = JOptionPane.YES_OPTION;
+          boolean reset;
           if (AppPreferences.initiativePanelWarnWhenResettingRoundCounter.get()) {
             reset =
-                JOptionPane.showConfirmDialog(
-                    null,
-                    I18N.getText("initPanel.warnWhenResettingRoundCounter.confirm"),
-                    I18N.getText("msg.title.messageDialogConfirm"),
-                    JOptionPane.YES_NO_OPTION);
+                (JOptionPane.showConfirmDialog(
+                        null,
+                        I18N.getText("initPanel.warnWhenResettingRoundCounter.confirm"),
+                        I18N.getText("msg.title.messageDialogConfirm"),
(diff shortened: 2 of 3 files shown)


src/main/java/net/rptools/maptool/client/AppPreferences.java
```diff
@@ -390,6 +390,9 @@ public class AppPreferences {
   public static final Preference<Boolean> showInitiativeGainedMessage =
       store.defineBoolean("showInitGainMessage", true);
 
+  public static final Preference<Boolean> initiativePanelWarnWhenResettingRoundCounter =
+      store.defineBoolean("initWarnWhenResettingRoundCounter", false);
```
kwvanderlinde: I think we can default to `true`, even though it's a change in behaviour.

src/main/java/net/rptools/maptool/client/ui/tokenpanel/InitiativePanel.java
```diff
@@ -92,6 +92,10 @@ public class InitiativePanel extends JPanel
    */
   private boolean initStateSecondLine = AppPreferences.initiativePanelShowsInitiativeOnLine2.get();
 
+  /** Flag indicating whether to display a confirmation dialog when resetting the round counter. */
+  private boolean warnWhenResettingRoundCounter =
```
kwvanderlinde: I'd recommend skipping the field and just checking the preference directly when needed. It just avoids any possibility of desyncing.

src/main/java/net/rptools/maptool/client/ui/tokenpanel/InitiativePanel.java
```diff
@@ -859,10 +887,24 @@ public void actionPerformed(ActionEvent e) {
             return;
           }
 
-          list.startUnitOfWork();
-          list.setRound(-1);
-          list.setCurrent(-1);
-          list.finishUnitOfWork();
+          int reset;
```
kwvanderlinde: Changing this to a `boolean` will make the intent clearer for future readers.

kwvanderlinde (changes requested):

src/main/java/net/rptools/maptool/client/AppPreferences.java
Baaaaaz: Okay, will do.

src/main/java/net/rptools/maptool/client/ui/tokenpanel/InitiativePanel.java
Baaaaaz: Okay, will do.  If the review rework is okay and to get some consistency (within the class at least!), I could do the same to similar other preference checks within this class either as a new maintenance ticket or this one?

src/main/java/net/rptools/maptool/client/ui/tokenpanel/InitiativePanel.java
Baaaaaz: Hmm, but all the `JOptionPane` stuff returns `int`s.  Are you suggesting converting these all to `boolean`?  What if instead I set the value of `reset` when it is declared, that may make it sufficiently clearer to future readers?  Will try updating that now, let me know what you think?

kwvanderlinde: No, nothing so drastic. I was just thinkg the `else` branch could set to `reset = true`, while the `if` would compare the dialog result against `JOptionPane.YES_OPTION`.

Baaaaaz: Ah okay.  I have made these suggested changes.  Thanks.

kwvanderlinde (approved):

PR #5650, opened 2025-06-27
Set the bundle address on the i18n helper for handlebars.

### LINKED ISSUE #4141: [Feature]: Localised field labels for stat-sheet handlebars templates
### Describe the Problem

Stat-sheet templates using handlebars don't have field names that have been localised. 
This means stat-sheet templates, particularly the built-in ones will not be truly border agnostic.

### The Solution you'd like

Include standard field names --
Name, GM, Label, Bubble/Callout, PC, NPC, Properties, Property, Name, Value, and Notes
as handlebars variables.

_No response_

### Additional Context

_No response_

bubblobill:
closes #4141 

### Description of the Change
We already had the functionality, it just wasn't set up correctly.
Added the properties file address to the Handlebars helper and now it flies.

### Possible Drawbacks
crazy people

### Documentation Notes
Added i18n functionality to Handlebars

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5650)


cwisniew:
For the draw backs you missed one big one

this functionality means you can no longer change the names of keys or remove or keys from the property bundle file once it's merged 

One a key goes into the i18n file and it's released is there forever. Which means you will no longer be able to do the reorganization and consolidation you periodically do

bubblobill:
Spoilsport.

bubblobill:
@cwisniew Made a whole new resource bundle for stat-sheets to play with. Problem solvered.

src/main/resources/net/rptools/maptool/library/builtin/stat-sheet-i18n.properties
```diff
@@ -0,0 +1,192 @@
...
+label.pc = PC
+#The Token type, NPC
+label.npc = NPC
+#The Token property list
+label.propertyList = Properties
+#Token states
+label.states = States
+
+property.agility        = Agility
+property.alertness      = Alertness
```
cwisniew, line 23: From here downwards kind of misses the point of the statsheets, the stat sheet is to display what the user has selected, not arbitrary values. There is a display name for each property which can be used for i18n

bubblobill: @cwisniew  Kinda yes, aaaaand kinda no.
Imagine a world where I roleplay with people who have differing primary languages and the GM has just picked some random framework off a forum somewhere.
How gob-smacked would I be if his stat-sheets and character-sheets would be delivered in my native language, even though the stats were actually their standard English in the system?

cwisniew: > @cwisniew Kinda yes, aaaaand kinda no. Imagine a world where I roleplay with people who have differing primary languages and the GM has just picked some random framework off a forum somewhere. How gob-smacked would I be if his stat-sheets and character-sheets would be delivered in my native language, even though the stats were actually their standard English in the system?

The way you are doing it is definaltey no not kinda no :) 
What you really want to do is have the display name as either a name the user types in, or a key value they can pick from  (think similar to the IntelliJ forms designer) in the properties dialog. Then the display name would be populated with either the lookup of the key or the value, and the person writing the statsheet would have to do nothing extra. Maybe also start a discord thread to collect all the other names people want, I am sure there will be plenty:)

bubblobill:
Closed due to lack of interest

PR #5651, opened 2025-06-28
Fix for get macro location

### LINKED ISSUE #5638: [Bug]: 1.18 - getMacroLocation() on Lib:tokens not consistent
### Describe the Bug

When clicking a macro button on a Lib:token, `getMacroLocation()` returns only the part of the token name after the `Lib:`. When invoking the macro via `[macro("someMacro@Lib:token"): ""]`, `getMacroLocation()` returns the full token name.

On 1.17, `getMacroLocation()` always returned the full token name.

### To Reproduce

1. Add a token to the map and name it `Lib:Dragon`.
2. Add a `Testing` macro to `Lib:Dragon` with these contents:
    ```
    [h: broadcast(getMacroLocation())]
    ```
3. Add a `Test2` macro to `Lib:Dragon` with these contents:
    ```
    [h: broadcast(getMacroLocation())]
    [macro("Testing@TOKEN"): ""]
    [macro("Testing@this"): ""]
    ```
4. Click `Testing` and observe the following printed to chat:
    ```
    Dragon
    ```
5. Click `Test2` and observe the following printed to chat:
    ```
    Dragon
    Lib:Dragon
    Dragon
    ```

### Expected Behaviour

`getMacroLocation()` should return the full token name when the location is a Lib:token.

In particular, the behaviour of the given macros should match 1.17. I.e., clicking on `Testing` should result in:
```
Lib:Dragon
```
And clicking on `Test2` should result in:
```
Lib:Dragon
Lib:Dragon
Lib:Dragon
```

### Screenshots

_No response_

### MapTool Info

1.18.0-rc.1

### Desktop

Linux Mint 22.1

### Additional Context

`@token` has a similar issue with `getMacroLocation()`, and so does `@this`.

Locations like `@gm` also differ from 1.17, but only in case.

cwisniew:
resolves #5638

### Description of the Change
The `getMacroLocation()` method was exposing the internal workings of how the parser tracked context, so when these changed, it caused an error with the return value in some instances. I have added a new `getCallableLocation()` method to the parser macro source location that exposes these values in a way compatible with `getMacroLocation()`, while allowing the context handling in the parser to remain hidden from macros.

### Possible Drawbacks
Before the change in 1.18 `getMacroLocation()` would not produce consistent values for token macros. I have remedied this by returning a consistent value that matches the documented behaviour on the wiki. This may break some macros, but it's unlikely, and probably better than trying to emulate the broken behaviour for an edge case for this macro call.

### Documentation Notes
Fix for bug in `getMacroLocation()` that was introduced with the changes to macro function call routing.

### Release Notes
- N/A Fix for regression

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5651)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/MacroFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/MacroFunctions.java
@@ -186,8 +186,8 @@ public class MacroFunctions extends AbstractFunction {
       return MapTool.getParser().getMacroName();
 
     } else if (functionName.equalsIgnoreCase("getMacroLocation")) {
-      return MapTool.getParser().getMacroSource().getLocation();
-
+      var source = MapTool.getParser().getMacroSource();
+      return source.getCallableLocation();
     } else if (functionName.equalsIgnoreCase("setMacroCommand")) {
       FunctionUtil.checkNumberParam(functionName, parameters, 2, 4);
       FunctionUtil.blockUntrustedMacro(functionName);
--- a/src/main/java/net/rptools/maptool/client/macro/MacroLocation.java
+++ b/src/main/java/net/rptools/maptool/client/macro/MacroLocation.java
@@ -68,7 +68,7 @@ public class MacroLocation {
    * @param source the source of the macro.
    * @param location the location of the macro.
    * @param uri the URI of the macro, if applicable.
-   * @param Token the associated token, if applicable.
+   * @param token the associated token, if applicable.
    */
   MacroLocation(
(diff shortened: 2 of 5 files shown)


kwvanderlinde:
@cwisniew Do you mind targeting this against the `release-1.18` branch?

cwisniew:
> @cwisniew Do you mind targeting this against the `release-1.18` branch?

Doh!
Closed in favour of #5653, which has the changes against the correct branch

PR #5656, opened 2025-06-29
i18n of Basic Campaign token type, property and light names

bubblobill:
progresses #4600 
https://github.com/RPTools/maptool/issues/4600#issuecomment-3016871037

### Description of the Change
i18n of

-  Basic token type,
-  property names
-  light names
-  sight names
-  state names
-  bar name

### Possible Drawbacks
Might not have caught all the method calls.

### Documentation Notes
Localisation of default Campaign setup language.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5656)


bubblobill:
Tested this with a garbage string `©§¦¤µ ¿ÃÒÑÐýø÷` in various i18n strings and it worked.
As far as I can tell, changing the name does no harm to any of the internationalised features.

src/main/java/net/rptools/maptool/model/LightSource.java
```diff
@@ -315,6 +380,7 @@ public int compareTo(@Nonnull LightSource o) {
   public static @Nonnull LightSource fromDto(@Nonnull LightSourceDto dto) {
     return new LightSource(
         dto.hasName() ? dto.getName().getValue() : null,
+        dto.hasI18NKey() ? dto.getI18NKey().getValue() : null,
```
kwvanderlinde: Unless I'm missing something, we don't need the i18nKey in DTO since it's only real use is when reading / creating the default light sources.

kwvanderlinde:
FYI, I've added #5665 to move the default light sources into code. Whenever we do that, it will also remove the need for the I18N support in `LightSource` itself as we can properly customize the creation of these lights.

src/main/java/net/rptools/maptool/model/LightSource.java
```diff
@@ -283,7 +290,65 @@ public record LightArea(Light light, Area area) {}
...
+          nameOut += " - " + rangeOut.toString();
+        } else if (rangeOut != null) {
+          nameOut = rangeOut.toString();
+        } else {
+          nameOut = name;
+        }
+        listOut.add(
+            new LightSource(
+                nameOut,
+                ls.getI18nKey(),
```
kwvanderlinde: One thing to watch out for - even though we need to read i18nKey from XML because of how default light sources are loaded, we shouldn't write it out into campaign files. I would suggest passing `null` here so the I18N key isn't carried around after loading.

kwvanderlinde:
Sorry for the merge conflicts...

Now that light sources are defined in code, the translation can be done in `CampaignProperties` when defining each light and `LightSource` no longer needs to support i18n directly. E.g.,
```
categorizedLights.addAllToCategory(
        "D20",
        List.of(
            createD20LightSource(I!8N.getString("Default.campaign.lightSource.light.candle"), 5),
…
```

src/main/java/net/rptools/maptool/model/CampaignProperties.java
kwvanderlinde, line 1: Need to add the light source translations here.

src/main/resources/net/rptools/maptool/model/defaultLightSourcesMap.xml
kwvanderlinde, line 1: This file should be removed again as it is no longer used.

src/main/proto/data_transfer_objects.proto
```diff
@@ -73,6 +73,7 @@ message LightSourceDto {
   }
   repeated LightDto lights = 1;
   google.protobuf.StringValue name = 2;
+  google.protobuf.StringValue i18nKey = 7;
```
kwvanderlinde: No more need for adding i18n keys to DTO.

src/main/java/net/rptools/maptool/model/LightSource.java
```diff
@@ -296,7 +296,7 @@ public int compareTo(@Nonnull LightSource o) {
       int onameLong = NumberUtils.toInt(o.name, Integer.MIN_VALUE);
       if (nameLong != Integer.MIN_VALUE && onameLong != Integer.MIN_VALUE)
         return nameLong - onameLong;
-      return name.compareTo(o.name);
+      return Objects.requireNonNull(name).compareTo(Objects.requireNonNull(o.name));
```
kwvanderlinde: These `requireNonNull()` calls don't actually do anything and could be removed. With or without them, a NullPointerException will be thrown.

kwvanderlinde (changes requested):

src/main/java/net/rptools/maptool/model/LightSource.java
bubblobill: But IntelliJ tells me to...

src/main/java/net/rptools/maptool/model/CampaignProperties.java
bubblobill: done

src/main/proto/data_transfer_objects.proto
bubblobill: done

src/main/java/net/rptools/maptool/model/LightSource.java
bubblobill: changed

kwvanderlinde (approved):

Issue #3777, opened 2022-12-13
[Refactoring]: Split maptool into different projects in the same repository


thelsing:
### Describe the problem

Currently we only have on big project than contains nearly all code. I would like to split this up into several jars like maptool-server, maptool-client-swing, maptool-client-common, maptool, maptool-model, maptool-lib-swing and so on. 

We could also reintegrate dicelib into the main repository if we want.

### The improvement you'd like to see

As a first step I would move the whole code into a subdirectory as specified [here](https://docs.gradle.org/current/userguide/multi_project_builds.html).

### Expected Benefits

This will reduce comping time and makes it possible to add different clients while sharing much code. (We would need to separate actual gui from logic before that.

cwisniew:
Unless the code was in different repositories and then those jars were added in the build as the dependencies, this will increase the time taken for the build process not decrease it. Given the issues GitHub actions has been having for a while that's really undesirable.

thelsing:
I meant rebuild times as I expect gradle to ignore all subprojects if nothing changed. Have to test it it really does.

cwisniew:
our biggest problem at the moment though is github actions hanging, the longer it takes to compile from scratch the more pain involved.

thelsing:
This may be true, but I fail to see how is this relevant for this issue.
With recompiling I'm talking about during development on a local machine.

cwisniew:
Well the reason it is relevant is because doing it will have the above stated  adverse effects so the problems it creates would be bigger than the benefits it creates.

Phergus:
> our biggest problem at the moment though is github actions hanging, the longer it takes to compile from scratch the more pain involved.

Looks like it was fixed by [#3751](https://github.com/RPTools/maptool/pull/3751)

kwvanderlinde:
I personally don't care about the JARs per se, but splitting up into subprojects would certainly be useful. We have a lot of bad interdependencies right now, and subprojects would avoid the worst of it. Thinking of how the model depends on the UI, how the `clientserver` code depends on some MT specifics, how there's no real separation of client and server state, etc.

Easiest to start with would be pulling out the `clientserver` code and some common libs. The model will probably be the worst to do, but if we can wrangle that away from the UI bits than the server code will come out pretty easily too.

cwisniew:
[...]

PR #5677, opened 2025-07-08
Factor common utilities into a separate subproject

kwvanderlinde:

Progress #3777

### Description of the Change

This pulls out most library code from `net.rptools.lib` and puts it into a new `common` subproject. There's also a few things that I moved into `net.rptools.lib` that used to live under `net.rptools.maptool` but did not really have anything to do with MT per se.

Some utility classes were partially MT-dependent. For these I split the classes so that the MT-dependent part could stay while the MT-independent part could be moved over to the `common` subproject, e.g.:
- `AWTUtil` was split from `SwingUtil` (more could have been moved from `SwingUtil` to `AWTUtil`, but the bits done now are what was needed for some other common code to be moved)
- `OsDetection` was split from `AppUtil` and `AppActions`.
- `ImageSupport` was split from `ImageUtil`, with `ImageUtil` moving over.
- `FileUtil#getConfiguredXStream()` was moved into `PersistenceUtil`.

Another issue was the reliance on `AppPreferences`, particularly for `RenderQuality`. Now the `RenderQuality` enum lives on its own in `common`, and `RenderQuality` values are passed as parameters to those `ImageUtil` methods that need it. In a similar vein, the key files for `PublicPrivateKeyStore` are now injected through the constructor, which also means that we need to pass around real `PublicPrivateKeyStore` instances instead of conjuring them from nothing.

This is almost a pure refactoring. The one change in behaviour is a bugfix for `CellPoint#midZonePoint(Grid, CellPoint)`: it has been replaced with `Grid#midZonePoint(CellPoint, CellPoint)` which properly calculates midpoints on hex grids as well as square grids. This is only used for debugging movement, so most users won't notice a change.

### Possible Drawbacks

Shouldn't be any.

### Documentation Notes

N/A

### Release Notes

- Refactored common code into a subproject.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5677)


kwvanderlinde:
I've pulled out the dicelib change. This PR will now only be about the new `common` subproject.

cwisniew (approved):

PR #5708, opened 2025-07-20
1.18 - fix for lib: prefix handling

### LINKED ISSUE #5706: [Bug]: @this when defining UDF does not resolve correctly
### Describe the Bug

Using `defineFunction()` with `@this` on a Lib:token results in the function definition missing the `Lib:` prefix.

### To Reproduce

1. Start a new campaign.
2. Add a token named `Lib:Dragon`.
3. Add an `onCampaignLoad` macro with this line: `[h: defineFunction("example", "example@this")]`
4. Save and reload the campaign.
5. Attempt to call the function: `[r: example()]`.
6. See this error: 
    ```
    Must specify a location for the macro "example@Dragon" to be run from.
    Error trace : example@Dragon <<< chat
    ```
7. List the defined functions (`[r: getDefinedFunctions("json", true)]`) and find the `"example"` entry. Notice the location is `"example@Dragon"` instead of `"example@Lib:Dragon"`.

### Expected Behaviour

`@this` should result in the full name of the Lib:token being used in the function definition.

### Screenshots

_No response_

### MapTool Info

1.18.1

### Desktop

Linux Mint 22.1

### Additional Context

_No response_

kwvanderlinde:

Fixes #5706

### Description of the Change

A few places have been updated to use the callable location instead of the plain location from the `MacroLocation`:
- When resolving `@this` in `defineFunction()`
- In the tooltips for `/say` and `/gm`.
- In the `"source"` value returned by `getMacroContext()`

A few other places directly handle `lib:` prefixes but weren't correct, particularly for tokens with strange but valid names like `Lib:Lib:Hero`:
- `LibraryTokenManager#getLibrary()` once again always adds the `lib:` prefix rather than doing so conditionally. All callers were checked to make sure they pass the namespace, thereby removing the need for the check in the first place.
- `MapToolLineParser#runMacro()` does not remove the `lib:` prefix since it is guaranteed to have the namespace already from `MacroLocation#getLocation()`.
- Similarly, functions like `getLibProperty()` do not try to remove the `lib:` prefix from the result of `MacroLocation#getLocation()`, but they do still need to be able to remove it from a user-supplied lib:token name.

Similar to the `Lib:Lib:Hero` naming issue, macros named like `macro/example` on lib:tokens had the same sort of problem when trying to reference them via URIs with the `macro()` roll option, e.g., `[r, macro(lib://Dragon/macro/macro/example): ""]`. Add-on macros, on the other hand, simply could not be referenced via `macro()` since a full URI could not be resolved by add-ons. Both of these have been fixed. As part of this fix, URI `MacroLocation` and library `MacroLocation` need to be resolved differently, so there is a new method `Library#getMTScriptMacroInfoForUriPath(String)` that supports the URI case, while `Library#getMTScriptMacroInfo(String)` continues to support the library case (among other things).

### Possible Drawbacks

Should be none

### Documentation Notes

N/A

### Release Notes

- Fix various places to correctly handle the `lib:` prefix on lib:tokens.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5708)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
+++ b/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
@@ -1232,28 +1232,24 @@ public class MapToolLineParser {
     switch (mloc.getSource()) {
       case library, uri -> {
         try {
-          String namespace;
-          String macro;
-          switch (mloc.getSource()) {
-            case uri -> {
-              namespace = mloc.getUri().getHost();
-              macro = mloc.getUri().toString();
-            }
-            case library -> {
-              namespace = mloc.getLocation();
-              macro = mloc.getName();
-            }
-            case null, default -> {
-              throw new IllegalStateException("Unexpected value: " + mloc.getSource());
-            }
-          }
+          var namespace = mloc.getLocation();
           var lib = new LibraryManager().getLibrary(namespace);
           if (lib.isEmpty()) {
             throw new ParserException(
                 I18N.getText("lineParser.unknownLibToken", mloc.getLocation()));
           }
           var library = lib.get();
-          var macroInfo = library.getMTScriptMacroInfo(macro).get();
+
+          var macroInfoFuture =
+              switch (mloc.getSource()) {
+                case uri -> library.getMTScriptMacroInfoForUriPath(mloc.getName());
+                case library -> library.getMTScriptMacroInfo(mloc.getName());
+                case null, default -> {
+                  throw new IllegalStateException("Unexpected value: " + mloc.getSource());
+                }
+              };
+
+          var macroInfo = macroInfoFuture.get();
           if (macroInfo.isEmpty()) {
             // if the macro source is the same as the location then check private macros.
             if (!contextStackEmpty()) {
--- a/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
+++ b/src/main/java/net/rptools/maptool/client/MapToolLineParser.java
@@ -1240,7 +1240,7 @@ public class MapToolLineParser {
               macro = mloc.getUri().toString();
             }
             case library -> {
-              namespace = mloc.getLocation().replaceFirst("^(?i)lib:", "");
+              namespace = mloc.getLocation();
               macro = mloc.getName();
             }
             case null, default -> {
--- a/src/main/java/net/rptools/maptool/client/functions/DefineMacroFunction.java
+++ b/src/main/java/net/rptools/maptool/client/functions/DefineMacroFunction.java
@@ -60,7 +60,7 @@ public class DefineMacroFunction extends AbstractFunction {
       if (macro.toLowerCase().endsWith("@this")) {
         macro =
             macro.substring(0, macro.length() - 4)
-                + MapTool.getParser().getMacroSource().getLocation();
+                + MapTool.getParser().getMacroSource().getCallableLocation();
       }
 
       boolean ignoreOutput = false;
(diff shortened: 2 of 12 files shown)


src/main/java/net/rptools/maptool/client/MapToolLineParser.java
```diff
@@ -1240,7 +1240,7 @@ public String runMacro(
               macro = mloc.getUri().toString();
             }
             case library -> {
-              namespace = mloc.getLocation().replaceFirst("^(?i)lib:", "");
```
cwisniew, line 1243: Does fetch from a macro and static file work after this change? It may break it

kwvanderlinde: I will give it this a test to be sure.

It _shouldn't_ break since `getLocation()` should already have the `lib:` stripped off.

kwvanderlinde: I've validated that `fetch()` still works with both `lib://` URIs and `macro:` URIs, and that `macro()` still works with `lib://` URIs.

I'll try to get a test for static files, but I don't do a whole lot with add-ons yet so that might take me a bit.

kwvanderlinde: Okay, I've now validated that fetch() works for both lib:tokens and add-on libraries, and checked that it can both call macros and retrieve static files.

For lib:tokens, this kind of call also works to invoke the `doFetch@Lib:Dragon`:
```
[h, macro("lib://Dragon/macro/doFetch"): ""]
```
For add-on libraries, I assume this is the correct syntax:
```
[h, macro("lib://Dragon/doFetch"): ""]
```
But it just errors out with "Unknown macro "doFetch"." I don't know yet if this means something is broken right now, or if I'm just doing something wrong with add-on - the latter comes with a high probability.

kwvanderlinde: Figured it out - the URI case is not correct. Will push a fix for that as well later today.

kwvanderlinde: To expand on the current problem: URIs cannot be unambiguously resolved to a library namespace and macro name without first knowing whether the referenced library is a token or an add-on. This is because the string `lib://Dragon/macro/example` could either reference:
- the `macro/example` macro in the `Dragon` add-on; or
- the `example` macro in the `Lib:Dragon` token.

Moreover `Library#getMTScriptMacroInfo()` also has completely different expectations depending on the implementation:
- Add-ons expect only the macro name proper to be provided, so that `public/` can be prepended to it and be resolved to a macro file name.
- Tokens need the whole URI so that the `macro/` part can be stripped off.

On top of that, the way `LibraryToken#getMTScriptMacroInfo()` parses the URI means macros named like `macro/example` cannot be called using the expected URI `lib://Dragon/macro/macro/example`.

^ All that is to explain the need for my last commit here. With these fixes, all my `lib:` URIs in different scenarios are now working.

cwisniew: > On top of that, the way `LibraryToken#getMTScriptMacroInfo()` parses the URI means macros named like `macro/example` cannot be called using the expected URI `lib://Dragon/macro/macro/example`.
> 
Yeah, because a macro and a property on a lib: token can have the same name it requires the macro/ and property/ "directory" in the URL to properly distinguish :(

cwisniew (approved):

PR #5727, opened 2025-07-24
Preferences Revamp

### LINKED ISSUE #4987: [Feature]: Reorganize the Preferences dialog
### Describe the Problem

I can never find anything in the Preferences dialog without 10 straight minutes of staring at my monitor. Especially for anything in the _Interactions_ or _Application_ tabs - I almost never guess which tab is right, and even if I do there's so much in both of them that it's overwhelming. The other tabs aren't bad though.

### The Solution you'd like

For preferences currently in the _Interactions_ and _Application_ tabs, move them into new tabs based on what they affect or relate to. The main goal here is any settings a user would naturally relate to each other should also be located by each other. I expect the _Interactions_ and _Applications_ tabs would no longer need to exist afterward.

I think these are a good set of new tabs to consider:
- Map: for map defaults and rendering.
- Token: for token defaults, behaviours.
- Chat: for any chat-related settings.
- Server: for personal server settings and other server setting if there are any.

None.

### Additional Context

As an example, map defaults like **New Maps have Fog of War** are in the _Interactions_ tab, but it doesn't really relate to an interaction (that is decided by actual map settings). Over in the _Application_ tab we have **Hide 'Map not visible to players' Warning** - alright that's certainly application behaviour, but why do I have to go somewhere else to modify map behaviours? If I want to change something about maps, I should be able to find a "maps" thingy somewhere and tweak it.

bubblobill:
resolves #4987 

### Description of the Change
Generates the Preferences dialog rather than using a form.
Based on the preference type, a label and control are generated and added to the preferences presented to the user.

Includes a search box for finding what you want quickly.

Adding a simple user preference to the form requires adding it to an enum of user preferences and including which section(think "tab") and group (e.g. token/map/etc.) it belongs to, and whether or not to use auto-generated controls.

Currently toying with different presentations. Present iteration is a Tree and a JideGroupTable.

### Possible Drawbacks
Too much hard-coding still. Will probably replace the enum with a record register. Keeping backwards compatibility currently involves some silliness.

### Documentation Notes
Reworked User Preferences

### Release Notes

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5727)


bubblobill:
@kwvanderlinde @cwisniew 
This is at alpha stage. Please have a gander and let fly with the critiques

PR #5740, opened 2025-07-31
Minor tweak to facing arrow

### LINKED ISSUE #5622: [Feature]: Facing arrow obscured
### Describe the Problem

Not really a bug, just inconvenient.
Bars are drawn over the facing arrow.

Image

Image

### The Solution you'd like

Arrow over bar

_No response_

### Additional Context

_No response_

### LINKED ISSUE #5741: [Feature]: make Facing Arrow colour settable
### Describe the Problem

Hard-coded arrow colour does not suit all cases.
[see discord](https://discord.com/channels/296230822262865920/1400568171165122582)

Image

### The Solution you'd like

User preference for arrow colour

_No response_

### Additional Context

_No response_

bubblobill:
progresses #3691, #3777
closes #5622
closes #5741 

### Description of the Change
Moved utility class to Common project.
Moved arrow renderer in the render order.
Added arrow colour preferences.
Made facing arrow prettier with a slight dovetail
image

### Possible Drawbacks
Utility harder to find

### Documentation Notes
n/a

### Release Notes

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5740)


src/main/java/net/rptools/maptool/client/ui/zone/renderer/tokenRender/FacingArrowRenderer.java
```diff
@@ -30,26 +32,37 @@
...
 public class FacingArrowRenderer {
   private static final Logger log = LogManager.getLogger(FacingArrowRenderer.class);
+  private static final double tailX = -0.25;
+  private static final double dovetailX = -0.15;
+  private static final double tailY = .35;
+  public static final Path2D DOVETAIL =
+      SVGUtils.svgToPath2D(
+          String.format("M 0,0 L %1$f2 -%2$f2 L %3$f2 0 L %1$f2 %2$f2 z", tailX, tailY, dovetailX));
+  public static final Path2D TRIANGLE =
+      SVGUtils.svgToPath2D(String.format("M 0,0 L %1$f2 -%2$f2 L %1$f2 %2$f2 z", tailX, tailY));
```
kwvanderlinde: Why the switch to SVG here. It seems to do essentially the same the existing `Path2D` construction, but with the possibility of failure.

bubblobill: 1. It justifies my svgToPaths2D method.
2. I wanted to keep the triangle for posterity, and I can't just comment it out and satisfy Spotless.
3. It opens the door to setting the facing arrow with a string programmatically.
4. Which means it is a good idea to have a try/catch, even though this will never happen in the current code.
5. It makes it easier for me to design arrows in a vector drawing program where I can just copy and paste the path which may well fail to parse.
6. I can potentially build arrows on-the-fly instead of using the arrow transform.

cwisniew: > 3. It opens the door to setting the facing 

> 5. It makes it easier for me to design arrows in a vector drawing program where I can just copy and paste the path which may well fail to parse.


So you have heard me say it before, and you will hear me say it a lot more again. Any functionality like this that is not purely macro dependant should be implemented via the UI before macros. In this case loading a SVG file rather than providing a SVG string in a macro. Or if the string you are talking about is hard coded then refer to previous conversation about loading the SVG from a resource rather than hard coding the string. :)

bubblobill: These two things are identical.
```
static {
    DOVETAIL = new Path2D.Double();
    DOVETAIL.moveTo(0, 0);
    DOVETAIL.lineTo(tailX, -tailY);
    DOVETAIL.lineTo(dovetailX, 0);
    DOVETAIL.lineTo(tailX, tailY);
    DOVETAIL.closePath();
  }
```
```
public static final Path2D DOVETAIL =
      SVGUtils.svgToPath2D(
          String.format("M 0,0 L %1$f2 -%2$f2 L %3$f2 0 L %1$f2 %2$f2 z", tailX, tailY, dovetailX));
```
1. They're both hard-coded. 
2. If I build them dynamically, they are still hard-coded.
3. Adding the capability to alter them by macro code is a long way short of adding the facility to alter them by macro code.

cwisniew: > These two things are identical.
> ```
> static {
>     DOVETAIL = new Path2D.Double();
>     DOVETAIL.moveTo(0, 0);
>     DOVETAIL.lineTo(tailX, -tailY);
>     DOVETAIL.lineTo(dovetailX, 0);
>     DOVETAIL.lineTo(tailX, tailY);
>     DOVETAIL.closePath();
>   }
> ```
> ```
> public static final Path2D DOVETAIL =
>       SVGUtils.svgToPath2D(
>           String.format("M 0,0 L %1$f2 -%2$f2 L %3$f2 0 L %1$f2 %2$f2 z", tailX, tailY, dovetailX));
> ```
> 1. They're both hard-coded. 
> 2. If I build them dynamically, they are still hard-coded.
> 3. Adding the capability to alter them by macro code is a long way short of adding the facility to alter them by macro code.
> 

If you are going to change it to a SVG string you might as well make the file a resource and do it the correct way. It's easier/better to replace a file then change a hard coded string. The difference in the two things you have above is SVG was designed as a file format where as java 2d was not.

bubblobill: Made it boring again

kwvanderlinde (approved):

Issue #5594, opened 2025-06-17
[Maintenance]: Consolidate token decorations into a package


bubblobill:
### Describe the problem

States/Bars, Halos, and Auras - i.e. Token Decorations; are all graphical representations of token properties, with common (duplicated) functionality and code. 
Their behaviour is occasionally inconsistent between grid types.
They are often generated on-the-fly instead of being cached for rendering.

### The improvement you'd like to see

Consolidate all decorations into one package with a common superclass/interface, renderer, and possibly a manager.

### Expected Benefits

Improved code managability.
Faster rendering.
Easier implementation of changes/improvements/new features.
Could offload related Token tracking properties to a central manager.
Could allow for creation and destruction of definitions outside of the campaign properties with macro functions.

### Additional Context

#5296  Token Split / Refactoring

Issue #5586, opened 2025-06-12
[Bug]: Token opacity applied to bars only when a state is set.


kwvanderlinde:
### Describe the Bug

A token with a reduced opacity will normally show any bars at the bar's configured opacity. But as soon as a state is set on the token, the bar also incorporates the token opacity.

### To Reproduce

1. Start a new campaign.
2. Add a token to the map.
3. Set the token's opacity to 50%.
4. Give the token the default "Health" bar. Notice the bar is opaque.
5. Give the token the "Dead" state. Notice the bar is no longer opaque.

### Expected Behaviour

Bars are always shown at their configured opacity regardless of the token's opacity.

OR

Bars always take the token's opacity into account.

### MapTool Info

1.16.4, 1.17.0

### Desktop

Linux Mint 22.1

Issue #4726, opened 2024-03-20
[Bug]: Token Opacity does not affect the token's States.


FullBleed:
### Describe the Bug

States currently ignore token opacity.  For example, a token at 0% opacity (not visible) will have fully visible states "floating" on the map.

### To Reproduce

Add an image state to a token and then reduce the token's opacity via Edit Token>Config.  The state's opacity does not change.

### Expected Behaviour

I'd expect token states would mirror the opacity level of the token.

### MapTool Info

1.15.0-rc2

FullBleed:
I doubt that this is a version specific "bug" and suspect it is how token opacity level has functioned since it's inception.  So it's more of an oversight in the implementation that has just never been called out.

I recognize two things:

1)  States have their own default transparency level.  I think that their adjusted levels should be multiplicative.
2)  Since this behaviour has been around awhile there is a chance that someone is using it as is.  So having an option in States to "Ignore Token Opacity" might be a nice feature.

bubblobill:
This likely applies to halos as well, being another feature that is visible to all.

FullBleed:
> This likely applies to halos as well, being another feature that is visible to all.

Confirmed.

Also noted that when applying a halo through the right-click dropdown, selecting a checked halo does not toggle it off.  That's atypical.

FullBleed:
I've got an update for this bug...

If the state is set to 100% opacity, it WILL heed the opacity of the token.

If it's set to anything below 100%, it will not heed to token's opacity.

PR #5742, opened 2025-08-02
Split more rendering aspects from ZoneRenderer

### LINKED ISSUE #5622: [Feature]: Facing arrow obscured

bubblobill:
You can link a pull request to an issue with one of
fixes #5622 #5746
progresses #5594 #5586 #4726 #4944 #3691 #2750 #2346 #2328 #2000

until I got sick of looking for affected issues

## LabelRenderer
 - Moved caches to class, added setters and static isLabelVisible method.
 - Moved token label generation in token rendering sequence preparatory to moving it out to DecorationRenderer.
## New class TokenDecorationRenderer
 - Paints things before and after token rendering.
 - Hub for halos, bars, states, etc. More bits to follow.
## New class OverlayRenderer
 - For painting states and bars.
## FacingArrowRenderer
 - Facing arrow now follows grid cell shape. Made some variables non-local.
## GeometryUtil 
 - Added functions for finding intersection points.
## HaloRenderer
 - Now honours preference for opacity.
## TokenRenderer
 - Updated to call decoration renderer before and after painting image. plus other tweaks.
## ZoneViewModel
 - Added method- isUsingVision().
## ZoneRenderer
 - Inverted hasMoveSelectionSetMoved(). 
 - Made imageLabelFactory a constant. 
 - LabelRenderingCache offloaded to LabelRenderer.
 - Renamed renderLabels() method as it is confusing. 
 - Removed redundant image loading and token flipping from renderTokens().
 - Other  housework.

### Possible Drawbacks
I'm not very good at my job

### Documentation Notes
Separated most token decoration rendering out of ZoneRenderer

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5742)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/renderer/HaloRenderer.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/renderer/HaloRenderer.java
@@ -30,6 +30,13 @@ import net.rptools.maptool.model.zones.GridChanged;
 public class HaloRenderer {
   private final RenderHelper renderHelper;
   private final Zone zone;
+  private float opacity = AppPreferences.haloOverlayOpacity.get() / 255f;
+  private float lineWeight = AppPreferences.haloLineWidth.get();
+
+  {
+    AppPreferences.haloOverlayOpacity.onChange(i -> opacity = i / 255f);
+    AppPreferences.haloLineWidth.onChange(i -> lineWeight = i);
+  }
 
   // region These fields need to be recalculated whenever the grid changes.
 
@@ -69,16 +76,15 @@ public class HaloRenderer {
                 .createTransformedShape(cachedHaloShape);
       }
     }
-
     return cachedHaloShape;
   }
 
   // Render Halos
-  public void renderHalo(Graphics2D g2d, Token token, ZoneViewModel.TokenPosition position) {
+  public void renderHalo(Graphics2D g2d, ZoneViewModel.TokenPosition position) {
+    Token token = position.token();
     if (token.getHaloColor() == null) {
       return;
     }
-
     var grid = zone.getGrid();
     if (grid == null) {
       return;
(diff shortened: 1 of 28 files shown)


src/main/java/net/rptools/maptool/client/ui/zone/renderer/HaloRenderer.java
```diff
@@ -30,6 +30,13 @@
 public class HaloRenderer {
   private final RenderHelper renderHelper;
   private final Zone zone;
+  private float opacity = AppPreferences.haloOverlayOpacity.get() / 255f;
+  private float lineWeight = AppPreferences.haloLineWidth.get();
+
+  {
+    AppPreferences.haloOverlayOpacity.onChange(i -> opacity = i / 255f);
```
Baaaaaz, line 37: I do not think AppPreferences.haloOverlayOpacity is intended to be used in this way as the opacity for all rendered halos, but was intended just for a token's vision when using the halo color (as per the logic in VisionOverlayRenderer.java).  Given the current default preference value of 50, this code would nerf everyone's halos!  

Maybe the existing AppPreferences.haloOverlayOpacity display text should be changed for clarity (even though it sits under the Map Visuals tab).  

In light of my halo feature PR, maybe there should instead be a new campaign setting or preference for global halo opacity (and/or configured for halos a la lights and auras)?

bubblobill: Then they should not have called it "halo overlay opacity", and I agree the default is stoopid.

If only there was some way to manage halos in a fine-grained manner that permitted individual variation from the global setting. :P

src/main/java/net/rptools/maptool/client/ui/zone/renderer/tokenRender/FacingArrowRenderer.java
```diff
@@ -45,11 +48,29 @@ public class FacingArrowRenderer {
...
   private final Color fillColour = Color.YELLOW;
   private final Color borderColour = Color.DARK_GRAY;
+  private boolean isIsometric;
+  private boolean isSquare;
+
+  @SuppressWarnings("unused")
+  @Subscribe
+  private void onGridChanged(GridChanged event) {
+    if (event.zone() != null) {
+      this.zone = event.zone();
```
kwvanderlinde, line 64: You don't want to change the zone here since each `FacingArrowRenderer` is specific to a single zone (as each `TokenRenderer` / `ZoneRenderer` is). Instead, check that the zone is the same, and ignore the event if it's not.

This event handling is also unnecessary. `isSquare` isn't used and should be removed, and `isIsometric` is so cheap it's not worth managing state just to get it.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/LabelRenderer.java
```diff
@@ -16,18 +16,34 @@
...
+
+  public static Map<GUID, LabelRenderer> getLabelCache() {
+    return labelCache;
+  }
+
+  public static Map<GUID, BufferedImage> getLabelImageCache() {
+    return labelImageCache;
+  }
 
   private final ZoneRenderer renderer;
```
kwvanderlinde, line 43: Can remove `renderer` as it's no longer used.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/LabelRenderer.java
```diff
@@ -16,18 +16,34 @@
...
+import net.rptools.maptool.client.ui.zone.ZoneViewModel;
 import net.rptools.maptool.model.GUID;
 import net.rptools.maptool.util.GraphicsUtil;
 
 /** Represents a delayed label render */
-class LabelRenderer implements ItemRenderer {
+public class LabelRenderer implements ItemRenderer {
+
+  private static final Map<GUID, LabelRenderer> labelCache = new HashMap<>();
+  private static final Map<GUID, BufferedImage> labelImageCache = new HashMap<>();
```
kwvanderlinde, line 33: The `labelImageCache` is a memory leak and needs to be changed. There needs to be some way to allow old entries to be garbage collected. The easy way to do this is to not make these `static` so that they can at least be collected when the zone is removed. Or you need a way to explicitly remove old entries. Compare with the original `ZoneRenderer.labelRenderingCache` which did both.

`labelCache` has similar issues. Although entries can be removed, it's only done as part of the render loop. So there's no way for entries from a removed zone to be collected.

src/main/java/net/rptools/maptool/client/ui/zone/renderer/HaloRenderer.java
kwvanderlinde: Just to echo what Baaaaaz said, don't change the behaviour of the preference in this way. Fix the wording if you don't like it.
This opacity setting should only apply to the vision overlay, not the halo around the cell. The two have nothing to do with each other, and this will only cause confusion.

kwvanderlinde (changes requested):

PR #5788, opened 2025-09-17
Add "no statsheet" option

### LINKED ISSUE #5643: [Feature]: Set a token to have no stat-sheet
### Describe the Problem

I often have lib:/utility tokens that need no stat-sheet cluttering up my view with stat-sheets as I am developing things. 

### The Solution you'd like

I would like to be able to set the tokens to have no stat-sheet.

Have an empty stat-sheet as an assignable option.

### Additional Context

_No response_

cwisniew:
#5643

### Description of the Change
Added a No StatSheet option and cleaned up the stat-sheet location combo box on the campaign properties and edit toke dialogs so that it is disabled when the location can't be set.

### Possible Drawbacks

Should be none

### Documentation Notes
Its now possible to set the stat sheet type as "No Stat Sheet"

### Release Notes
Its now possible to set the stat sheet type as "No Stat Sheet"

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5788)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/campaignproperties/TokenPropertiesManagementPanel.java
+++ b/src/main/java/net/rptools/maptool/client/ui/campaignproperties/TokenPropertiesManagementPanel.java
@@ -475,9 +475,7 @@ public class TokenPropertiesManagementPanel extends AbeillePanel<CampaignPropert
     var combo = getStatSheetComboBox();
     combo.removeAllItems();
     var ssManager = new StatSheetManager();
-    ssManager.getStatSheets(propertyType).stream()
-        .sorted(Comparator.comparing(StatSheet::description))
-        .forEach(ss -> combo.addItem(ss));
+    ssManager.getOrderedStatSheets(propertyType).forEach(ss -> combo.addItem(ss));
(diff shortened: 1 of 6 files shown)


kwvanderlinde:
The *Stat Sheet Location* field in the **New Token** dialog should also be disabled when set to "No Stat Sheet":
image

I also think it would make sense if the "No Stat Sheet" option would be presented first in the various combo boxes rather than being sorted alphabetically with the rest.

cwisniew:
> The _Stat Sheet Location_ field in the **New Token** dialog should also be disabled when set to "No Stat Sheet": image
> 
> I also think it would make sense if the "No Stat Sheet" option would be presented first in the various combo boxes rather than being sorted alphabetically with the rest.

Doh, had the preference to display that dialog off so I missed it, brain is getting really rusty, thanks for picking it up.

I have updated that dialog and changed all 3 places (new token dialog, campaign properties, and edit token dialog) so that the stat sheets are ordered in such a way that all internal stat sheets (default, legacy, no) appear in the list before non internal stat sheets.

kwvanderlinde (approved):

PR #5807, opened 2025-10-06
Fix handlebars url fetch

### LINKED ISSUE #5806: [Bug]: Loading Handlebars Partial templates by URL is broken
### Describe the Bug

The Java bridge glue is injected at URL fetch time to support injecting the glue into navigated-to pages.
The detection of which fetches to inject the glue into is based on file mime type detection, but handlebars doesn't have a registered mime type and get returned as HTML, so has glue injected that are invalid handlebars.

### To Reproduce

Use the AddOn in https://gitlab.com/richardmaw/maptool-bar-test/ with the properties and set a token to the Bars stat sheet.

Observe no stat sheet is rendered and the console logs an exception.

### Expected Behaviour

Stat sheet renders.

### Screenshots

_No response_

### MapTool Info

develop

### Desktop

_No response_

### Additional Context

Likely broken since https://github.com/RPTools/maptool/pull/5461, https://github.com/RPTools/maptool/pull/5489 fixed other types of file.

fishface60:

Fixes https://github.com/RPTools/maptool/issues/5806

### Description of the Change

This passes the file path to file type deduction in more places

### Possible Drawbacks

There may be valid handlebars partial templates that don't appear to be html and wouldn't get returned as Type.HANDLEBARS.

### Release Notes

* Fixed handlebars partial template loading for stat sheets and potential fetching of .hbs files from WebViews.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5807)

### DIFF
--- a/src/main/java/net/rptools/maptool/util/HandlebarsUtil.java
+++ b/src/main/java/net/rptools/maptool/util/HandlebarsUtil.java
@@ -52,6 +52,12 @@ import org.apache.logging.log4j.Logger;
  * @param <T> The type of the bean to apply the template to.
  */
 public class HandlebarsUtil<T> {
+  public static boolean isAssetFileHandlebars(String filename) {
+    if (filename == null) {
+      return false;
+    }
+    return filename.toLowerCase().endsWith(".hbs");
+  }
 
   /** The compiled template. */
   private final Template template;
--- a/src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLContent.java
+++ b/src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLContent.java
@@ -313,23 +313,23 @@ public class HTMLContent {
     if (!(content instanceof UrlContent urlContent)) {
       throw new IllegalStateException("HTMLContent is not a URL");
     }
+    var url = urlContent.url();
 
     try {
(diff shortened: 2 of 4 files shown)


src/main/java/net/rptools/maptool/util/HandlebarsUtil.java
```diff
@@ -45,6 +45,12 @@
  * @param <T> The type of the bean to apply the template to.
  */
 public class HandlebarsUtil<T> {
+  public static boolean isAssetFileHandlebars(String filename) {
+    if (filename == null) {
+      return false;
+    }
+    return filename.endsWith(".hbs");
+  }
```
cwisniew, line 60: This should be case-insensitive

cwisniew (changes requested):

src/main/java/net/rptools/maptool/util/HandlebarsUtil.java
fishface60: Thanks, amended.

cwisniew: Thanks!

cwisniew (approved):

PR #5809, opened 2025-10-07
Fixing reseting of bar when visibility is toggled.

### LINKED ISSUE #2326: Toggling bar visibility resets the bar
**Describe the bug**
When using the bars in tokens, if you hide the bar it resets the value to 100%. I assume this is undesirable, as the visibility of the bar should be unrelated to its value.

**To Reproduce**
1. Double click a token to open the edit window and go to State
2. At the bottom of the page choose any bar (e.g. "Health") and lower its value below 100% (or pick one that is already below 100%), then click the "Hide"checkbox.
3. Click "OK" to confirm the edit.
4. Open the token editor once again and the bar will have been reset to 100%.

The same happens when using the `setBarVisible()` function.

**Expected behavior**
I would expect the bar to become invisible (which it does), but the value of the bar not be altered. If you are, for instance, toggling show/hide of a health bar using `setBarVisible("Health", 1-isBarVisible("Health"))` every time the bar pops back up it's at 100%. The current workaround is to call the macro that sets the bar's value in the first place, but it doesn't seem like that would be the expected behavior.

**Screenshots**
N/A

**MapTool Info**
- Version: 1.7.0?
- Install: New

 - OS: Windows
 - Version 10

**Additional context**
N/A

svenviktorm:

resolves #2326 

### Description of the Change

The stored value of Token bars is separated into its shown value and a hidden value which is restored when the bar is set to visible. This is done by adding an additional value in the Token state map with the original bar name with a suffix added currently "StoredValue". Also a check has been added to when bars/states are added/updated in campaign properties so the names do not clash with the new hidden value.

The bar sliders in the Token popup editor and Token editor have been updated to handle the hidden values, I have not made it so that you can edit them while hidden is checked.

Example:
bar "Health" has value 0.8,
setBarVisible("Health", 0) is called upon the Token which sets "HealthStoredValue" to 0.8 and "Health" to null inside the Token status map making the health bar invisible.
setBar("Health", 0.3) is called setting "HealthStoredValue" to 0.3 not revealing the bar
setBarVisible("Health", 1) is called, setting "Health" to 0.3 making the bar visible
setBar("Health",0.9) is called setting both "Health" and "HealthStoredValue" to 0.9 updating the bar

The reason I have implemented this change in the rather crude way of having two instances of each bar stored in the states map is because then nothing has to be changed in how Token behaves and therefore TokenDTO and a lot more does not need to be changed. If the suffix "StoredValue" needs to be switched out it can be changed in TokenBarFunction.java.

### Possible Drawbacks
If a state or bar has the name of another state or bar (not two bars) but with the hidden value suffix (currently "StoredValue") that may cause them to be entangled, having one state / bar affecting the other. This should only be possible if an old Token is imported with a clashing bar or state name since it should not be possible to add a clashing state / bar.

Another possible side-effect is from any part of the program which use the keys from the State in Token.Java or getStatePropertyNames() or similar, now that part of the code will get duplicate results of every bar.

Backwards compatibility will be affected since setBarVisible(), setBar() and getBar() all have changed functionality.
setBar() can't be used to set a bar as visible,
getBar() can't be used to figure out if a bar is visible
and setBarVisible() don't reset/remove the bar's value.

### Documentation Notes
The wiki page about setBar and setBarVisible needs to be changed if this request is accepted, i don't know how to do that myself. getBar does not need to be updated.

### Release Notes

- Fixed an issue where a change in visibility of a bar reset the value.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5809)


cwisniew:
> Backwards compatibility will be affected since setBarVisible(), setBar() and getBar() all have changed functionality. setBar() can't be used to set a bar as visible, getBar() can't be used to figure out if a bar is visible and setBarVisible() don't reset/remove the bar's value.

The general rule is don't break macro functions compatability without a VERY good reason, so this change would need to be adjusted to keep the existing behaviour for the macro functions

src/main/java/net/rptools/maptool/client/functions/TokenBarFunction.java
```diff
@@ -38,6 +38,9 @@ private TokenBarFunction() {
   /** singleton instance of this function */
   private static final TokenBarFunction instance = new TokenBarFunction();
 
+  /** Suffix added to the bar name to store the hidden value of the bar */
+  public static final String hidenBarSuffix = "StoredValue";
```
cwisniew, line 42: Typo (should be hiddenBarSuffix)

src/main/java/net/rptools/maptool/client/functions/TokenBarFunction.java
```diff
@@ -116,7 +133,34 @@ public static BigDecimal isVisible(Token token, String bar) {
...
   public static BigDecimal setVisible(Token token, String bar, boolean show) {
-    MapTool.serverCommand().updateTokenProperty(token, Token.Update.setState, bar, show);
+    Object valueShown = token.getState(bar);
+    Object valueHidden = token.getState(bar + hidenBarSuffix);
+    if (show) {
+      if (valueHidden != null) {
+        // if we want to show the bar, and we have a hidden value, restore it
+        MapTool.serverCommand()
+            .updateTokenProperty(
+                token, Token.Update.setState, bar, getBigDecimalValue(valueHidden));
```
cwisniew, line 143: Need to handle the case where setVisible(..., true) is called an a bar that is already visible (i.e. it shouldn't copy the hidden value over the visible value). Also the opposite case where setVisible(..., false) is called on a bar that is already hidden, you don't want to copy the visible state (null) over the hidden state.

cwisniew (changes requested):

PR #5824, opened 2025-10-19
New Custom Halos 

### LINKED ISSUE #5786: [Feature]: More token halo shapes and styles.
### Describe the Problem

Different halos on a map can be difficult to differentiate by those with color blindness as they are limited to just being a colored grid shaped line (or circle shaped on gridless maps).

Examples of things I would also like to do with halos but cannot are:
-  to have a circle shaped halo around a circle shaped token on a square gridded map 
- a 'true' halo that envelopes a top-down token
- dotted and dashed halos
- visually link tokens together using something other than color during a game

### The Solution you'd like

Increase the accessibility and utility and of halos by providing more visual encoding channel(s) beyond color, such as a selection of different token halo shapes and line styles to pick from and apply to individual tokens.

None.

### Additional Context

None

Baaaaaz:
closes #5786

### Description of the Change
Lots!

New halo customization options, configurable under a new tab in Campaign Properties including:
* custom halo categories
* halo shape types:
  * basic geometric shapes such as `CIRCLE`, `ARC`, `CHORD`, `PIE`, `TRIANGLE`, `SQUARE`
  * regular `POLYGON` and `STAR` polygon shapes with configurable vertices
  * derived shapes such as `GRID`, `TOKEN`, `FOOTPRINT`, `TOPOLOGY`, `MBL`
  * mini-shapes versions of basic shapes and polygon shapes which revolve around the token
* halo can be made visible to `GM` or `OWNER` only
* transformations options such as `FLIPH`, `FLIV`, `ROTATE`, `FILL`, `SCALE` (with token), `scaleX=`, `scaleY=`, `angle=` (for arcs, chords, pies), `rotate=`
* additional mini-shape transformation options
* halo line width configurable (default to App Preferences line width)
* halo colors configurable (with or without transparency #RRGGBB/#RRGGBBAA)
* line dash patterns
* multiple halo segments definitions per halo
* multiple custom halos per token
* also fixed other campaign property expandable help section dividers

Notes:
* Mini shapes are filled and do not use strokes, normal halo shape types use stroke (though can be optionally filled)
* Original 'simple' halos are still supported and are rendered as the outmost halo if a token has multiple halos.  They still can determine the vision overlay color and is the only halo shown in initiative window.
* As per the original halos, strokes get clipped unless they are filled.  Arc halo shapes get clipped by the arc's circle not the arc itself.
* Note that halos will only be rendered if the token is visible and on screen.
* Performance times and counters added for the different halo render groups/halo shape types.

PopupMenu:
Categorized halos are available via the token popup menu.
* Elevated previous halo popup menu from TokenPopupmenu to AbstractPopupmenu and included within new Halos menu items
* Halos menu item now available to GMs not on the token layer which addresses inability to set halos on non-token layer but which you could do via MTScript

AppMenuBar:
Given potential for a lot of halos to be displayed, client option to toggle showing halos.  Also fixed highlighting toggle display of Show Token Names.

MTScript Functions:
* `getHalos`, `setHalos`, `clearHalos`, `hasHalos` deal with new custom halos
* `getHalo`, `setHalo` remain for original halos
* `showHalos` affects all client halos (as per new AppMenu bar item)

HexGrid:
`createCellShape()` now contains `edgeLengthAreaFudge` adjustment to increase the size of the individual cell area in one dimension so when compositing cells in a footprint they now touch/overlap and do not create subpaths, which otherwise would render as small lines withing a `FOOTPRINT` HaloShapeType.  There may be a better fix for this.

Heavily borrowed from the following Light classes in the code base:
- Light -> Halo
- Lights -> Halos
- CategorizedLights -> CategorizedHalos
- LightSources -> HaloSources (not too happy with this name, but made mental jump from lights to new halos easier)
- AttachedLights -> AttachedHalos
- TokenLightFunctions -> added to TokenHaloFunction and class renamed TokenHaloFunctions
- getInfoFunction -> added halos to campaignInfo

### Possible Drawbacks
Legacy halos still work and combined with the new custom halos.  Need help on deprecating those.
Maybe better ways of doing things, I'm a java novice.

### Documentation Notes

To be documented here: https://wiki.rptools.info/index.php/Introduction_to_Halos and wiki pages for the new macro functions.

Examples of some new custom halos (note tokens set at 50% opacity for clarity).

Square grid basics:
image

some options:
image

mini-shapes (and options):
image

multiple halos and current campaign property defaults:
image

some more examples:
image

basics on a hex map:
image

basics on an iso map:
image
 
new token menu items:
image

new app menu item "Show Token Halos":
image

new Campaign properties tab:
image

### Release Notes
New configurable token halos in Campaign Properties

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5824)


Baaaaaz:
Reedy for a review...

kwvanderlinde:
Thanks for the thorough description, it's going to make it much easier to grok what's going on.

I'll have a proper look at this soon. For now, I've got a couple observations after a really quick reading.

I think the `TOPOLOGY` option for halos should have a different name.  We have an existing use of the term "topology" to refer to VBL/MBL, so seeing `TOPOLOGY` next to `MBL` is already a bit confusing. I assumed `TOPOLOGY` must refer to the token's VBL, but looking at the implementation it is actually computing the geometry of the token image, the same way "Generate Token Topology" does in the Edit Token dialog. So it's actually entirely different from VBL/MBL. If that's deliberate, we should distinguish this from VBL/MBL with a different name.

_Of course, you can also make the case that our existing use of the term "topology" for VBL/MBL is misguided and should change, while yours is the better use of term..._

Also, since there is an `MBL` option, it makes sense to me to also include the various VBLs as options, too.

Baaaaaz:
Thanks for taking a look!  Apologies for the initial mega commit, there is a lot going on so felt the description of the changes and of some design choices would help.  

1. Not wedded to having `TOPOLOGY` here. `OUTLINE` seems better, but open to better suggestions if anyone has any?

2. I couldn't see a use case for other VBL as halos, but for completeness they can be added relatively easily.

Also, just noticed having worked on 5833 that there may be a keystroke conflict for the Show Token Halos `CTRL+P` App Menu item, but then on investigation it appears not?

PointerTool.installKeystrokes() documents this;
```
   * <td>Meta P
   * <td>Expose fog from last path
```

But then the AppActions.EXPOSE_LAST_PATH_ACTION.getKeyStroke() seems to just have a keystroke of "P" with no modifier, so the PointerTool docs seem wrong?  Would you be able to confirm my findings please, otherwise I'll have to find another keystroke?

FullBleed:
>     1. Not wedded to having `TOPOLOGY` here. `OUTLINE` seems better, but open to better suggestions if anyone has any?

No horse in this race but maybe some other names that could be more unique to that action:
`PERIMETER`
`BOUNDARY`
`EDGE`
`GEOMETRY`

And if the action is as similar (same?) as what's done with "Generate Token Topology" perhaps they could share the same syntax?  The various uses of `TOPOLOGY` in MT has always seemed like something of a misnomer.

cwisniew:
> The various uses of `TOPOLOGY` in MT has always seemed like something of a misnomer.

If by "like something of a" you mean "a complete" you would be correct :)

Every time I see the word TOPOLOGY in the code, discord etc...

image

Baaaaaz:
Help!  So I seem to have an issue with this PR as when trying to open a pre-custom halos campaign file (e.g. 1.18.5) I am getting the following error...:
image
```
(MapTool.java:331) [AWT-EventQueue-0] ERROR - Could not load campaign.
java.lang.NullPointerException: Cannot invoke "java.util.List.stream()" because "this.haloSourceList" is null
        at net.rptools.maptool.model.Token.toDto(Token.java:3308)
        at net.rptools.maptool.model.Zone.lambda$toDto$22(Zone.java:2331)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
        at java.base/java.util.HashMap$ValueSpliterator.forEachRemaining(HashMap.java:1787)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
        at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
        at net.rptools.maptool.model.Zone.toDto(Zone.java:2331)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
        at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
        at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1939)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
        at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
        at net.rptools.maptool.model.Campaign.toDto(Campaign.java:764)
        at net.rptools.maptool.client.ServerCommandClientImpl.setCampaign(ServerCommandClientImpl.java:97)
[...]
```
I'm not sure why `toDto` is being called as doesn't it use `fromDto` when loading stuff in? I've had a go at fixing it but whatever I do seems to have ripple effects through the code.  What is the best way of fixing this?

cwisniew:
> Help! So I seem to have an issue with this PR as when trying to open a pre-custom halos campaign file (e.g. 1.18.5) I am getting the following error...: image

> I'm not sure why `toDto` is being called as doesn't it use `fromDto` when loading stuff in? I've had a go at fixing it but whatever I do seems to have ripple effects through the code. What is the best way of fixing this?

`toDto` is being called as the zone and token are being serialised to be sent to the server (even if you just start MapTool with no server it runs a "personal server").  One of the pitfalls of automated Java Serialization is when unmarshalling an object it does not run the initialization code for the object (constructors + variable assignments in method declaration). This is why your new field will be null. To fix this you need to place any initialisation code for that field (creating new map, populating etc) that would normally happen when the `new` operator is used in the readResolve() method of the Token class (as well as leaving it where it is for when `new` is used), e.g. in the `readResolve` method, check if the field is null, if so use `new` to assing to it, then perform any work that would have been done in the constructor on this field.

kwvanderlinde:
> Also, just noticed having worked on 5833 that there may be a keystroke conflict for the Show Token Halos `CTRL+P` App Menu item, but then on investigation it appears not?
> 
> PointerTool.installKeystrokes() documents this;
> 
> ```
>    * <td>Meta P
>    * <td>Expose fog from last path
> ```
> 
> But then the AppActions.EXPOSE_LAST_PATH_ACTION.getKeyStroke() seems to just have a keystroke of "P" with no modifier, so the PointerTool docs seem wrong? Would you be able to confirm my findings please, otherwise I'll have to find another keystroke?

In `AppActions.EXPOSE_LAST_PATH_ACTION`, the keystroke will also be <kbd>Ctrl + P</kbd> (except on mac). The `withMenuShortcut(...)` call should be adding the platform-specific modifier key.

src/main/java/net/rptools/maptool/model/HaloSource.java
```diff
@@ -0,0 +1,176 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ * Represents a halo source that can be attached to tokens, a halo source can contain one or more
+ * halo segments.
+ *
+ * <p>This class is immutable.
+ */
+public final class HaloSource implements Comparable<HaloSource>, Serializable {
```
kwvanderlinde: I would suggest renaming `Halo` to `HaloPart` or something, and then `HaloSource` to just `Halo`. The "source" term doesn't really carry over from lights.

src/main/java/net/rptools/maptool/model/AttachedHaloSource.java
```diff
@@ -0,0 +1,81 @@
...
+ * <http://www.gnu.org/licenses/> and specifically the Affero license
+ * text at <http://www.gnu.org/licenses/agpl.html>.
+ */
+package net.rptools.maptool.model;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import net.rptools.maptool.server.proto.AttachedHaloSourceDto;
+
+public final class AttachedHaloSource {
```
kwvanderlinde: I think we can skip this class and just reference halos by their GUID.

`The AttachedHaloSource#resolve(Token, Campaign)` method could be replaced by a method on `CategorizedHalos` that looks up a halo `GUID`.

src/main/java/net/rptools/maptool/model/HaloSource.java
```diff
@@ -0,0 +1,176 @@
...
+
+  private static final Logger log = LogManager.getLogger(HaloSource.class);
+
+  private final @Nullable String name;
+  private final @Nullable GUID id;
```
kwvanderlinde: These shouldn't support `null`. Lights do it because of personal lights, but halos don't have that.

src/main/java/net/rptools/maptool/model/HaloSource.java
```diff
@@ -0,0 +1,176 @@
...
+  public Object writeReplace() {
+    // Make sure XStream keeps the serialization nice. We don't need the XML to contain
+    // implementation details of the ImmutableList in use.
+    return new HaloSource(name, id, scaleWithToken, new ArrayList<>(haloList));
+  }
+
+  @SuppressWarnings("ConstantConditions")
+  @Serial
+  private @Nonnull Object readResolve() {
+    final List<Halo> halos = Objects.requireNonNullElse(haloList, Collections.emptyList());
```
kwvanderlinde: Since these halos are new and don't have backward compatibility concerns, we don't need to handle the case of `haloList` being `null`.

src/main/java/net/rptools/maptool/model/HaloSource.java
```diff
@@ -0,0 +1,176 @@
...
+    } catch (NoSuchElementException e) {
+      maxWidth = AppPreferences.haloLineWidth.get();
+    }
+
+    if ((maxWidth < AppPreferences.haloLineWidth.get() && countNulls > 0L) || maxWidth == null) {
+      maxWidth = AppPreferences.haloLineWidth.get();
+    }
+
+    return maxWidth;
+  }
```
kwvanderlinde: This method can be removed.

src/main/java/net/rptools/maptool/model/HaloSource.java
```diff
@@ -0,0 +1,176 @@
...
+  public int compareTo(@Nonnull HaloSource o) {
+    if (o != this) {
+      int nameLong = NumberUtils.toInt(name, Integer.MIN_VALUE);
+      int onameLong = NumberUtils.toInt(o.name, Integer.MIN_VALUE);
+      if (nameLong != Integer.MIN_VALUE && onameLong != Integer.MIN_VALUE)
+        return nameLong - onameLong;
+      return name.compareTo(o.name);
+    }
+    return 0;
+  }
```
kwvanderlinde: I don't think halos are actually sorted anywhere, so they don't need to be comparable.

src/main/java/net/rptools/maptool/model/HaloSource.java
```diff
@@ -0,0 +1,176 @@
...
+
+/**
+ * Represents a halo source that can be attached to tokens, a halo source can contain one or more
+ * halo segments.
+ *
+ * <p>This class is immutable.
+ */
+public final class HaloSource implements Comparable<HaloSource>, Serializable {
+
+  private static final Logger log = LogManager.getLogger(HaloSource.class);
```
kwvanderlinde: Unused field.

src/main/java/net/rptools/maptool/model/Token.java
```diff
@@ -989,6 +1003,133 @@ public List<AttachedLightSource> getLightSources() {
...
+      AttachedHaloSource ahs = i.next();
+      HaloSource haloSource = ahs.resolve(this, MapTool.getCampaign());
+      if (haloSource != null) {
+        List<Halo> halos = haloSource.getHaloList();
+        for (Halo halo : halos) {
+          i.remove();
+          break;
+        }
+      }
+    }
```
kwvanderlinde: `haloSourceList.clear()` may interest you.

src/main/java/net/rptools/maptool/model/Halo.java
```diff
@@ -0,0 +1,308 @@
...
+import net.rptools.maptool.model.drawing.DrawablePaint;
+import net.rptools.maptool.server.proto.HaloDto;
+
+public final class Halo implements Serializable {
+  private final @Nonnull DrawablePaint paint;
+  private final @Nullable HaloShapeType haloShapeType; // otherwise use default
+  private final @Nullable Integer width; // otherwise use default from AppPreferences
+  private final @Nullable ArrayList<Float> dashedPattern; // otherwise a solid line
+  private final boolean isGM;
+  private final boolean ownerOnly;
```
kwvanderlinde: I know this follows the same pattern as for auras in `LightSource`/`Light`, but it would be better if the entire `HaloSource` was marked GM or owner-only rather than individual components. 

FWIW I don't think it's such a great idea for auras either.

src/main/java/net/rptools/maptool/model/Token.java
```diff
@@ -989,6 +1003,133 @@ public List<AttachedLightSource> getLightSources() {
...
+  }
```
kwvanderlinde: This method can be removed. It is essentially `clearHaloSources()` except it is unable to remove unrecognized halos or halos that somehow have now parts.

src/main/java/net/rptools/maptool/model/Token.java
```diff
@@ -989,6 +1003,133 @@ public List<AttachedLightSource> getLightSources() {
...
+  public void removeHaloSource(GUID haloSourceId) {
+    haloSourceList.removeIf(ahs -> ahs.matches(haloSourceId));
+  }
+
+  /** Clear the haloSourceList */
+  public void clearHaloSources() {
+    haloSourceList.clear();
+  }
+
+  public boolean hasHaloSource(HaloSource source) {
```
kwvanderlinde: Can just pass a `GUID` here.

src/main/java/net/rptools/maptool/model/Token.java
kwvanderlinde, line 1: For all the new methods here that call `MapTool.getCampaign()` to look up halos: they should instead take a `Campaign` as a parameter.

src/main/java/net/rptools/maptool/model/Token.java
```diff
@@ -334,6 +337,13 @@ public String toString() {
    */
   private List<AttachedLightSource> lightSourceList = new ArrayList<>();
 
+  /**
+   * All halos sources attached to the token.
+   *
+   * <p>The elements should be unique, i.e., no two should reference the same halo source.
+   */
+  private List<AttachedHaloSource> haloSourceList = new ArrayList<>();
```
kwvanderlinde: This ties in with my comment on `AttachedLightSource`: if we use `GUID` to reference halos instead of `AttachedHaloSource`, we can also switch this field to a `LinkedHashSet<GUID>`. That would get us easy lookup, removal, zero duplicates, and would maintain insertion order, all without any extra code on our end.

kwvanderlinde:
> 1. Not wedded to having `TOPOLOGY` here. `OUTLINE` seems better, but open to better suggestions if anyone has any?

I like the term `OUTLINE`.

kwvanderlinde (changes requested):
On the whole I like the concept. Just noting some technical aspects that will make it easier to work with halos.

Baaaaaz:
> On the whole I like the concept. Just noting some technical aspects that will make it easier to work with halos.

Thanks for all these review comments.  I especially appreciate the feedback on the technical aspects as that is not an area I'm strong in so I'll have a go at addressing them...!

Baaaaaz:
> 
> In `AppActions.EXPOSE_LAST_PATH_ACTION`, the keystroke will also be Ctrl + P (except on mac). The `withMenuShortcut(...)` call should be adding the platform-specific modifier key.

But that is just it, there is no modifier key there that I can see!  Or am I missing something?
https://github.com/RPTools/maptool/blob/239a521d12420181a8c4237187ce0cc560e0abc4/src/main/java/net/rptools/maptool/client/AppActions.java#L2962-L2965

Baaaaaz:
> > In `AppActions.EXPOSE_LAST_PATH_ACTION`, the keystroke will also be Ctrl + P (except on mac). The `withMenuShortcut(...)` call should be adding the platform-specific modifier key.
> 
> But that is just it, there is no modifier key there that I can see! Or am I missing something?
> 
> https://github.com/RPTools/maptool/blob/239a521d12420181a8c4237187ce0cc560e0abc4/src/main/java/net/rptools/maptool/client/AppActions.java#L2962-L2965

Ah I did not understand your reply initially about the `withMenuShortcut(...)` , but following that method I think I get it now.

src/main/java/net/rptools/maptool/model/Token.java
Baaaaaz: Help!  How does one best handle such a LinkedHashSet<GUID> on `token` in its `data_transfer_objects` and `toDto` and `fromDto`?  I'm struggling getting the `toDto` and `fromDto` to work and cannot find an example to work from.  

In `data_transfer_objects` under `tokenDto` I've tried using a repeating `string`, `google.protobuf.StringValue`, and `HaloDto` (and then try to get the id to populate the LinkedHashSet) with and without stream() but I cannot seem to get them to work.  I need a bone.  TIA!

kwvanderlinde: A repeated string should work. I'll give it a go in a bit and see if I can share something concrete with you.

Baaaaaz: Thanks, eventually I managed to get the following to parse, but have I not tested it with non-personal server running and do know know if it is the best way!:

data_transfer_objects TokenDto:
`repeated string halo_guids = 73;`

token.fromDto:
`dto.getHaloGuidsList().stream().forEach(id -> token.haloIdSet.add(new GUID(id)));`

token.toDto:
`dto.addAllHaloGuids(haloIdSet.stream().map(GUID::toString).collect(Collectors.toList()));`

src/main/java/net/rptools/maptool/model/HaloSource.java
Baaaaaz: Done.  Renamed `Halo` to `HaloPart` and then `HaloSource` to `Halo`.

src/main/java/net/rptools/maptool/model/AttachedHaloSource.java
Baaaaaz: Removed.

src/main/java/net/rptools/maptool/model/HaloSource.java
Baaaaaz: Made `@Nonnull`

src/main/java/net/rptools/maptool/model/HaloSource.java
Baaaaaz: Removed.

src/main/java/net/rptools/maptool/model/Halo.java
Baaaaaz: Moved to `Halo` (what was `Halosource`)

src/main/java/net/rptools/maptool/model/Token.java
Baaaaaz: Removed.

src/main/java/net/rptools/maptool/model/Token.java
Baaaaaz: Done.  Note renamed to `hasHalo()`

src/main/java/net/rptools/maptool/model/Token.java
Baaaaaz: Done.

src/main/java/net/rptools/maptool/model/HaloSource.java
Baaaaaz: Removed.

src/main/java/net/rptools/maptool/model/HaloSource.java
Baaaaaz: Removed.

Baaaaaz:
@kwvanderlinde ready for another review!  Not sure what to do about the failed check - says it is a macos-13 brownout?

kwvanderlinde:
> @kwvanderlinde ready for another review! Not sure what to do about the failed check - says it is a macos-13 brownout?

Thanks for highlighting this. It's not any issue with your changes, so just ignore it for now. I opened #5842 and will work to address it.

messages/src/main/proto/data_transfer_objects.proto
```diff
@@ -100,6 +100,58 @@ message LightDto {
   int32 lumens = 9;
 }
 
+message HaloDto {
+  google.protobuf.StringValue id = 1;
+  google.protobuf.StringValue name = 2;
```
kwvanderlinde: Sorry, I can't believe I missed this the first time around since I commented on `HaloSource` (now `Halo`) about the same thing. These should just be `string id` and `string name` since `Halo#id` and `Halo#name` are now `@Nonnull`.

kwvanderlinde (changes requested):
One minor point, otherwise I think this is good to merge.

messages/src/main/proto/data_transfer_objects.proto
Baaaaaz: Now corrected.

kwvanderlinde (approved):

kwvanderlinde:
Now I get to merge this with my own giant changes :smile:

PR #5851, opened 2025-11-08
Make `Scale` read-only

### LINKED ISSUE #5849: [Maintenance]: Read-only `Scale`
### Describe the problem

The current `Scale` is explicitly mutable, and is even responsible for updating listeners about changes. This does not work very well in asynchronous situations, as the class is neither thread safe nor can the object be guaranteed not to have changed by the time it is used. This is already an issue with the pending LibGDX renderer despite running in the AWT thread, and will be more of an issue if the LibGDX renderer ultimately runs in its own thread.

- Make `Scale` read-only
- Move the ownership of the zone's `Scale` from `ZoneRenderer` to `ZoneViewModel`.
- Prefer operating directly on the `Scale` rather than using `ZoneRenderer` methods to modify it.

### Expected Benefits

- It will be easier to rely on the `Scale` object in more places.
- Code will be less dependent on the `ZoneRenderer` for state management, further aligning it with future alternative renderers.

### Additional Context

_No response_

kwvanderlinde:

Resolves #5849

### Description of the Change

Changes the interface of `Scale` so its properties can no longer be modified. The fields are not `final` so that serialization continues to work, but they are effectively immutable. As a proper value type, `Scale` now has `#equals(Object)` and `#hashCode()` methods. This also means `Scale` is no longer reponsible for pumping events when modified, so the `PropertyChangeSupport` has been removed.

`Scale` is now the only place that defines transformations between world and screen space. It provides general `AffineTransform`s in both directions via `#toWorldTransform()` and `#toScreenTransform()`. Also, most conversion methods on `ScreenPoint` have been removed in favour of using the `Scale` for conversions. Those conversion methods that remain in `ScreenPoint` are wrappers around the ones in `Scale` for specific circumstances, and they all accept a simple `Scale` rather than an entire `ZoneRenderer`.

The `ZoneViewModel` now manages the `Scale` for a `Zone` (used to be in `ZoneRenderer`). This is just one more step along the road of removing dependencies on `ZoneRenderer`, though for now most components still look up the view model from the renderer. `ZoneRenderer` now longer offers methods to directly change the current `Scale`, though it does offer a few convenience methods that may do so among other things.

A few `Scale`-related improvements as well:
- The new `RepaintZoneRequested` event gives a way for any component to trigger a repaint via the event bus. For now only `ZoneViewModel` uses it when the `Scale` is changed, but there are plenty of other places that will benefit from this as well.
- `HexGrid#getOffV(ZoneRenderer)` and `HexGrid#getOffU(ZoneRenderer)` now accept a `Scale` instead of a `ZoneRenderer`. Similarly, `HexGrid#getRendererSizeV(ZoneRenderer)` and `HexGrid#getRendererSizeU(ZoneRenderer)` are now `HexGrid#getSizeV(Dimension2D)` and `HexGrid#getSizeU(Dimension2D)`.
- `DrawablePaint` and `DrawableNoise` now accept a full `Scale` when getting the AWT paint. This isn't used much now, but will be more common with some upcoming rendering changes.

### Possible Drawbacks

Some scaling operations are more verbose to write now, though that will improve as `ZoneRenderer` is pushed out of the limelight.

### Documentation Notes

N/A

### Release Notes

- Refactored zone scale

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5851)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/AppActions.java
+++ b/src/main/java/net/rptools/maptool/client/AppActions.java
@@ -2363,7 +2363,7 @@ public class AppActions {
           if (campaign.currentView != null) {
             current.setZoneScale(campaign.currentView);
           }
-          current.getZoneScale().reset();
+          current.setZoneScale(current.getZoneScale().withResetZoomLevel());
         }
         MapTool.getAutoSaveManager().tidy();
 
--- a/src/main/java/net/rptools/maptool/client/AppActions.java
+++ b/src/main/java/net/rptools/maptool/client/AppActions.java
@@ -831,9 +831,11 @@ public class AppActions {
           ScreenPoint screenPoint = renderer.getPointUnderMouse();
           if (screenPoint == null) {
             // Pick the middle of the map
-            screenPoint = ScreenPoint.fromZonePoint(renderer, renderer.getCenterPoint());
+            var centerPoint = renderer.getCenterPoint();
+            screenPoint =
+                renderer.getViewModel().getZoneScale().toScreenSpace(centerPoint.x, centerPoint.y);
           }
-          ZonePoint zonePoint = screenPoint.convertToZone(renderer);
+          ZonePoint zonePoint = screenPoint.convertToZone(renderer.getViewModel().getZoneScale());
           pasteTokens(zonePoint, renderer.getActiveLayer());
           keepIdsOnPaste = false; // once pasted, subsequent paste should have new ids
           renderer.repaint();
--- a/src/main/java/net/rptools/maptool/client/AppActions.java
+++ b/src/main/java/net/rptools/maptool/client/AppActions.java
@@ -1809,7 +1809,10 @@ public class AppActions {
           ZoneRenderer renderer = MapTool.getFrame().getCurrentZoneRenderer();
           if (renderer != null) {
             Dimension size = renderer.getSize();
-            renderer.zoomIn(size.width / 2, size.height / 2);
+
+            var viewModel = renderer.getViewModel();
+            viewModel.setZoneScale(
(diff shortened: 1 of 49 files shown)


cwisniew (approved):

Baaaaaz:
Was just going to raise a bug ticket to say that MapTool-nightly-20251109 is broken, as when trying to run the win exe I get the following error:

```
java.lang.NullPointerException: Cannot invoke "net.rptools.maptool.client.ui.Scale.equals(Object)" because "this.lastZoneScale" is null
	at net.rptools.maptool.client.ui.zone.renderer.ZoneRenderer.renderBoard(ZoneRenderer.java:1057)
	at net.rptools.maptool.client.ui.zone.renderer.ZoneRenderer.renderZone(ZoneRenderer.java:829)
	at net.rptools.maptool.client.ui.zone.renderer.ZoneRenderer.lambda$paintComponent$2(ZoneRenderer.java:679)
	at net.rptools.lib.CodeTimer.using(CodeTimer.java:50)
	at net.rptools.maptool.client.ui.zone.renderer.ZoneRenderer.paintComponent(ZoneRenderer.java:656)
	at java.desktop/javax.swing.JComponent.paint(Unknown Source)
	at java.desktop/javax.swing.JComponent.paintChildren(Unknown Source)
	at java.desktop/javax.swing.JComponent.paint(Unknown Source)
	...
```

kwvanderlinde:
Whoops, I'll get this fixed up ASAP.

kwvanderlinde:
On the plus side, I think this is the first time sentry has correctly identified the source of a problem :sweat_smile:

cwisniew:
> On the plus side, I think this is the first time sentry has correctly identified the source of a problem 😅

You know what they say about broken clocks :)

PR #5868, opened 2025-11-17
Add UI to Edit Table dialog for pick once tables

### LINKED ISSUE #5818: [Feature]: UI support for pick-once tables
### Describe the Problem

If I want a pick-once table, I have to use the [`setTablePickOnce()`](https://wiki.rptools.info/index.php/setTablePickOnce) macro function. This makes it inaccessible to users who are not familiar with macros and who have a vanilla MT installation without libraries that can do it for them.

### The Solution you'd like

Add a checkbox to the **Edit Table** dialog to mark the table as pick-once.

_No response_

### Additional Context

The original FR #1337 called for such a checkbox, but only macro functions were added.

kwvanderlinde:

Resolves #5818

### Description of the Change

Adds a few elements to the Edit Lookup table dialog to support pick once tables:
1. A checkbox to mark the table as pick once.
2. A button to reset the table's picks.
3. A new column in the table to allow the user to view and modify which entries have been picked.

When (1) is selected, the _Roll_ field is disabled, (2) is enabled, and (3) is made visible. When (1) is not selected, the _Roll_ field is enabled, (2) is disabled, and (3) is hidden.

As with the existing UI, none of the new element effect the table until the _Accept_ button is hit. Also, since picks are actually handled now, they won't be lost when a table is edited unless the user explicitly resets them.

Also did some refactoring and fixes on `EditLookupTablePanel`:
- It now uses `GenericDialog` for some nice features.
- The Accept button is now the default button for the panel, and is highlighted as such
- Leverages the view's field to get components in a type-safe way. Added a bit to AbeillePanel to allow replacing concrete component without using names.
- The table rows are modeled with a custom type rather than being a list of `String`.  This is needed to support the new `boolean` "Picked?" column.
- Fixes the size of the Table Image to 150x150 to solve solve some strange reflowing as the dialog is resized.

### Possible Drawbacks

Should be none

### Documentation Notes

The UI can now be used to set a table as pick once and to modify the picked entries.

When pick once is enabled:
image

When pick once is disabled:
image

### Release Notes

- Added the ability to set a table as pick-once in the Edit Table dialog.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5868)


FullBleed:
This is great.

I do, however, wish the "Reset Picks" button was inline with the "Pick Once" check box and disabled when the checkbox is unchecked though.

kwvanderlinde:
> This is great.
> 
> I do, however, wish the "Reset Picks" button was inline with the "Pick Once" check box and disabled when the checkbox is unchecked though.

Sure, I can move it inline. The button is already being disabled as you'd like.

FullBleed:
Heh, right after I hit send my brain registered your disabled description in the OP.  ;)

cwisniew (approved):

Baaaaaz:
@kwvanderlinde nice change, but I just noticed odd things when running the nightly build:

1. The individual entry add/remove image buttons in the Edit Lookup table dialog now only seem to work if the new **Pick Once Table** checkbox is `checked`.  Otherwise they seem unresponsive.

2. Individual entry images are now a lot smaller compared to before if the **Pick Once Table** checkbox is **unchecked**.  e.g. if you load the default tables and then open one

In 1.18.6:
image

In nightly build & **Pick Once Table** checkbox is `unchecked`
image

However if you set the table to be a Pick Once Table and then click on the Image column, they suddenly all get bigger.  

Or if you set to be a Pick Once Table, Accept the changes and reopen the Edit Lookup table dialog for that table it the entry images are now bigger. e.g.

In nightly build & **Pick Once Table** checkbox is `checked`
image

kwvanderlinde:
I'll take a look into why that is (and try to remember what is going on in there :sweat_smile:)

Issue #5831, opened 2025-10-22
[Bug]: New Campaign Properties dialog loses changes if there are definition syntax errors (Develop/1.19)


Baaaaaz:
### Describe the Bug

New Campaign Properties dialog [#5657](https://github.com/RPTools/maptool/issues/5657) due in 1.19 closes and loses any changes if there are sight/light/aura syntax errors and the user presses OK.

Previous behavior (i.e. in 18.5) is to stay open and allow the user to correct the syntax errors, or to click Cancel.

### To Reproduce

Run Develop or a current nightly build
Open Campaign Properties
Add an incorrect sight, light, or aura definition (e.g. `error: blah`)
Click OK
Observe Feedback dialog, press OK
Observe Error dialog, press OK
Campaign Properties dialog closes and any changes are lost

### Expected Behaviour

New Campaign Properties dialog to stay open after syntax errors detected to give the user the opportunity to fix the syntax errors, or to click Cancel.

### Screenshots

<img width="759" height="747" alt="Image" src="https://github.com/user-attachments/assets/8e64485e-d9c6-4856-831c-e76e53b8b7fe" />

<img width="758" height="747" alt="Image" src="https://github.com/user-attachments/assets/58c25d83-dc0a-41c4-aa06-813a41e5cb02" />

Then Campaign Dialog closes.

### MapTool Info

Develop, current nightly builds

### Additional Context

Found when developing new custom halos https://github.com/RPTools/maptool/pull/5824, as this was also affected by the CD dialog closing when syntax errors were detected.

PR #5874, opened 2025-11-18
Fix for partial commits in Edit Table

### LINKED ISSUE #5869: [Bug]: Edit Table dialog partially commits when applying invalid range
### Describe the Bug

When editing an exsting table, if I modify the range of some entry in the middle to be invalid and then try to apply it, I will get an error telling me about the problem and allowing me to correct it. But if I then _Cancel_, the table will contain only those entries higher up than the error, losing everything else.

### To Reproduce

1. Add a new table named `Error test`. Give it six entries with the ranges set to `1`, `2`, `3`, `4`, `5`, `6`, respectively.
2. Edit the table. Modify the `3` range to `3-four` and click _Accept_.
3. See the error dialog saying `Table "Error test" contains bad range "3-four" on row 2`. C this dialog.
3. In the Edit Table dialog, press the _Cancel_ button.
4. Edit the table again. See that only the `1` and `2` entries still exist.

### Expected Behaviour

A table should not be modified when the Edit Table dialog is closed by the _Cancel_ button. It should not be possible to commit partial results whether using the _Accept_ or _Cancel_ buttons.

### Screenshots

Original table:
Image

Modified, with error:
Image

After cancelling and reopening:
Image

### MapTool Info

1.18.5

### Desktop

Linux Mint 22.2

### Additional Context

_No response_

kwvanderlinde:

Fixes #5869

### Description of the Change

Changes how the **New Table** / **Edit Table** commits the entries to a `LookupTable`. All entries are now parsed prior to any modification of the `LookupTable`, stored in a list for later consumption. Only once all validations are completed are any table properties modified or entries added.

Also fixes an issue where `GenericDialog` would ignore the result of `AbeillePanel#commit()`. This caused a problem where dialogs with validation errors would close themselves after the user dismissed the validation error modal. So far this has only affected the **Edit Token** and **Add Resource** dialogs in development builds. Other dialogs do not yet take full advantage of `GenericDialog` and have to handle the commit failures explicitly, so they do not have this issue.

### Possible Drawbacks

None

### Documentation Notes

N/A

### Release Notes

- Fixed a bug where canceling the Edit Token dialog after a validation error would result in table data being lost.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5874)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/swing/GenericDialog.java
+++ b/src/main/java/net/rptools/maptool/client/swing/GenericDialog.java
@@ -337,7 +337,10 @@ public class GenericDialog extends JDialog {
     if (this._content instanceof AbeillePanel<?> abeillePanel && abeillePanel.getModel() != null) {
       // wrap up any AbeillePanel commits and unbinding
       if (this.getDialogResult().equals(AFFIRM)) {
-        abeillePanel.commit();
+        if (!abeillePanel.commit()) {
+          // Do not close the dialog since validation failed.
+          return;
+        }
       }
       abeillePanel.unbind();
     }
--- a/src/main/java/net/rptools/maptool/client/ui/lookuptable/EditLookupTablePanel.java
+++ b/src/main/java/net/rptools/maptool/client/ui/lookuptable/EditLookupTablePanel.java
@@ -222,52 +222,64 @@ public class EditLookupTablePanel extends AbeillePanel<LookupTable> {
       MapTool.showError(I18N.getText("EditLookupTablePanel.error.invalidSize", name));
       return false;
     }
+    var isPickOnce = view.getPickOnce().isSelected();
+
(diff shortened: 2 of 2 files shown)


Baaaaaz:
> Also fixes an issue where GenericDialog would ignore the result of AbeillePanel#commit(). This caused a problem where dialogs with validation errors would close themselves after the user dismissed the validation error modal. So far this has only affected the Edit Token and Add Resource dialogs in development builds. Other dialogs do not yet take full advantage of GenericDialog and have to handle the commit failures explicitly, so they do not have this issue.

Admittedly I could wait and check once this ticket is merged but....  do you think fixing this issue will also resolve https://github.com/RPTools/maptool/issues/5831?

cwisniew (approved):

PR #5891, opened 2025-11-28
Further Drawing/Template improvements and new MTScript functions

### LINKED ISSUE #2743: Drawing and Template Improvements
Drawings and templates cannot be easily manipulated once drawn, whether by moving them around on the map or by macros. Currently, the only way to get the ID of a drawing is via the Draw Explorer interface, so it is difficult to use macros to affect them. Even if that were possible, what can be done with those drawings is currently very limited.

I would like for drawings to be objects that can be interacted with similar to a token. This would be especially useful for templates that show area of effect, allowing players to create an appropriately-sized area and move it around to see how many targets they can affect, rather than having to redraw the template each time.

If possible, it might be useful for AoE templates to enable 'scrolling' through them to rotate them, especially for shapes like cones and lines.

More powerful usages of the drawings would require increased macro support via:
-  Addition of drawings to the getTokens() function, or a getDrawings() function to allow for collection of drawing IDs outside of Draw Explorer.
- A getLastDrawing() function, which might be useful for tracking AoEs since they are often used 
- A createTemplate() function that could make templates based on defined parameters - could return the ID of the created drawing as well for use in macro. createTemplate() and not createDrawing() because I have no idea how you could possibly replicate everything a drawing does in a macro.
- A copyDrawing() or moveDrawing() function that works similarly to copy/moveToken()
- A getTokensOverDrawing() (idk maybe needs a better name) function that returns all token IDs contained within a drawing to identify targets affected by an AoE
- There are probably more things that could be done with this (e.g. changing the size/color of templates) but I admittedly don't use drawings excessively for much at this stage

Proposed macro flow for the above in action:
- Player casts an AoE spell that creates an AoE template (either with createTemplate(), moveDrawing(), or copyDrawing())
- Player moves the newly created/moved template into position underneath enemies
- Second macro applies damage to targets within the AoE using getTokensOverDrawing()

Alternative to the above using getLastDrawing():
- Player casts AoE spell
- Player makes a drawing
- Second macro applies damage to targets within the AoE using getTokensOverDrawing(getLastDrawing())

As far as I'm aware, there is no alternative option for physically moving the drawings. There is a possible alternative for the functionality of grabbing tokens within a certain AoE using current functionality, though - it could be partially simulated using a token with an aura. However, auras do not have the same capability for creating shapes (e.g. lines, freeform wall templates) and are more cumbersome if an effect is centered at the intersection of 4 grid squares as opposed to on a single grid square.

Baaaaaz:

closes #2743

### Description of the Change
Lots of new template and drawable MTScript functions and some new tool features.  Drawables selected via the appropriate tool are now accessible to Drawable (i.e. drawing and template) and template-only functions.

**DrawingPointerTool**
now exposes tool selected drawings/templates to new `getSelectedDrawings()` MTScript function, and also to the helper function `getSingleSelectedDrawnElement()`.
- Slight change to movement line dash pattern when dragging templates
- New ALT drag feature on Blast templates now allows moving the template around the vertex without resizing it.
- New ALT drag feature on Cone templates now allows rotating the template around the vertex without resizing it.

** ConeTemplate**

- Change direction to be protected so is accessible from extending classes (i.e. the BlastTemplate)

** BlastTemplate**

- Now get/sets direction property based on where the blast template is placed

**DrawingFunctions**
 new/updated MapTool methods to support the new MTScript functions
- enum `DrawableUnit` to facilitate cells, distance, or (zone) pixels units for creating templates, moving drawables/templates , or getting info on dimensions about them
- `getDrawingJSONInfo()` -> expose more information on templates
- `boundsToJSON()` -> corrected for templates which previously had a padding applied and was incorrect
- `getDrawingBounds()` -> helper method used in the new `DrawingGetterFunctions`
- `findDrawnElement()` -> used by MT script template/drawable functions to find drawings and templates using the name or id
- `getSingleSelectedDrawnElement()` -> used in MT script template/drawable functions to find a drawable selected by the Drawing/Template pointer tools.

**TemplateFunctions**
new class with new MTScript functions just for templates:
- `createTemplate()` can create any type of template
- `getTemplateX()` gets a template's vertex x position
- `getTemplateY()` gets a template's vertex y position
- `getTemplateRadius()` gets a template's radius
- `moveTemplate()` move a template's vertex position

**DrawingMiscFunctions**
new MTScript functions (and supporting methods):

- `duplicateDrawable()`
- `getSelectedDrawables()`
- `getDrawablesUnderToken()`
- `getTokensOverDrawable()`
- `moveDrawable()`

**DrawingGetterFunctions**
new MTScript functions that expand on existing drawing functions and makes getting some drawable properties easier, such as they also work with drawables selected via the respective tool, and take the units parameter.

- `getDrawableInfo()`
- `getDrawableLayer()`
- `getDrawableName()`
- `getDrawableX()`
- `getDrawableY()`
- `getDrawableCenterX()`
- `getDrawableCenterY()`
- `getDrawableMaxX()`
- `getDrawableMaxY()`
- `getDrawableHeight()`
- `getDrawableWidth()`

**DrawingSetterFunctions**

- `setDrawableLayer()`
- `setDrawableName()`

### Possible Drawbacks

Over use of templates in games ;)
I have not java coded things in the best way.

### Documentation Notes
t.b.c.

### Release Notes
Added new MTScript functions for templates and drawables, inc. `createTemplate()`, `moveTemplate()`

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5891)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
@@ -20,23 +20,15 @@ import java.awt.Point;
 import java.awt.geom.Ellipse2D;
 import java.awt.geom.PathIterator;
 import java.math.BigDecimal;
+import java.util.ArrayList;
 import java.util.List;
 import net.rptools.maptool.client.MapTool;
+import net.rptools.maptool.client.tool.drawing.DrawingPointerTool;
+import net.rptools.maptool.client.tool.drawing.TemplatePointerTool;
 import net.rptools.maptool.language.I18N;
-import net.rptools.maptool.model.GUID;
-import net.rptools.maptool.model.Zone;
+import net.rptools.maptool.model.*;
 import net.rptools.maptool.model.Zone.Layer;
-import net.rptools.maptool.model.drawing.AbstractDrawing;
-import net.rptools.maptool.model.drawing.AbstractTemplate;
-import net.rptools.maptool.model.drawing.Drawable;
-import net.rptools.maptool.model.drawing.DrawableColorPaint;
-import net.rptools.maptool.model.drawing.DrawablePaint;
-import net.rptools.maptool.model.drawing.DrawableTexturePaint;
-import net.rptools.maptool.model.drawing.DrawablesGroup;
-import net.rptools.maptool.model.drawing.DrawnElement;
-import net.rptools.maptool.model.drawing.LineSegment;
-import net.rptools.maptool.model.drawing.Pen;
-import net.rptools.maptool.model.drawing.ShapeDrawable;
+import net.rptools.maptool.model.drawing.*;
 import net.rptools.maptool.util.FunctionUtil;
 import net.rptools.parser.Parser;
 import net.rptools.parser.ParserException;
@@ -47,6 +39,67 @@ public class DrawingFunctions extends AbstractFunction {
 
   public DrawingFunctions(int minParameters, int maxParameters, String... aliases) {
     super(minParameters, maxParameters, aliases);
+  }
+
+  /**
+   * DrawableUnit enumeration, including an <code>index</code> property based on the "units" boolean
+   * argument found in certain MTScript functions, but as a tri-state with <code>CELLS</code>,
+   * <code>DISTANCE</code>, and <code>PIXELS</code> options. For template radii the units used
+   * typically will be <code>DISTANCE</code>.
+   */
+  protected enum DrawableUnit {
+    CELLS(0),
+    DISTANCE(1),
+    PIXELS(2);
+
+    DrawableUnit(int index) {
+      this.index = index;
+    }
+
+    private final int index;
+
+    public static DrawableUnit valueOfIndex(int index) {
+      for (DrawableUnit drawableUnit : values()) {
+        if (drawableUnit.index == index) {
+          return drawableUnit;
+        }
+      }
+      return null;
+    }
+
+    public static int convert(Zone zone, int value, DrawableUnit fromUnit, DrawableUnit toUnit) {
+      double fromConversion = getConversionFactor(zone, fromUnit);
+      double toConversion = getConversionFactor(zone, toUnit);
+      return (int) (value * toConversion / fromConversion);
+    }
+
+    public static int convertToCells(Zone zone, int value, DrawableUnit fromUnit) {
+      return convert(zone, value, fromUnit, DrawableUnit.CELLS);
+    }
+
+    public static double getConversionFactor(Zone zone, DrawableUnit drawableUnit) {
+      return switch (drawableUnit) {
+        case CELLS -> 1d;
+        case DISTANCE -> zone.getUnitsPerCell();
+        case PIXELS -> zone.getGrid().getSize();
+      };
+    }
+
+    public static int getIndex(DrawableUnit drawableUnit) {
+      return drawableUnit.index;
+    }
+
+    public static int getCellsIndex() {
+      return DrawableUnit.CELLS.index;
+    }
+
+    public static int getDistanceIndex() {
+      return DrawableUnit.DISTANCE.index;
+    }
+
+    public static int getPixelsIndex() {
+      return DrawableUnit.PIXELS.index;
+    }
   }
 
   @Override
@@ -124,8 +177,47 @@ public class DrawingFunctions extends AbstractFunction {
     DrawnElement drawnElement = findDrawnElement(map.getAllDrawnElements(), guid);
     if (drawnElement != null) return drawnElement;
     throw new ParserException(
-        I18N.getText(
-            "macro.function.drawingFunction.unknownDrawing", functionName, guid.toString()));
+        I18N.getText("macro.function.drawable.unknownDrawable", functionName, guid.toString()));
+  }
+
+  /**
+   * Try and find a drawn element, using its {@link Drawable}'s {@link GUID} or case-insensitive
+   * <code>name</code>.
+   *
+   * <p>Warning: If name is used, and there are multiple drawables with the same name on the same
+   * map, only the first one encountered will be returned
+   *
+   * @param functionName this is used in the exception message
+   * @param zone the map to search for the template
+   * @param drawnElementRef either the name, or the id
+   * @return the template's drawn element, or null if not found
+   */
+  public static DrawnElement findDrawnElement(
+      String functionName, Zone zone, String drawnElementRef) throws ParserException {
+
+    DrawnElement drawnElement = null;
+
+    GUID guid = null;
+    try {
+      guid = GUID.valueOf(drawnElementRef);
+    } catch (Exception e) {
+      // wasn't a GUID after all, OK to ignore
+    }
+
+    for (DrawnElement de : zone.getAllDrawnElements()) {
+      AbstractDrawing d = (AbstractDrawing) de.getDrawable();
+      if (d.getId().equals(guid)) {
+        drawnElement = de;
+        break;
+      } else if (drawnElementRef.equalsIgnoreCase(d.getName())) {
+        drawnElement = de;
+        break;
+      }
+    }
+
+    if (drawnElement != null) return drawnElement;
+    throw new ParserException(
+        I18N.getText("macro.function.drawable.unknownDrawable", functionName, drawnElementRef));
   }
 
   private DrawnElement findDrawnElement(List<DrawnElement> drawableList, GUID guid) {
@@ -236,6 +328,22 @@ public class DrawingFunctions extends AbstractFunction {
   }
 
   /**
+   * Get the name of the drawing (if it has one).
+   *
+   * @param functionName this is used in the exception message
+   * @param zone the zone that should contain the drawing
+   * @param guid the id of the drawing
+   * @return the drawing name
+   * @throws ParserException the exception
+   */
+  protected String getDrawingName(String functionName, Zone zone, GUID guid)
+      throws ParserException {
+    DrawnElement de = getDrawnElement(functionName, zone, guid);
+    AbstractDrawing ad = (AbstractDrawing) de.getDrawable();
+    return ad.getName() == null ? "" : ad.getName();
+  }
+
+  /**
    * Converts the selected drawing element into a JSON string.
    *
    * @param functionName this is used in the exception message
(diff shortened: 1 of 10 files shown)


Baaaaaz:
Example GM panel macroset with new functions:
[2743 gm macros for templates and drawables.zip](https://github.com/user-attachments/files/23812255/2743.gm.macros.for.templates.and.drawables.zip)

Baaaaaz:
Ready for review!

Baaaaaz:
Will this one be able to go into 1.19?

src/main/java/net/rptools/maptool/model/drawing/BlastTemplate.java
```diff
@@ -53,6 +54,7 @@ public BlastTemplate(BlastTemplate other) {
     super(other);
     this.offsetX = other.offsetX;
     this.offsetY = other.offsetY;
+    this.direction = other.direction;
```
kwvanderlinde: The direction is already copied in the `ConeTemplate` constructor and does not need to be set here.

src/main/java/net/rptools/maptool/model/drawing/BlastTemplate.java
```diff
@@ -135,6 +138,77 @@ public void setControlCellRelative(int relX, int relY) {
...
+   * Set the {@link BlastTemplate}'s {@link Direction} and then set the control offsets for that
+   * direction. While this is coarser than setting the control offsets explicitly, it provides an
+   * alternative and user-friendly way of setting the control offsets when not using the {@link
+   * BlastTemplateTool}.
+   *
+   * @param direction The direction to draw the blast from the center vertex.
+   */
+  public void setDirectionControlCellOffset(Direction direction) {
+    if (direction != null) {
+      this.direction = direction.name();
```
kwvanderlinde: Can be replaced with `setDirection(direction)`.

src/main/java/net/rptools/maptool/model/drawing/ConeTemplate.java
```diff
@@ -38,7 +38,7 @@ public class ConeTemplate extends RadiusTemplate {
    * of the selected vertex. Saved as a string as a hack to get around the hessian library's problem
    * w/ serializing enumerations.
    */
-  private String direction = Direction.SOUTH_EAST.name();
+  protected String direction = Direction.SOUTH_EAST.name();
```
kwvanderlinde: Along with my comments on `BlastTemplate`, this can remain `private` as an implementation detail.

src/main/java/net/rptools/maptool/client/functions/TemplateFunctions.java
```diff
@@ -0,0 +1,782 @@
...
+      public AbstractTemplate template() {
+        return new WallTemplate();
+      }
+    };
+
+    TemplateType(String alias) {
+      this.alias = alias;
+    }
+
+    public AbstractTemplate template() {
```
kwvanderlinde: Can be `abstract` to force the enum constants to provide it.

src/main/java/net/rptools/maptool/client/functions/TemplateFunctions.java
```diff
@@ -0,0 +1,782 @@
...
+    FunctionUtil.blockUntrustedMacro(functionName);
+
+    if (FUNC_CREATE_TEMPLATE.equalsIgnoreCase(functionName)) {
+      // templateType, radius/wallPath [, units, [, options [, delim [, mapRef ]]]]
+      FunctionUtil.checkNumberParam(FUNC_CREATE_TEMPLATE, parameters, 2, 6);
+
+      // template type, determine first by alias, then name (i.e. enum value)
+      String templateType = parameters.getFirst().toString().toUpperCase().trim();
+      TemplateType tt = TemplateType.valueOfAlias(templateType);
+      if (tt == null) {
```
kwvanderlinde: I think we can just use the abbreviated names in `createTemplate()` and not bother supporting the `…Template` alternatives.

src/main/java/net/rptools/maptool/client/functions/TemplateFunctions.java
```diff
@@ -0,0 +1,782 @@
...
+              : findDrawnElement(FUNC_GET_TEMPLATE_RADIUS, zone, parameters.get(1).toString());
+      if (!(de.getDrawable() instanceof AbstractTemplate)) {
+        throw new ParserException(
+            I18N.getText(
+                "macro.function.template.unknownTemplate", functionName, de.getDrawable().getId()));
+      }
+
+      return BigDecimal.valueOf(
+          getTemplateRadius(FUNC_GET_TEMPLATE_RADIUS, zone, de, drawableUnit));
+
```
kwvanderlinde, line 238: We should reject `WallTemplate` in this function since it doesn't have a meaningful concept of a "radius". Because it is an `AbstractTemplate` it has to implement `#getRadius()`, but the returned value doesn't really have anything in common with what all other templates call a "radius".

It's been on my TODO list for a while to refactor the `AbstractTemplate` hierarchy specifically to separate `WallTemplate` from all the other "radial templates" as I've started calling them.

src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
```diff
@@ -260,19 +368,78 @@ protected JsonObject getDrawingJSONInfo(String functionName, Zone map, GUID guid
...
     dinfo.addProperty("penWidth", el.getPen().getThickness());
     dinfo.add("path", pathToJSON(d));
-    if (el.getDrawable() instanceof AbstractTemplate t) {
-      dinfo.addProperty("templateSize", t.getRadius());
+    dinfo.addProperty(
+        "isTemplate",
+        el.getDrawable() instanceof AbstractTemplate ? BigDecimal.ONE : BigDecimal.ZERO);
+    if (el.getDrawable() instanceof AbstractTemplate at) {
+      JsonObject template = new JsonObject();
+      template.addProperty("radius", at.getRadius());
```
kwvanderlinde: Backwards compatibility issue here: we shouldn't change the `"templateSize"` key to `"radius"` since macros written pre-1.19 may expect to find the former.

Also, in keeping with my other comment on the new template functions, wall templates don't really have a "radius" per se, so even if we added `"radius"` in addition to `"templateSize"`, we ought to exclude that property for wall templates.

src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
```diff
@@ -315,8 +482,8 @@ private JsonArray pathToJSON(AbstractDrawing d) {
         while (!pathIter.isDone()) {
           pathIter.currentSegment(coords);
           JsonObject info = new JsonObject();
-          info.addProperty("x", coords[0]);
-          info.addProperty("y", coords[1]);
+          info.addProperty("x", (int) coords[0]);
+          info.addProperty("y", (int) coords[1]);
```
kwvanderlinde: Why cast to `int` here? Fractional coordinates are valid in shapes.

src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
```diff
@@ -260,19 +368,78 @@ protected JsonObject getDrawingJSONInfo(String functionName, Zone map, GUID guid
...
+            for (CellPoint cp : wt.getPath()) {
+              ZonePoint zp = map.getGrid().convert(cp);
+              JsonObject pointInfo = new JsonObject();
+              pointInfo.addProperty("x", zp.x);
+              pointInfo.addProperty("y", zp.y);
+              pathInfo.add(pointInfo);
+            }
+            template.add("wallPath", pathInfo);
+          } else {
+            template.addProperty("wallPath", "null");
```
kwvanderlinde: A `null` path is not an expected case for `WallTemplate`. If it's possible for `WallTemplate#getPath()` to return `null` we should fix that in `WallTemplate` and not expose this case to macros.

src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
```diff
@@ -260,19 +368,78 @@ protected JsonObject getDrawingJSONInfo(String functionName, Zone map, GUID guid
...
+          controlOffset.addProperty("x", bt.getOffsetX());
+          controlOffset.addProperty("y", bt.getOffsetY());
+          template.add("controlOffset", controlOffset);
+        }
+        case ConeTemplate ct -> template.addProperty("direction", ct.getDirection().toString());
+        case WallTemplate wt -> {
+          if (wt.getPath() != null) {
+            JsonArray pathInfo = new JsonArray();
+            for (CellPoint cp : wt.getPath()) {
+              ZonePoint zp = map.getGrid().convert(cp);
```
kwvanderlinde: We should be careful here since this conversion is not correct on hex grids. Granted, wall templates are already completely broken on hex grids, but there's extra sensitivity when exposing information to macros.

My suggestion is to write out the path list as `CellPoint` coordinates and not bother converting to `ZonePoint` at all.

src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
```diff
@@ -260,19 +368,78 @@ protected JsonObject getDrawingJSONInfo(String functionName, Zone map, GUID guid
...
+    dinfo.addProperty(
+        "isTemplate",
+        el.getDrawable() instanceof AbstractTemplate ? BigDecimal.ONE : BigDecimal.ZERO);
+    if (el.getDrawable() instanceof AbstractTemplate at) {
+      JsonObject template = new JsonObject();
+      template.addProperty("radius", at.getRadius());
+      JsonObject vertex = new JsonObject();
+      vertex.addProperty("x", at.getVertex().x);
+      vertex.addProperty("y", at.getVertex().y);
+      template.add("vertex", vertex);
```
kwvanderlinde, line 384: I'm unsure about exposing precise template vertex positions to macros. Even though the vertex (and for line templates, the path vertex) are `ZonePoint`, they're constrained to cell boundaries since templates are cell-oriented. So from a user perspective, equivalent cell coordinates seem more natural.

There are ways to get templates off-grid, but I don't believe there's any intended way of doing so. E.g., we can adjust the grid size after drawing a template, and since vertex positions aren't updated we end up with off-grid templates. But I think it's pretty clear this isn't intended.

On the other hand, I can see a desire to support a future where templates don't have to be locked to the grid despite being such a grid-oriented feature ... Thoughts?

kwvanderlinde (changes requested):
I've still got to go through about half of the new functions yet, but I'll have to pick it up again another day. Here's my thoughts for the time being.

src/main/java/net/rptools/maptool/model/drawing/BlastTemplate.java
Baaaaaz: Done.

src/main/java/net/rptools/maptool/model/drawing/BlastTemplate.java
Baaaaaz: Done.

src/main/java/net/rptools/maptool/model/drawing/ConeTemplate.java
Baaaaaz: Done.

src/main/java/net/rptools/maptool/client/functions/TemplateFunctions.java
Baaaaaz: Done.

src/main/java/net/rptools/maptool/client/functions/TemplateFunctions.java
Baaaaaz: I did originally have for example "Wall" as the enum value and "w" as the abbreviation, but went with the `...Template` enum values as this is what gets returned in the `getDrawingInfo()` function.  Saying that I agree that unlikely anyone will use the long form in practice so have removed, and if there is a need for an alias (such as games system specific names or abbreviations) they could always be added at a later point.

src/main/java/net/rptools/maptool/client/functions/TemplateFunctions.java
Baaaaaz: Done.

src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
Baaaaaz: I sat on the naughty step for 5 minutes and have now put `"templateSize"` back in, also have excluded radius for wall templates as per your suggestion.

src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
Baaaaaz: You got me there.  I cannot recall.  Have removed.

src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
Baaaaaz: Whatever the issue was why that `if` block was there for, I cannot recall or replicate why, so have removed it.

src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
Baaaaaz: Done.

src/main/java/net/rptools/maptool/client/functions/DrawingFunctions.java
Baaaaaz: Also, users can draw (and now move) templates on gridless maps, so while their shape is cell based their position is not (I'm avoiding mentioning other issues with templates on gridless map here though!).

The reason for exposing vertex positions (either here or in the new `getTemplateX()` and `getTemplateY()` functions) was to enable relative movement of the template using the `moveTemplate()` function.  Similar to the existing `moveToken()` and the new `moveDrawable()` functions, but instead moving the template's vertex.

kwvanderlinde: Good point on gridless maps. They don't really do "cells" the way other grids do.

kwvanderlinde (approved):

PR #5892, opened 2025-11-28
Respect UI theme in token properties management panel, and other tweaks

### LINKED ISSUE #5888: [Bug]: Token Properties table in Campaign Properties does not render correctly
### Describe the Bug

In the **Campaign Properties** dialog, *Token Properties* tab, the property table does not follow the theme, nor is it consistent within itself. Some headers are coloured like cells and vice versa, and there are coloured lines between rows that do not come from the current theme.

### To Reproduce

1. Open a new campaign
2. Open the *Campaign Proeprties* dialog, and go to the _Token Properties_ tab.
3. Select the _Basic_ token type.
4. Look at the property table.

### Expected Behaviour

The property table should be more consistent with itself and with the applied theme:
- Header cells should be coloured according to the current theme, consistent with one another and distinguished from regular cells.
- Regular cells should be rendered according to the current theme, 
- Table lines should match the theme.

### Screenshots

This is the dialog on 1.18.5 with the "Cyan Light" theme:
Image

It has changed in dev builds so that the cells within a column are now more consistent. But the colouring of the GM column is still inconsistent and the header text is generally suffering. Also, the core issue of not respecting the theme is still present:
Image

### MapTool Info

1.18.5, nightly-20251127

### Desktop

Linux Mint 22.2

### Additional Context

_No response_

kwvanderlinde:

Fixes #5888 

### Description of the Change

Makes some changes to `TokenPropertiesManagementPanel` so that the theme is better respected:
1. No more custom grid lines, instead using the default colours and intercell spacing.
2. Header cells use a header cell renderer instead of a regular cell renderer (`DefaultTableCellRenderer`), so that they are rendered according to the theme. This requires decorating the default renderer (done via `TableCellRendererDecorator`) since Java does not expose the header cell renderer implementation.
3. No more custom cell colours, in particular alternating column colours have been removed.

Some other UI tweaks:
1. The one label for the list of token types now says "Token Types" instead of "Token Type", to avoid confusion with the label for settings of a particular token type.
2. The "Help" button is now toggles when the help text is shown, instead of being a plain button with no toggle state.
3. Remove the multi-line header text / HTML in favour of plain text. Although this expands the widths of the checkbox columns and the short name column, there is still room for everything, especially with …
4. Reorganized the panel.
    - The type list no longer spans the entire panel top to bottom, but is in the top row alongside the token type settings. To make this fit comfortably, I've replaced the textual buttons for the type list with iconified alternatives, similar to the _Tables_ window.
    - The properties table is now on its own row that can span the entire width of the panel, gaining much needed horizontal space.
    - The help text has been relocated to the bottom so that the other elements don't have to move around when the help text is shown or hidden.

The total size of the panel is essentially the same as before. I've tested it on resolutions as low as 800x600, and for most themes the panel fits comfortably without the dialog needing to scroll. For a theme like **Ahh (Very Large Print)**, some scrolling is need to access the right-most parts of the panel, but this was already the case.

### Possible Drawbacks

The table columns are now equally spaced. Since everything fits, I'm not too worried about it, but it also means that the checkbox columns and short name columns take up more space than they need to.

### Documentation Notes

Old panel (from Cyan Light theme on nightly-20251128):
image

New panel:
image

### Release Notes

- Modified the *Token Properties* tab of the **Campaign Properties** dialog to better follow the theme and to make better use of the screen real estate.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5892)


FullBleed:
I like the improvements, but I have a lot of Token Types (43 and growing).  Is that Token Types window fixed height or does it size at all when you stretch that dialog?

kwvanderlinde:
> I like the improvements, but I have a lot of Token Types (43 and growing). Is that Token Types window fixed height or does it size at all when you stretch that dialog?

Wow, that certainly is a lot!

The types list _is_ fixed height, but it does scroll:
image
So I don't think it will be much of a hit to usability. Though I could see that it could get tedious if you're doing a lot of back and forth changing and testing a property that is "below the fold".

We could tweak the amount of vertical space allotted to each row - the properties tables probably doesn't need that much vertical space. At the same time, though, I imagine properties outnumber property types in most cases, so it is better to optimize for a larger properties table vs a larger property type list.

kwvanderlinde:
@FullBleed I've added adjustability, and modified the default allotment of the rows. The top row now has a bit more space by default, enough to show 11 property types vs just 8. And the divider in the middle can be dragged to view more types or more properties as desired.

image

FullBleed:
> @FullBleed I've added adjustability, and modified the default allotment of the rows. The top row now has a bit more space by default, enough to show 11 property types vs just 8. And the divider in the middle can be dragged to view more types or more properties as desired.

Perfect solution.

Thanks!

cwisniew (approved):

PR #5911, opened 2025-12-13
onChangeSelection update

### LINKED ISSUE #5079: [Bug]: onChangeSelection event in HTML5 can get triggered twice unnecessarily
### Describe the Bug

If you change token selection to a single new unselected token by clicking on the token using the mouse left-hand button, the onChangeSelection event gets triggered twice: once on mouse left-hand button down and again on either mouse left-hand button up or mouse drag.

### To Reproduce

1. Create a blank campaign.
2. Create some tokens, one being a lib token.
3. Add the following macro to the lib token and run it:
```
[h, if(getSelected()==""): broadcast("NO TOKENS SELECTED") ; broadcast(getSelected())]
[h: oCS = macroLinkText(getMacroName() + "@" + getMacroLocation(), "none")]
[frame5("TEST"), code: {
	<head>
		<link rel='onChangeSelection' type='macro' href='[r: oCS]'>
	</head>
	<body>
		[r: getSelected()]
	</body>
}]
```
5. Try selecting tokens as described in the bug description.
6. Observe the triggering twice issue where the message is broadcasted on mouse down, and again on mouse up or mouse drag.

### Expected Behaviour

onChangeSelection should only get fired the once in this scenario.

### Screenshots

image

### MapTool Info

1.15.2

### Desktop

Windows 11

### Additional Context

The name of the onChangeSelection event is a bit misleading, however the wiki does state that the event is fired when a token is selected (rather than when the token selection is actually changed!).

e.g. if you left mouse click on a single token that is also the sole currently selected token, onChangeSelection will also be triggered on left mouse up (i.e. unnecessarily as the selection has not effectively changed).  Right mouse click on an already selected token does not trigger it.  Similarly if no tokens are selected. and you left click somewhere else on the map (that is not a token), onChangeSelection is triggered but again the token selection has not effectively changed.

Jmr3366:
fixes #5079 

### Description of the Change
The is a large update `SelectModel`.

### Possible Drawbacks

This changes the behavior of `onChangeSelection` to only fire when a token selection change is made.
Previously, `onChangeSelection` would miss-fire multiple times on the same token.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5911)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/zone/SelectionModel.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/SelectionModel.java
@@ -69,7 +69,7 @@ public class SelectionModel {
 
     // limit the history to a certain size
     if (selectionHistory.size() > maxHistoryLength) {
-      selectionHistory.subList(maxHistoryLength, selectionHistory.size() - 1).clear();
+      selectionHistory.subList(maxHistoryLength, selectionHistory.size()-1).clear();
     }
   }
   private boolean selectionEquals(Set<GUID> a, Set<GUID> b) {
@@ -81,7 +81,6 @@ public class SelectionModel {
   private void selectionChanged(Set<GUID> previousSelection) {
     boolean selectionDifferent = !selectionEquals(previousSelection, currentSelection) || currentSelection.isEmpty();
     boolean zoneChanged = !zone.getId().equals(lastSelectionZoneId);
-
     if (selectionDifferent || zoneChanged) {
       lastSelectionZoneId = zone.getId();
       new MapToolEventBus().getMainEventBus().post(new SelectionChanged(zone));
--- a/src/main/java/net/rptools/maptool/client/ui/zone/SelectionModel.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/SelectionModel.java
@@ -82,13 +82,10 @@ public class SelectionModel {
   /** Fire a SelectionChanged event. */
   private void selectionChanged(Set<GUID> previousSelection) {
     long now = System.currentTimeMillis();
-
     boolean selectionDifferent =
             !selectionEquals(previousSelection, currentSelection) || currentSelection.isEmpty();
-
     boolean timeElapsed =
             (now - lastSelectionChangedTime) >= SAME_SELECTION_REFIRE_DELAY_MS;
-
     if (selectionDifferent || timeElapsed) {
       lastSelectionChangedTime = now;
       new MapToolEventBus().getMainEventBus().post(new SelectionChanged(zone));
--- a/src/main/java/net/rptools/maptool/client/ui/zone/SelectionModel.java
+++ b/src/main/java/net/rptools/maptool/client/ui/zone/SelectionModel.java
@@ -69,17 +69,20 @@ public class SelectionModel {
 
     // limit the history to a certain size
     if (selectionHistory.size() > maxHistoryLength) {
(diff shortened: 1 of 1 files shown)


Jmr3366:
At the moment, this will not fire on a selected token after you change maps.  I'd like to see if a developer could help determine if this is possible.

cwisniew:
> fixes #5079
> 
> ### Description of the Change
> The is a large update `SelectModel`.
> 
> ### Possible Drawbacks
> This changes the behavior of `onChangeSelection` to only fire when a token selection change is made. Previously, `onChangeSelection` would miss-fire multiple times on the same token.
> 
> This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5911)

Have you tested this change to see if it still works the same way (minus duplicate events) when the user drag-selects the same token(s) that are already selected? Or if they use a macro to select the same token(s) that are already selected? i.e. You should get the same events before and after the change (without duplicates)

Jmr3366:
> Have you tested this change to see if it still works the same way (minus duplicate events) when the user drag-selects the same token(s) that are already selected? Or if they use a macro to select the same token(s) that are already selected? i.e. You should get the same events before and after the change (without duplicates)

This [d2f9315](https://github.com/RPTools/maptool/pull/5911/commits/d2f93158c7bf27f72e0e251ef5ba2b9c29e43ca1) ,  does not work the way anyone would want but does in fact do what onChangeSelection is supposed to do, it will only fire when a selection changes,  You can't reselect the same token to trigger, a selection change of any kind is _required_ (no body would want this compared to what oCS does _today_).  The zone changed check will not achieve the desired end result and will not _work_.

This [d3b7f18](https://github.com/RPTools/maptool/pull/5911/commits/d3b7f18f24b8800a1e0845a0246db076e6f3bb14) , I believe is what would be expected for oCS.  It will only register a selection change when a selection change takes place, OR, after 2s of time have lapsed from a prior selectionChange.  I believe this is important to retain a non-bugged version of token re-selection that the current oCS allows.   A remaining issue I'd like to overcome is if you spam mouse-clicks on different tokens, _right now_  a 2s delay is being _enforced_ (not acceptable).  I'd like the 2s delay to only be enforced with the selection is the same.

You can drag box select, use `selectTokens`, add or remove tokens from a selection, all will trigger the event ONCE.  The only issue I see is to find a better way to handle the forced delay when you _want_ to change selection.  I'm going to work on the code to only enforce the 2s delay IF the selection is the same as before.

cwisniew:
If that is the case it  breaks backwards compatibility with macros. People may be depending on that behaviour. You would need to maintain that behaviour for the change to be included. Introduction of a timer to distinguish on change selection is also not really going to work either.

Jmr3366:
> If that is the case it breaks backwards compatibility with macros. People may be depending on that behavior. You would need to maintain that behavior for the change to be included

I've just fixed the code for the 2s enforced delay ONLY if the selection is the same.

I'd like to say that _covers everything_

The timer isn't being used to distinguish selectionChange, is a _break_ to mitigate "reselection of the same thing".
(or maybe I'm missing your point).

edit
Now if you're saying script writers have code where they're wanting oCS to fire on the same token in under 2s, yes, that would fail.
edit x2
I've changed the time check to 300ms and it's working as _I'm expecting_

Jmr3366:
The original state of SelectionModel could trigger selectedChanged() as little as 4 times but could trigger upwards to 10+.
This change minimizes the selectedChanged() trigger to _once_ when a _meaningful_ change takes place.  Meaningful should be any token selection change OR 250ms _after_ any prior token selection.

PR #5918, opened 2026-01-03
createAsset improvements in error checking and capabilities.

### LINKED ISSUE #5917: [Bug]: createAsset and base64 images
### Describe the Bug

The current implementation of createAsset uses a string variables in hopes to determine the image type.
When it comes to raw base64 images, this check fails.

You can't take a base64 encoded valid image string and use it with createAsset

### To Reproduce

In a macro, do this:
```
[h: testB64 ="iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mP8/x8AAwMCAO+XnWQAAAAASUVORK5CYII="]
[r: asset = createAsset("baselineTest", testB64)]
```
This error is returned:
`java.lang.NullPointerException: Cannot invoke "String.equalsIgnoreCase(String)" because the return value of "java.net.URI.getScheme()" is null error executing expression asset = createAsset("baselineTest", testB64).`

### Expected Behaviour

I should be able to take any valid base64 image and use it with createAsset.

### Screenshots

_No response_

### MapTool Info

1.18.6

### Desktop

Win 11

### Additional Context

_No response_

Jmr3366:
fixes #5917

### Description of the Change

createAsset was using strings to check for image format and could fail.  This is now using bytes.
You could not take a base64 encoded PNG string (but base64 encoded webp would work??) and use it with createAsset.

### Possible Drawbacks
There should be none.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5918)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/TokenImage.java
+++ b/src/main/java/net/rptools/maptool/client/functions/TokenImage.java
@@ -65,9 +65,6 @@ public class TokenImage extends AbstractFunction {
   public static final String SET_IMAGE = "setImage";
   public static final String SET_PORTRAIT = "setTokenPortrait";
   public static final String SET_HANDOUT = "setTokenHandout";
-  public static final String FILE_HEADER_WEBP = "RIFF";
-  public static final String FILE_HEADER_JPG = "ÿØÿà";
-  public static final String FILE_HEADER_PNG = "‰PNG";
 
   private TokenImage() {
     super(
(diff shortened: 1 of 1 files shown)


cwisniew (changes requested):
Instead of adding the coding to try determine what type of asset is in the base64 string, convert the base64 string to a byte array and use

`Asset.createAssetDetectType(String name, byte[] data)` This way the function will support any image asset MT supports now or in the future without having to constantly update/fix this function

Then you can use `asset.getType()` to determine if it is an image or text, and check if it is a URI if it is text

PR #5923, opened 2026-01-17
Fix for localised string being unhelpful in shape.getProperties()

### LINKED ISSUE #5922: [Bug]: shape.getProperties return depends on UI localization
### Describe the Bug

Current realization of `shape.getProperties(shapeName)` returns a StrProp with data about the shape `"shapeName"`.
But the layer name in field `layer` matches not the English version of the layer name (Token/Object/Hidden/Background) or possible values of layer name for `shape.create` (Token/Object/GM/Background), but the name of the layer in the UI localization of MapTool. For example if user switches language to Russian, then he will get `Токены/Объекты/Скрытое/Фон` (their current translation into Russian).

This will broke scripts that automatically process shapes in cache using `layer` field, because string comparison result will depend on used localization.

### To Reproduce

Run one of the following code
1. `[r: json.indent(shape.getProperties(shape.create("square", "Token", true, "rectangle", "x=0;y=0;w=10;h=10",";")),3)]`
2. `[r: json.indent(shape.getProperties(shape.create("square", "GM", true, "rectangle", "x=0;y=0;w=10;h=10",";")),3)]`

For Russian/English localization the `layer` field in the output is equal to:
1. `"Токены"`/`"Token"`
2. `"Скрытое"`/`"Hidden"`

### Expected Behaviour

Layer name returned through `shape.getProperties()` should be hardcoded in English and match possible layer name for `shape.create` function ("Token"/"Object"/"GM"/"Background").

### Screenshots

_No response_

### MapTool Info

1.18.3

### Desktop

Windows

### Additional Context

_No response_

bubblobill:
fixes #5922
closes #5922

## Problem
Stems from shape.properties() calling _toString()_ on **ShapeDrawable**
https://github.com/RPTools/maptool/issues/5922#issuecomment-3762888092

##Fix
- Created new methods 
  - _ShapeDrawable.toString(boolean localisation)_, and
  - _AbstractDrawing.toString(boolean localisation)_ 
- On `localisation = FALSE`, Zone.Layer.name() is used instead of Zone.Layer.value()
- Changed _ShapeDrawable.toString()_ & _AbstractDrawing.toString()_ to call _toString(true)_ instead, i.e. defaults to true, leaving previous uses unaffected.

Token used for testing [lib_shapeTest.rptok.zip](https://github.com/user-attachments/files/24686159/lib_shapeTest.rptok.zip)

### Possible Drawbacks
none

### Documentation Notes
Fixed localisation string issue with shape.properties()

### Release Notes
Fixed localisation string issue with shape.properties()

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5923)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/ShapeFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/ShapeFunctions.java
@@ -614,7 +614,7 @@ public class ShapeFunctions extends AbstractFunction {
               seg, coords[0], coords[1], coords[2], coords[3], coords[4], coords[5], coords[6]));
       pi.next();
     }
-    StringBuilder stringBuilder = new StringBuilder(sd.toString());
+    StringBuilder stringBuilder = new StringBuilder(sd.toString(false));
     stringBuilder.append("segments=").append(String.join(",", segments)).append(";");
 
     if (delimiter.equalsIgnoreCase("json")) {
--- a/src/main/java/net/rptools/maptool/model/drawing/AbstractDrawing.java
+++ b/src/main/java/net/rptools/maptool/model/drawing/AbstractDrawing.java
@@ -145,9 +145,17 @@ public abstract class AbstractDrawing implements Drawable, ImageObserver {
 
   @Override
   public String toString() {
+    return toString(true);
+  }
+
+  public String toString(boolean localised) {
     StringBuilder sb = new StringBuilder();
     sb.append("name=").append(getName()).append(";");
-    sb.append("layer=").append(getLayer()).append(";");
+    if(localised) {
+      sb.append("layer=").append(getLayer()).append(";");
+    } else {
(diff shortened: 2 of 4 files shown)


cwisniew (changes requested):
Its clearer if you create a new function and have
- toString()
- toLocalisedString()

Having just a boolean variable makes it harder to follow unless its obvious what the value represents (e.g. if the function is something like `setVisible`)

bubblobill:
> Its clearer if you create a new function and have
> 
> * toString()
> * toLocalisedString()
> 
> Having just a boolean variable makes it harder to follow unless its obvious what the value represents (e.g. if the function is something like `setVisible`)

Well, I could just go through every Enum and remove the override on toString and make sure every UI call specifically called getDisplayName(), but this seemed easier.

bubblobill:
@cwisniew Done now

cwisniew (approved):

PR #5926, opened 2026-01-19
Fix Drawing ID not unique on shape.draw()

### LINKED ISSUE #5920: [Bug]: shape.draw can produce drawings with the same id
### Describe the Bug

Function shape.draw executed to draw the same shape many times each time produces a drawing with the same ID.
Image
Due to this produced drawing are treated by MapTool as the same drawing and, for example, an attempt to move drawing to a different layer "deletes" all extra copies, so only one instance remains.

### To Reproduce

1. Execute
```
[h: shape.create("square", "Token", true, "rectangle", json.set("", "x", 300, "y", 300, "w", 150, "h", 150))]
```
to create the shape "square" in cache.
2. Execute
```
[h: pen ="width=8; foreground = white; background = transparent; squareCap=1; opacity = 0.3;eraser=false;"]
[r: shape.draw("square",getCurrentMapName(), pen)]
```
twice to create two drawings, named "square". `shape.draw` returns the drawing id, so you already can see in chat the same id written twice.
3. Move one of the drawings to any layer. After that the second one will disappear.

### Expected Behaviour

Each drawing generated by `shape.draw` should have a unique id that doesn't depend on the shape name, used to generate it.

### Screenshots

_No response_

### MapTool Info

1.18.3

### Desktop

Windows

### Additional Context

_No response_

bubblobill:
fixes #5920
closes #5920

### Description of the Change
Changed `AbstractDrawing(AbstractDrawing other)` to generate a new ID.
Changed `shape.draw()` to use `... drawnElement = new DrawnElement(shapeDrawable.copy()...` to enforce a unique ID.

Test results for the same shape drawn ten times;
**lib:shapeTest:** 
1. Drawing Id: 3D6EF77ED31245029CD3D14B07879DC7
2. Drawing Id: 075888047DB34FF09810E495E09B8C3A
3. Drawing Id: 34476124167A40E7BEAE19BF117BB958
4. Drawing Id: BCC4D776184B4FA7A490DF8EBF16D372
5. Drawing Id: F3ABCFCD9AA74D9CA381A01373DCDAAE
6. Drawing Id: 45BD6B9CC88A45B6B0189C8D55CF827C
7. Drawing Id: C1F9FFB596224C1D9ED3E80D3FC37981
8. Drawing Id: 5FAECE5E51AC4F75AD3A86481165D9A4
9. Drawing Id: 8876E0B3BE2D474B99C9311CB56A0E31
10. Drawing Id: D53D4819EFEA4D16AB6A2964C09AB68A

### Possible Drawbacks
none

### Documentation Notes
BugFix: shape.draw() returning duplicate IDs on multiple calls for the same shape.
### Release Notes
BugFix: shape.draw() returning duplicate IDs on multiple calls for the same shape.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5926)

### DIFF
--- a/src/main/java/net/rptools/maptool/model/drawing/AbstractDrawing.java
+++ b/src/main/java/net/rptools/maptool/model/drawing/AbstractDrawing.java
@@ -47,7 +47,7 @@ public abstract class AbstractDrawing implements Drawable, ImageObserver {
 
   protected AbstractDrawing(AbstractDrawing other) {
     // The only thing we don't preserve is the ID.
-    this.id = other.id;
+    this.id = new GUID();
     this.layer = other.layer;
     this.name = other.name;
   }
--- a/build.gradle
+++ b/build.gradle
@@ -360,7 +360,7 @@ dependencies {
     implementation(libs.okhttp)
 
     implementation(libs.bundles.imageio)
-    implementation(libs.batik)
+    implementation(libs.bundles.batik)
     implementation(libs.bundles.pdfbox)
     implementation(libs.bcmail)
     implementation(libs.bundles.jai.imageio)
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -16,7 +16,7 @@ dependencies {
     implementation(libs.gson)
 
     implementation(libs.bundles.imageio)
-    implementation(libs.batik)
+    implementation(libs.bundles.batik)
     implementation(libs.jsvg)
 
     implementation(libs.apache.commons.io)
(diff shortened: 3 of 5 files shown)


cwisniew (approved):

src/main/java/net/rptools/maptool/model/drawing/AbstractDrawing.java
```diff
@@ -47,7 +47,7 @@ protected AbstractDrawing(GUID id) {
 
   protected AbstractDrawing(AbstractDrawing other) {
     // The only thing we don't preserve is the ID.
-    this.id = other.id;
+    this.id = new GUID();
```
Baaaaaz, line 50: This change has broken the drawing and template improvements in develop and due in 1.19.  i.e. 
- no longer able to drop drawings/templates to a new location after selecting and dragging them
- no longer able to resize templates
- holding down ctrl when dragging a drawing causes a NPE error.

Reverting this line back to `this.id = other.id;` seems to fix the issues listed above.

Baaaaaz: @cwisniew @kwvanderlinde @bubblobill any suggestions on how to fix, as reverting obviously then breaks this PR fix duplicate drawing IDs.

PR #5932, opened 2026-01-20
shape.properties output string fix

### LINKED ISSUE #5927: [Bug]: shape.getProperties output in json-mode is incorrectly constructed from StrProp
### Describe the Bug

If `shape.getProperties` is called with a delimiter `"json"`, then a field `bounds` will be constructed incorrectly: there will be no field `bounds` containing StrProp object with all the fields, but the field `bounds` containing a part of the StrProp until the first `";"` and all the other fields of StrProp with their value, so `shape.getProperties` output gets extra fields.

### To Reproduce

Execute
```
[h: shape.create("square", "", true, "rectangle", json.set("","x", 300,"y", 300,"w", 150,"h", 150))]
[r: "<pre>"+shape.getProperties("square")+"</pre>"]
[r: "<pre>"+json.indent(shape.getProperties("square", "json"),3)+"</pre>"]
```
and compare `bounds` field for these to outputs of `shape.getProperties`.

### Expected Behaviour

Quotes should not be ignored during construction of the JSON-object.

### Screenshots

_No response_

### MapTool Info

1.18.3

### Desktop

Windows

### Additional Context

_No response_

bubblobill:
fixes #5927
closes #5927

### Description of the Change
Altered toString methods on ancestor classes to flatten structure.
Added toJson methods on ancestor classes to prevent things getting munted in conversion to JSON.
Unable to elegantly resolve nested string lists without the addition of another delimiter, so there are now three delimiter arguments available on the function
0. Shape Name
1. "json" or String-Property-List delimiter - default `;`
2. String-List delimiter - default `,`
3. Nested String-List delimiter, default `:`

### Possible Drawbacks
Users will expect to be able to access the nested string list without specifying a delimiter and will become angstful at the unexpected results.

### Documentation Notes
Fix to shape.properties output
### Release Notes
Fix to shape.properties output

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5932)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/ShapeFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/ShapeFunctions.java
@@ -605,13 +605,13 @@ public class ShapeFunctions extends AbstractFunction {
             ? ":"
             : FunctionUtil.paramAsString(functionName, parameters, 3, false);
     if (delimiter1.equals(delimiter2)) {
-      throw new RuntimeException(
+      throw new ParserException(
           I18N.getText("macro.function.general.argumentTypeB", functionName, 3, 2));
     } else if (delimiter2.equals(delimiter3)) {
-      throw new RuntimeException(
+      throw new ParserException(
           I18N.getText("macro.function.general.argumentTypeB", functionName, 4, 3));
     } else if (delimiter3.equals(delimiter1)) {
-      throw new RuntimeException(
+      throw new ParserException(
           I18N.getText("macro.function.general.argumentTypeB", functionName, 4, 2));
     }
     if (!CACHED_SHAPES.containsKey(name)) {
--- a/src/main/java/net/rptools/maptool/client/functions/ShapeFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/ShapeFunctions.java
@@ -17,16 +17,13 @@ package net.rptools.maptool.client.functions;
 // @formatter:off
 
 import com.google.common.primitives.Floats;
-import com.google.gson.JsonArray;
-import com.google.gson.JsonElement;
-import com.google.gson.JsonObject;
+import com.google.gson.*;
 import java.awt.*;
 import java.awt.geom.*;
 import java.util.*;
 import java.util.List;
 import java.util.function.BiFunction;
 import net.rptools.maptool.client.MapTool;
-import net.rptools.maptool.client.functions.json.JSONMacroFunctions;
 import net.rptools.maptool.language.I18N;
 import net.rptools.maptool.model.GUID;
 import net.rptools.maptool.model.Zone;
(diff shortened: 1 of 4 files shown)


src/main/java/net/rptools/maptool/model/drawing/ShapeDrawable.java
```diff
@@ -178,15 +180,57 @@ private Object applyAA(Graphics2D g) {
...
+        + getBounds().width
+        + ";"
+        + "height="
+        + getBounds().height;
+  }
+
+  @Override
+  public JsonObject toJson() {
+    JsonObject jo = super.toJson();
+    jo.add("antiAliasing", new JsonPrimitive(getName()));
```
cwisniew: This should be getUseAntiAliasing()

src/main/java/net/rptools/maptool/model/drawing/ShapeDrawable.java
```diff
@@ -178,15 +180,57 @@ private Object applyAA(Graphics2D g) {
...
+    jo.add("shapeType", new JsonPrimitive(getLayer().name()));
```
cwisniew: This should be getShapeTypeName()

cwisniew (changes requested):

src/main/java/net/rptools/maptool/model/drawing/ShapeDrawable.java
bubblobill: Obvious brain fart number one. Now fixed.

src/main/java/net/rptools/maptool/model/drawing/ShapeDrawable.java
bubblobill: Obvious brain fart number two. Now fixed.

src/main/java/net/rptools/maptool/model/drawing/AbstractDrawing.java
```diff
@@ -152,6 +154,14 @@ public String toNonLocalisedString() {
     return "name=" + getName() + ";" + "layer=" + getLayer().name() + ";" + "id=" + getId() + ";";
   }
 
+  public JsonObject toJson() {
+    JsonObject jo = new JsonObject();
+    jo.add("name", new JsonPrimitive(getName()));
```
cwisniew: If name can be null you will need to check it before calling `JsonPrimitive`

src/main/java/net/rptools/maptool/client/functions/ShapeFunctions.java
```diff
@@ -592,12 +590,30 @@ private Object drawShape(
...
+    if (delimiter1.equals(delimiter2)) {
+      throw new RuntimeException(
+          I18N.getText("macro.function.general.argumentTypeB", functionName, 3, 2));
+    } else if (delimiter2.equals(delimiter3)) {
+      throw new RuntimeException(
+          I18N.getText("macro.function.general.argumentTypeB", functionName, 4, 3));
+    } else if (delimiter3.equals(delimiter1)) {
+      throw new RuntimeException(
+          I18N.getText("macro.function.general.argumentTypeB", functionName, 4, 2));
+    }
```
cwisniew, line 616: These should throw `ParserException` not `RuntimeException`

src/main/java/net/rptools/maptool/model/drawing/ShapeDrawable.java
```diff
@@ -196,8 +197,7 @@ public String toString() {
         + getBounds().width
         + ";"
         + "height="
-        + getBounds().height
-        + "\";";
```
cwisniew, line 200: Previously this was all under `bounds` key, this change will break backwards compatability. Please reconsider :)

cwisniew (changes requested):

src/main/java/net/rptools/maptool/model/drawing/AbstractDrawing.java
bubblobill: done

src/main/java/net/rptools/maptool/client/functions/ShapeFunctions.java
bubblobill: Whoops. Fixed.

src/main/java/net/rptools/maptool/model/drawing/ShapeDrawable.java
bubblobill: Disagree but willing to compromise.
"bounds" is a nested string property list. As such it cannot be accessed using getStrProp.
Compromise is to leave the old value as was but keep the new flattened structure which actually allows access to the bounds values.

cwisniew (approved):

PR #5934, opened 2026-01-29
json.toHtmlTable function

### LINKED ISSUE #5906: [Feature]: A new jsonToHtmlTable function
### Describe the Problem

Having to construct different html tables or forms based on different json schemas.

### The Solution you'd like

Provide a native jsonToHtmlTable function to recursively convert json objects and arrays into nested html tables.

Option to pivot arrays of objects and objects of objects, so similar nested structures can be tabularized.

Option to show content as cell values or as html inputs which could be embedded in an html form.

Continue to develop and use this MTScript equivalent and suffer the slow performance with large and complex json datasets: https://discord.com/channels/296230822262865920/1401537394460725328

### Additional Context

Image

Baaaaaz:
closes #5906

### Description of the Change
New MTScript function `json.toHtmlTable(json [, options])` which recursively transforms json into nested html tables with applied options.

### Possible Drawbacks
None foreseen, but added experimental flag while it matures.

### Documentation Notes
To be done, but options are (in no particular order):

-   LeadObjectKeys - Return a specific order of object keys (if present) to start with
-   RearObjectKeys - Return a specific order of object keys (if present) to end with
-   SortObjectKeys - Return object keys in a sorted order (for keys not defined as lead and rear object keys)
-   Attributes - Sets any html attributes on the root table
-   Caption - Adds a caption element to the root table
-   Collapsible - Whether to output html tables with a detail/summary wrapper
-   CollapsibleOpenDepth - Whether a detail/summary wrapper is open by default
-   Titles - Whether to output the json path as html title attributes
-   ArrayOfObjects - Whether to pivot json objects within a json array
-   ObjectOfObjects - Whether to pivot json objects within a json object
-   AssetImage - Whether json values starting with "assetid://" should returned as an html `` element
-   EscapeHtml - Whether json values containing certain html characters in should be escaped.
-   SanitizeHtml - Whether to sanitize the html returned from jsonToHtmlTable.
-   Input - Whether json values should be converted to input elements for use in a form (defaults to false)

### Release Notes
Add new MTScript function `json.toHtmlTable()`

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5934)


Baaaaaz:
An example MTScript is:

```
[h: vJson = getInfo("map")]

[h: vOptions = json.set("{}", 
	"caption", "Example Data",
	"attributes", json.set("{}", "id","exampleId", "class","exampleClass"),
	"collapsible", 1,
	"collapsibleOpenDepth", 2,
	"escapeHtml", 0,
	"sanitizeHtml", 1,
	"assetImage", 1,
	"arrayOfObjects", 1,
	"objectOfObjects", 1,
	"leadObjectKeys", json.append("", "id", "name", "level"),
	"sortObjectKeys", "asc",
	"rearObjectKeys", json.append("", "source", "page"),
	"input", 0
)]

[frame5("INFO"): {
	<html>
		<head>
			<link rel='stylesheet' type='text/css' href='lib://JsonHtmlTable/macro/style.css?cachelib=false'>
			<style>
				[r:'
					span.json-array, th.json-array {background-color: LightSkyBlue;}
					span.json-object, th.json-object {background-color: DarkSeaGreen;}
					[data-json-key="name"] {color: red}
					[data-json-index="1"] {background-color: yellow}
				']
				</style>
		</head>
		
		<body>
			[r: json.toHtmlTable(vOptions, json.set("{}", "caption", "json.toHtmlTable Options"))]
			<hr>
			[r: json.toHtmlTable(vJson, vOptions)]
			<hr>
			[r: replace(json.toHtmlTable(vJson, vOptions), "<", "&lt;")]			
		</body>
	</html>
}]
```

image

Baaaaaz:
Or another example using jmrs 5E spell library json:
image

Baaaaaz:
Ready for a review :)

src/main/java/net/rptools/maptool/client/functions/json/JsonHtmlFunctions.java
```diff
@@ -0,0 +1,906 @@
...
+import java.math.BigDecimal;
+import java.util.*;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.jsoup.Jsoup;
+import org.jsoup.safety.Safelist;
+
+public class JsonHtmlFunctions {
+
+  JsonHtmlFunctions(JsonMTSTypeConversion converter) {
```
cwisniew: Can you add a @Nonnull annotation to the argument to help IDEs pick up potential errors

src/main/java/net/rptools/maptool/client/functions/json/JsonHtmlFunctions.java
```diff
@@ -0,0 +1,906 @@
...
+  /** SortMethod enumeration */
+  private enum SortMethod {
+    NONE,
+    REVERSE,
+    ASCENDING,
+    DESCENDING,
+  }
+
+  /** Track the current depth in the json hierarchy during the conversion to html table(s) */
+  private Integer jsonPathDepth;
```
cwisniew: Since you are not using `null` as a "not defined" value this should be just an `int` to avoid all the auto boxing/unboxking every time you use it.

src/main/java/net/rptools/maptool/client/functions/json/JsonHtmlFunctions.java
```diff
@@ -0,0 +1,906 @@
...
+            ? !options.get("assetimage").getAsBigDecimal().equals(BigDecimal.ZERO)
+            : true;
+    optionEscapeHtml =
+        options.has("escapehtml")
+            ? !options.get("escapehtml").getAsBigDecimal().equals(BigDecimal.ZERO)
+            : true;
+    optionSanitizeHtml =
+        options.has("sanitizehtml")
+            ? !options.get("sanitizehtml").getAsBigDecimal().equals(BigDecimal.ZERO)
+            : true;
```
cwisniew: Because all of this is user input you need to check they passed the correct type and raise a meaningfull error if they didn't, rather than the JSON library throwing a vague error they would not be able to debug. 

Especially since the problems with MT Json not being able to use booleans have been addressed many versions ago, I would expect to be able to use boolean true/false instead of 1/0 in the options.

cwisniew (changes requested):

src/main/java/net/rptools/maptool/client/functions/json/JsonHtmlFunctions.java
Baaaaaz: Done

src/main/java/net/rptools/maptool/client/functions/json/JsonHtmlFunctions.java
Baaaaaz: Done

src/main/java/net/rptools/maptool/client/functions/json/JsonHtmlFunctions.java
Baaaaaz: Can take either boolean, or 1/0.  Also added some option value validation and relevant exception messages.

cwisniew (approved):

PR #5940, opened 2026-02-14
Add VBL Pen tool for freehand topology drawing

### LINKED ISSUE #5939: [Feature]: VBL Pen tool for freehand topology drawing
### Describe the Problem
Currently, drawing Vision Blocking Layer (VBL) or other topology types is limited to geometric shapes (Rectangles, Ovals, Polygons) or line-based tools. These tools often require multiple clicks and only apply the final shape once the drawing is completed (mouse release). This makes it difficult to draw complex, organic, or fine-grained VBL areas quickly and with real-time feedback.

### The Solution you'd like
Add a "VBL Pen" tool that allows for freehand drawing of topology with a configurable radius. Key features should include:
- **Freehand Drawing:** Continuous drawing while the left mouse button is held down.
- **Immediate Application:** Topology changes should be applied incrementally during the mouse drag, providing real-time visual feedback on the map.
- **Configurable Radius:** A dedicated radius control panel and support for dynamic resizing using `Ctrl + Mouse Wheel`.
- **Eraser Mode:** Instant switching to eraser mode using the `Shift` key, with the cursor indicator changing color to reflect the mode.
- **Support for All Topology Types:** The tool should respect the currently selected modes in the Topology Toolbar (Wall VBL, MBL, Hill VBL, etc.).

Using the Polygon tool or drawing many small rectangles, but these are inefficient and lack the fluid experience of a pen-style tool.

### Additional Context
I have already developed a prototype implementation of this feature which includes:
- `VBLPenTool.java` for the core logic.
- Real-time incremental application of topology segments.
- Integration with `AppStatePersisted` for radius persistence.
- Immediate visual feedback for modifier keys (Shift for Erase).
- Integration into the Topology toolbar using the Freehand icon.

KnutZuidema:
closes #5939

### Description of the Change
This PR implements a freehand VBL Pen tool designed for high-responsiveness topology drawing. The tool employs an incremental application strategy, submitting individual area segments during mouse drag events to provide real-time visual feedback across all selected topology layers. It includes a dedicated radius configuration panel, supports dynamic resizing via Ctrl+Mouse Wheel, and provides immediate visual mode indicators by tracking modifier key states.

### Possible Drawbacks
Frequent network updates during continuous drawing, though mitigated by segmenting the area updates.

### Documentation Notes
Adds VBL Pen tool to the topology toolbar. Supports Ctrl+Scroll for radius and Shift for eraser mode.

### Release Notes
Added a VBL Pen tool for real-time freehand drawing of topology with a configurable radius and Ctrl+Scroll support.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5940)


cwisniew:
I will leave this one to @kwvanderlinde  to review as he is working on VBL updates/changes and also the new renderer and a large path may be preferable to many small areas. Might be a bit of wait though as currenttly real life has interfered with his MapTool coding time.

src/main/java/net/rptools/maptool/client/tool/drawing/VBLPenTool.java
```diff
@@ -0,0 +1,272 @@
...
+    MapTool.getFrame().removeControlPanel();
+    super.detachFrom(renderer);
+  }
+
+  @Override
+  protected void resetTool() {
+    lastPoint = null;
+    shiftDown = false;
+    renderer.repaint();
+    super.resetTool();
```
kwvanderlinde: To be consistent with other tools: if the user hits <kbd>Escape</kbd> while drawing, we should discard what hasn't been committed yet and keep this tool active. We should only call `super.resetTool()` if not in the process of drawing.

src/main/java/net/rptools/maptool/client/tool/drawing/VBLPenTool.java
```diff
@@ -0,0 +1,272 @@
...
+
+    if (lastPoint != null) {
+      BasicStroke stroke =
+          new BasicStroke(thickness, BasicStroke.CAP_ROUND, BasicStroke.JOIN_ROUND);
+      Path2D line = new Path2D.Double();
+      line.moveTo(lastPoint.x, lastPoint.y);
+      line.lineTo(point.x, point.y);
+      segmentPath.append(stroke.createStrokedShape(line), false);
+    }
+    submit(segmentPath);
```
kwvanderlinde: _Relates to [my comment on `resetTool()`](https://github.com/RPTools/maptool/pull/5940/changes#r3495277770)_

We shouldn't submit changes until the user can committed by releasing the mouse. Until that point, we should only stage the changes. This allows the user to change their mind by hitting <kbd>Escape</kbd>, discarding any changes.

src/main/resources/net/rptools/maptool/language/i18n.properties
```diff
@@ -2960,6 +2961,8 @@ tool.ovalexpose.tooltip          = Expose/Hide an oval on the Fog of War.
 tool.ovaltopology.instructions   = LClick: set initial/final point, Shift+LClick: start erase oval
 tool.ovaltopology.tooltip        = Draw an oval VBL.
 tool.ovaltopologyhollow.tooltip  = Draw a hollow oval VBL.
+tool.vblpen.instructions        = LClick+Drag: draw VBL with configurable radius; Shift+LClick+Drag: erase; Ctrl: snap-to-grid
```
kwvanderlinde: There is no snap-to-grid functionality with the VBL Pen. Not sure if this is an oversight in the pen or in the instructions.

src/main/java/net/rptools/maptool/client/tool/drawing/VBLPenTool.java
```diff
@@ -0,0 +1,272 @@
...
+  @Override
+  public void mouseMoved(MouseEvent e) {
+    super.mouseMoved(e);
+    shiftDown = e.isShiftDown();
+    renderer.repaint(); // Repaint to update the pen preview.
+  }
+
+  @Override
+  public void mouseWheelMoved(MouseWheelEvent e) {
+    if (net.rptools.maptool.client.swing.SwingUtil.isControlDown(e)) {
```
kwvanderlinde: I think a different modifier key would be better here, e.g., <kbd>Alt</kbd>. For other VBL tools, <kbd>Control</kbd> engages snap-to-grid functionality, and if we ever add that to the VBL Pen it would make zooming tedious.

src/main/java/net/rptools/maptool/client/tool/drawing/VBLPenTool.java
```diff
@@ -0,0 +1,272 @@
...
+    segmentPath.append(circle, false);
+
+    if (lastPoint != null) {
+      BasicStroke stroke =
+          new BasicStroke(thickness, BasicStroke.CAP_ROUND, BasicStroke.JOIN_ROUND);
+      Path2D line = new Path2D.Double();
+      line.moveTo(lastPoint.x, lastPoint.y);
+      line.lineTo(point.x, point.y);
+      segmentPath.append(stroke.createStrokedShape(line), false);
+    }
```
kwvanderlinde: This would be clearer if we didn't bother with the ellipse at all and instead always stroked a line. In the case of `lastPoint == null`, just make the two points of the line the same.

src/main/java/net/rptools/maptool/client/tool/drawing/VBLPenTool.java
```diff
@@ -0,0 +1,272 @@
...
+ *   <li>Ctrl+Wheel: Dynamic radius adjustment.
+ * </ul>
+ */
+public final class VBLPenTool extends AbstractDrawingLikeTool {
+  private final TopologyModeSelectionPanel topologyModeSelectionPanel;
+  private final TopologyTool.MaskOverlay maskOverlay;
+  private static final RadiusPanel RADIUS_PANEL = new RadiusPanel();
+
+  private ZonePoint lastPoint;
+  private boolean shiftDown;
```
kwvanderlinde: No need for `shiftDown` since `AbstractDrawingLikeTool` already provides `setIsEraser(boolean)` and `isEraser()` to track this.

src/main/java/net/rptools/maptool/client/tool/drawing/VBLPenTool.java
```diff
@@ -0,0 +1,272 @@
...
+      g2.draw(circle);
+    }
+
+    g2.dispose();
+  }
+
+  @Override
+  public void mousePressed(MouseEvent e) {
+    if (SwingUtilities.isLeftMouseButton(e)) {
+      shiftDown = e.isShiftDown();
```
kwvanderlinde: Instead of checking `e.isShiftDown()`, use `isEraser(e)` to ensure all drawing-like tools stay consistent.

kwvanderlinde (dismissed):

kwvanderlinde:
> I will leave this one to @kwvanderlinde to review as he is working on VBL updates/changes and also the new renderer and a large path may be preferable to many small areas. Might be a bit of wait though as currenttly real life has interfered with his MapTool coding time.

Just want to point out that while a path could be built, it wouldn't really be advantageous over areas in this case. If the radius were constant, it would be better to build a simple path through each point and then stroke the final result. But because the radius can change we do have to ultimately union up the various bits whether through areas or paths. If we get complaints about performance we can look at this more.

kwvanderlinde:
One thing I'm going to need to test with this is the performance hit for vision. Curves are a lot better than they used to be, but it's still one of those cases that has the potential to cause big slow downs.

kwvanderlinde:
> One thing I'm going to need to test with this is the performance hit for vision. Curves are a lot better than they used to be, but it's still one of those cases that has the potential to cause big slow downs.

Yeah, I tried it out. It doesn't take much scribbling to grind vision to a halt. We'll want to flatten the produced areas to avoid this. I'm happy to take on that part as the core functionality is otherwise good.

KnutZuidema:
@kwvanderlinde 
If you think this could be helpful to others, please feel free to take this over.

I haven't had time to do much map building, recently. 😞

src/main/resources/net/rptools/maptool/language/i18n.properties
KnutZuidema: Initially thought about adding it, but then discarded that idea, as it wasn't useful, and ran counter to the basic idea of the tool. Should be removed here

kwvanderlinde:
I've added the changes I requested. Hopefully someone else will be able to review from here on.

For performance, I flattened the areas prior to submitting them. It seems to be alright now, but we'll have to keep an eye out for it getting too bad. I don't think this tool's primary use should be to draw large, complicated areas, though.

cwisniew (approved):

Issue #5941, opened 2026-02-21
[Bug]: Drawing and Template regression


Baaaaaz:
### Describe the Bug

PR [#5926](https://github.com/RPTools/maptool/pull/5926) introduces a regression which breaks the drawing and template improvements in develop and due in 1.19. i.e.
- no longer able to drop drawings/templates to a new location after selecting and dragging them
- no longer able to resize templates
- holding down ctrl when dragging a drawing causes a NPE error.

_Originally posted in https://github.com/RPTools/maptool/pull/5926#discussion_r2807352727_

### To Reproduce

1. Run a Develop or a current Nightly Build.
2. Create a new map.
3. Create a new drawing or template.
4. Select the Drawing Pointer Tool or Template Pointer Tool.
5. Try either:
5.1 dragging a drawable and dropping -> drawable does not drop.
5.2 resizing a template -> template does not change.
5.3 holding down ctrl when dragging a drawing -> causes a NPE error.

### Expected Behaviour

Able to drag and drop drawings and templates.
Able to resize templates.
Able to drag drawings whilst snapping to grid.

### Screenshots

e.g. NPE

<img width="950" height="696" alt="Image" src="https://github.com/user-attachments/assets/0af6612c-4a15-419a-801d-396bcd103307" />

### MapTool Info

Develop or current Nightly Build

Baaaaaz:
Noting that the commit https://github.com/RPTools/maptool/commit/6ae4b6ac3fe7b1634e08ef413239a38c17ed56e6 which introduced `protected AbstractDrawing(AbstractDrawing other)` has the following comment:

> To enable this, `Drawable` now has a `copy()` method, which is consistently implemented in child classes as delegating
to each implementation's copy constructor. The copy constructor always keeps the same ID, but callers can now modify the
ID through a new method `Drawable.setId(GUID)` for those cases where the ID needs to be given a different ID from the
original.

Unknown impact of PR https://github.com/RPTools/maptool/pull/5926 on other Drawable features beyond what has been identified so far in this issue, but seems we either need to revert it, or fix it so it follows the commit guidance above, or something else...  Given `ShapeFunctions` still have an experimental warning, such things should not break non-experimental stuff!

kwvanderlinde:
@Baaaaaz Can you confirm whether the changes in #5942 have fixed this issue?

Baaaaaz:
[...]

PR #5942, opened 2026-02-27
Fix Drawing ID not unique on shape.draw() - Mk II

bubblobill:
Original #5920 
hopefully fixes [5926 comment](https://github.com/RPTools/maptool/pull/5926#discussion_r2807352727) without screwing up the original issue.

### Description of the Change
Revert to copying the id in the AbstractDrawing copy constructor; otherwise it breaks drawing user interactions.
Updated ShapeFunctions method drawShape to assign a new ID to the copied ShapeDrawable which should propagate its way to the new drawing ID.

### Possible Drawbacks
Who knows, obviously I can't be trusted.

### Release Notes

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5942)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/ShapeFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/ShapeFunctions.java
@@ -522,6 +522,8 @@ public class ShapeFunctions extends AbstractFunction {
       throw new ParserException(I18N.getText(OBJECT_NOT_FOUND, functionName, shapeName));
     }
     ShapeDrawable shapeDrawable = CACHED_SHAPES.get(shapeName);
+    shapeDrawable.setId(
+        new GUID()); // this gives the drawn element a separate ID to the cached shape
     Rectangle bounds = shapeDrawable.getBounds();
 
     /* Sanity checks */
--- a/src/main/java/net/rptools/maptool/client/functions/ShapeFunctions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/ShapeFunctions.java
@@ -521,7 +521,7 @@ public class ShapeFunctions extends AbstractFunction {
     if (!CACHED_SHAPES.containsKey(shapeName)) {
       throw new ParserException(I18N.getText(OBJECT_NOT_FOUND, functionName, shapeName));
     }
-    ShapeDrawable shapeDrawable = CACHED_SHAPES.get(shapeName);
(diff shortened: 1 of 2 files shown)


bubblobill:
@Baaaaaz This hopefully fixes your drawing issues

src/main/java/net/rptools/maptool/model/drawing/AbstractDrawing.java
```diff
@@ -46,8 +46,8 @@ protected AbstractDrawing(GUID id) {
   }
 
   protected AbstractDrawing(AbstractDrawing other) {
-    // The only thing we don't preserve is the ID.
-    this.id = new GUID();
+    // The only thing we don't preserve is the ID. - umm except that we obviously do
+    this.id = other.id;
```
cwisniew: You know you could have just deleted the comment rather than making it confusing :)

cwisniew (changes requested):

src/main/java/net/rptools/maptool/model/drawing/AbstractDrawing.java
bubblobill: I don't like deleting other people's comments which I don't understand and might serve a purpose

bubblobill: removed

src/main/java/net/rptools/maptool/client/functions/ShapeFunctions.java
```diff
@@ -522,6 +522,8 @@ private Object drawShape(
       throw new ParserException(I18N.getText(OBJECT_NOT_FOUND, functionName, shapeName));
     }
     ShapeDrawable shapeDrawable = CACHED_SHAPES.get(shapeName);
+    shapeDrawable.setId(
+        new GUID()); // this gives the drawn element a separate ID to the cached shape
     Rectangle bounds = shapeDrawable.getBounds();
```
cwisniew: This also changes the ID of the cached shape
You need to make a copy and use that if you want to change the id.

bubblobill: Fair cop. Fixed,

Baaaaaz:
@cwisniew please could you review again when you get the opportunity?  This change fixes some broken drawable UI improvements (see https://github.com/RPTools/maptool/issues/5941) which do not currently work in the nightly build or develop.  Thanks.

cwisniew (approved):

PR #5962, opened 2026-05-05
Add client color and accessibility prefs to `getClientInfo()`

### LINKED ISSUE #5961: [Feature]: MapTool theme and stat sheet CSS improvements
### Describe the Problem

Currently unable to style frame/dialog/overlay/chat content according to specific user color/accessibility preferences.

Also some html elements are not handled in the current MapTool theme files (other buttons types, other form control elements, etc), so look to improve style coverage.

### The Solution you'd like

Incorporate color/accessibility preferences into the generated MapTool theme and stat sheet CSS files, along with other CSS improvements for styling html elements similar to MapTool UI.

Expose certain user styling preferences via `getInfo("Client")`, including map label colors, arrow colors, if a dark theme is selected, etc so frames/dialogs/overlays/chat outputs can be styled individually for each user.  Some user preferences are already exposed via `getInfo("Client")` - attempted in [[PR](https://github.com/RPTools/maptool/pull/5962)] but a better way was suggested, now added as the solution I'd like above.

### Additional Context

_No response_

Baaaaaz:
closes #5961 

### Description of the Change
Add client color and accessibility prefs to `getClientInfo()`.

### Possible Drawbacks
None anticipated

### Documentation Notes
Update https://wiki.rptools.info/index.php/getInfo with expanded `getInfo("Client")` sample output.

### Release Notes
Added client color and accessibility preferences to `getInfo("Client")`

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5962)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/getInfoFunction.java
+++ b/src/main/java/net/rptools/maptool/client/functions/getInfoFunction.java
@@ -33,6 +33,7 @@ import net.rptools.maptool.client.MapToolExpressionParser;
 import net.rptools.maptool.client.ui.htmlframe.HTMLDialog;
 import net.rptools.maptool.client.ui.htmlframe.HTMLFrame;
 import net.rptools.maptool.client.ui.htmlframe.HTMLOverlayManager;
+import net.rptools.maptool.client.ui.theme.ThemeSupport;
 import net.rptools.maptool.client.ui.token.*;
 import net.rptools.maptool.client.ui.zone.renderer.ZoneRenderer;
 import net.rptools.maptool.language.I18N;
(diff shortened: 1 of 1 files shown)


cwisniew:
I am not sure this will work the way most people expect when used in chat. 
As the colouring is currently done on the sending client and not the rendering client (I know its awful). 

Aside from that, there is already a mechanism to expose the theme colours etc to the frames, these values should also go into the css so that they can be used consistantly with the other colour values.

https://github.com/RPTools/maptool/blob/develop/src/main/resources/net/rptools/maptool/library/builtin/mt-theme-css.hbs

https://github.com/RPTools/maptool/blob/develop/src/main/resources/net/rptools/maptool/library/builtin/mt-stat-sheet-css.hbs

Which are processed in 

https://github.com/RPTools/maptool/blob/develop/src/main/java/net/rptools/maptool/model/library/builtin/themecss/ThemeCssContext.java

Baaaaaz:
Okay, thanks for the guidance and links, I will revisit.

PR #5980, opened 2026-06-18
Remove support for -v option

### LINKED ISSUE #5973: [Maintenance]: Include environment and version information inside the JAR file
### Describe the problem

To get accurate sentry reporting and versioning information, we rely on users of the JAR files providing correct values for the `sentry.environment` system property and the `-v` application argument. This is not reliable.

Prior to packaging the jar, write any relevant build information to resource files to be consumed by the application when run.

### Expected Benefits

We will no longer rely on users of the JAR files to provide accurate information related to their build.

### Additional Context

_No response_

kwvanderlinde:

Resolves #5973

### Description of the Change

This removes support for the `-v` command line option that overrides the built-in version. The option itself is still registered (so that existing scripts can run new JARs), but a warning is logged to explain that the version number is no longer respected.

This has one change in behaviour for development runs (e.g., `gradlew run`). Tthe splash screen will now show the version as "DEVELOPMENT" rather than "0.0.1" since it is no longer provided in the application arguments and there is no JAR manifest from which to pull a `Implementation-Version`.

### Possible Drawbacks

Can't directly control the version details anymore.

### Documentation Notes

The `-v` command line option no longer has an effect.

### Release Notes

- Removed support for the `-v` command line parameter.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5980)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/MapTool.java
+++ b/src/main/java/net/rptools/maptool/client/MapTool.java
@@ -1545,9 +1545,6 @@ public class MapTool {
     // Start the update check early. Will be handled after the splash screen goes away.
     var updateFuture = AppUpdate.autoUpdateCheck();
 
-    String versionImplementation = version;
-    String versionOverride = version;
-
     if (OsDetection.MAC_OS_X) {
       // On OSX the menu bar at the top of the screen can be enabled at any time, but the
       // title (ie. name of the application) has to be set before the GUI is initialized (by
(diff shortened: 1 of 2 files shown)


src/main/java/net/rptools/maptool/client/MapTool.java
```diff
@@ -1563,8 +1560,8 @@ public static void main(String[] args) {
     }
 
     if (MapTool.class.getPackage().getImplementationVersion() != null) {
-      versionImplementation = MapTool.class.getPackage().getImplementationVersion().trim();
-      log.info("getting MapTool version from manifest: " + versionImplementation);
+      version = MapTool.class.getPackage().getImplementationVersion().trim();
```
cwisniew, line 1563: getPackage().getImplementationVersion() can return null, in which case this would result in an NPE

kwvanderlinde: There is already a check for `null` just above. Same code as was always used for `versionImplementation`.

cwisniew (approved):

PR #5982, opened 2026-06-18
Programmatically provide system property defaults

### LINKED ISSUE #5972: [Maintenance]: Programmatically set applicable system properties
### Describe the problem

There are a few system properties we currently set via command line options but which could be set programmatically. This adds extra effort for users trying to run MapTool JAR files.

Set the following command line options via `System.setProperty()` instead:
- `-Dsun.java2d.d3d=false`
- `-Dpolyglot.engine.WarnInterpreterOnly=false`
- `-Djava.util.Arrays.useLegacyMergeSort=true`

This should be done early in `LaunchInstructions` so they can take effect before the relevant systems are initialized. For bonus points, only set these if the user did not provide their own option on the command line, so they can experiment with things if they wish.

### Expected Benefits

Users of JAR files won't need to remember as many options to get the same behaviour out of MapTool as users of installed packages.

### Additional Context

_No response_

kwvanderlinde:

Resolves #5972

### Description of the Change

Sets the `sun.java2d.d3d`, `polyglot.engine.WarnInterpreterOnly`, and `java.util.Arrays.useLegacyMergeSort` system properties immediately on application start rather than requiring them to be set on the command line. If alternative values are provided on the command line, those will be used instead of the defaults. Also added is `sun.java2d.opengl` to be explicit that it should not be used by default, though it just so happens that the JVM does disable this by default (unliked `sun.java2d.d3d`).

### Possible Drawbacks

None

### Documentation Notes

N/A

### Release Notes

- Set `sun.java2d.d3d`, `polyglot.engine.WarnInterpreterOnly`, and `java.util.Arrays.useLegacyMergeSort` programmatically so they do not need to be provided on the command line.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5982)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -93,9 +93,7 @@ ext {
 }
 
 def javaArgs = [
-        "-Xss8M", "-Dsun.java2d.d3d=false",
-        "-Dpolyglot.engine.WarnInterpreterOnly=false",
-        "-Djava.util.Arrays.useLegacyMergeSort=true",
+        "-Xss8M",
         "--add-opens=java.desktop/java.awt=ALL-UNNAMED", "--add-opens=java.desktop/java.awt.geom=ALL-UNNAMED",
         "--add-opens=java.desktop/sun.awt.geom=ALL-UNNAMED", "--add-opens=java.base/java.util=ALL-UNNAMED",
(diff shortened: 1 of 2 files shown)


kwvanderlinde:
Keeping this as a draft for now as it should be merged after the other JAR improvement PRs.

cwisniew (approved):

PR #5983, opened 2026-06-18
Add Add-Opens to JAR manifest

### LINKED ISSUE #5975: [Maintenance]: Allow the JAR release to be run without --add-opens
### Describe the problem

Most of the effort in running the JAR file comes from the several `--add-opens` options required to make JPMS happy. If users didn't have to do this, it would be a much more friendly option for running MapTool.

I'm not sure which approach is most achievable, but something along the lines of one of these options:
1. Change our release JAR to be modular. _How this could interact with the shadow JAR I don't know. And maybe not all our dependencies will like to be in a modular build._
2. Change the JAR main class to act as a shim, forking a separate process with the necessary `--add-opens` provided.

### Expected Benefits

Users will be able to run our JAR without having to worry about the implementation detail of how modular we are.

### Additional Context

_No response_

kwvanderlinde:

Resolves #5975

### Description of the Change

Includes an `Add-Opens` attribute in the JAR manifest files. This means the user does not have to open these modules via `--add-opens` on the command line when running MT via `java -jar ${jarFile}`. *Note: the `Add-Opens` attribute isn't documented until [Java 23](https://docs.oracle.com/en/java/javase/23/docs/specs/jar/jar.html#main-attributes) but is functional at least as far back as Java 21*

The set of modules is defined in one list so that the shadow JAR, regular JAR, gradle `:run` task, and jpackage image all open the same set of modules.

Opening `java.desktop/com.sun.java.swing.plaf.windows` was obsolete and so has been removed.

### Possible Drawbacks

Should be none

### Documentation Notes

N/A

### Release Notes

- Add `Add-Opens` options to the released JAR manifests so that `--add-opens` command-line options are not required when running the JAR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5983)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -92,15 +92,37 @@ ext {
     println "OS Detected: " + osdetector.os
 }
 
+def modulesToOpen = [
+        'java.desktop/java.awt',
+        'java.desktop/java.awt.geom',
+        'java.desktop/sun.awt.geom',
+        'java.base/java.util',
+        'javafx.web/javafx.scene.web',
+        'javafx.web/com.sun.webkit',
+        'javafx.web/com.sun.webkit.dom',
+        'java.base/java.lang',
+        'java.desktop/sun.awt',
+        'java.desktop/sun.java2d',
+        'java.desktop/javax.swing',
+        'java.desktop/sun.awt.shell',
+]
+def jarManifestAttributes = [
+        'Implementation-Title': project.name + developerRelease,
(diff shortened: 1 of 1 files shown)


kwvanderlinde:
Keeping this as a draft for now as it should be merged after the other JAR improvement PRs.

cwisniew (commented):
Bogus comment ignore

cwisniew (approved):

PR #5991, opened 2026-06-29
Extend mt-theme-css with new contexts and element styles

### LINKED ISSUE #5961: [Feature]: MapTool theme and stat sheet CSS improvements

Baaaaaz:
Resolves #5961 

### Description of the Change
Towards html style parity with MapTool UI theme styles for all input controls, scroll bars (with or without tracks and arrow buttons) and Preference colors.

Changes:
* Standardized class names of `ScrollBarCSSContext` and `ProgressBarCSSContext` to previous naming convention i.e. *CssContext.
* New CSS contexts for swing controls (`CheckboxCssContext`, `ComboboxCssContext`, `ComponentCssContext`, and `MeterProgressBarCssContext`) and added to others.
* Added CSS context for App Preferences (inc. map labels).
* Added sections to the CSS file.
* Expanded `mt-theme-css` with new variables, new CSS properties for various html input controls and other elements, new CSS helper variables and classes.
* Added CSS comments for rationale of some choices.
* Replaced HTML `<body>` colors (previsously was using label colors for panel colors).
* Moved overflow scrolling from `<html>` to `<body>` due to `::-webkit-scrollbar viewport` state limitation.
* Use of java and/or handlebars logic to control the setting of some CSS properties based on theme differences.
* (Lazily) re-used checkbox variables for radio styling.
* Centralized null color handling - not all themes have the exact same them color keys needed and it is a chore to manually test them all as previously if you selected a theme without a specific key the theme generation borks.  Added error handling and logging here as well.
* Appended an approach for deprecating old CSS variables (in the absence of MapTool to serving different versioned CSS theme files instead)
* Implemented cut down version of variables, elements, and properties in `mt-stat-sheet-css`.

### Possible Drawbacks
May change users' existing HTML5 content appearance unexpectedly (if they reference the CSS and do not read the release notes before upgrading).

e.g. 
| Theme | v1.18.6 | after PR |
|---- |----|----|
|Arc Dark - Orange |  image    |   image    |
| Aah |  image | image|

### Documentation Notes
Documentation here: https://wiki.rptools.info/index.php/Theme_CSS may need updating.

### Release Notes
* Extended `mt-theme-css` with new contexts and element styles

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5991)


bubblobill:
TODO: What styles should be implemented in mt-stat-sheet-css?
_Tricky question._ 
**Technically** the popup cannot be interacted with so there is no need for any input, select, hover:, focus:, disabled: styles.

**Problems** 
1. My OS doesn't necessarily display things the same way as your OS.
2. You can use input elements to display useful stuff, e.g. A checkbox can be a handy visual for a Boolean value.
3. Things like scroll-bars cannot be interacted with but they still show that information overflows. I have notes in \<div> bocks that auto-scroll  when they are oversized. So the scroll bar shows.
4. Specificity pain. Over-riding the mt-stat-sheet.css can be a debugging nightmare.
5. Colour consistency.


So, we need to;
1. Identify if content normalising is necessary (likely true as the web engine probably uses system values). Provide a cut down version of something like [normalise](https://github.com/csstools/normalize.css/blob/main/normalize.css).
2. Ensure that the barest minimum of specifier is used so it is easily over-ridden.
3. Deal with input fields:
3.1 Exclude all of them and let people code their own, or,
3.2 Include everything and take the performance hit, or,
3.3 Provide a means of selectively including styles, e.g. separate files for the bits they want, i.e. mt-theme-checkbox.css, or, `@import checkbox.css`
4. I hate FlatLaF. Colours and variables. Go through this [themeCss.zip](https://github.com/user-attachments/files/29541087/themeCss.zip) and find all the conflicts/missing data/etc. and sensiblise everything to construct a consistent product.

Once you've done all that, I'm sure I can add to the list.

Baaaaaz:
1. I do not know about other OSs.  So unlikely to take that on in this PR.
2. Wilco.
3. t.b.h. I suspect I will do the bare minimum (well option 3.1 plus a bit).  Providing some new color variables which users can reference would be some progress, also perhaps some basic stuff for checkboxes, progressbars, and scrollbars?
4. That sounds more like FlatFaff!  I prefer my exception handling of theme key color differences (still TODO), and offloading specific variant handling to future user requests!

bubblobill:
New challenge. 
What CSS info would be appropriate for rendering states and bars?

Baaaaaz:
> New challenge. What CSS info would be appropriate for rendering states and bars?

It depends.

Who is in control of what html elements they are represented by?  

Do you want to mimic how they can be shown relative to the token on a map?  e.g. states can be shown in different sized grids over a token.  e.g. progress bars cannot be vertical so that counts them out for left or right sided bars.

bubblobill:
I include arbitrary max width and height for stat-sheets to ensure they do not cover the available screen. This should probably become a user setting and a variable, e.g. instead of this
```
:root {
    --max-height: 80vb;
    --max-width:  min(80vi, 850px);
}
```
have this
```
#statSheet {
	max-height: var(--mt-stat-sheet-max-block-size);
	max-width:  var(--mt-stat-sheet-max-inline-size);
}
```

This is what I use to normalise display [normalise.zip](https://github.com/user-attachments/files/29669977/normalise.zip)

I adjust font-size to ensure legibility, allowing it to grow where space permits.
```
font-size:clamp(var(--mt-theme-font-size), calc(11px + 0.390625vmin), calc(var(--mt-theme-font-size) * 1.4));
```

The fixed size of the portrait is somewhat naiive as it does not allow for small display spaces. I adjust it for limited room. Sample:
```
  --portrait-width:  max(80px, min({{portraitWidth}}px, 20vi));
  --portrait-height: calc({{portraitHeight}} / {{portraitWidth}} * var(--portrait-width));
  --max-attribute-width: calc(var(--max-width) - var(--portrait-width) - 7ch);
```

When a user setting prevents the portrait being displayed, how long/tall is a bar? What scale do you use to display a state? The layout and display is up to the author, the question is probably more around should we have sensible minimum values.

Another problem is missing values in the Handlebars data. For example, when there is no portrait, there are no values for portraitWidth/Height, which can invalidate the CSS above. We don't have access to CSS [if]([url](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/if)) for butt-covering purposes, so you might want to think about the sort of tests you are doing for colour and apply the same thinking to other values that people might rely on.

As for vertical bars, there are many ways to achieve it :)

image

Baaaaaz:
Sounds good but also sounds like that needs to be spec'd out in a different issue and beyond the scope of this PR which now resolves #5961 

(my real aversion to doing in this PR is that I have zero XP in developing custom stats sheets / addons and do not want that fact to hold up this change!)

cwisniew (approved):

Baaaaaz:
@cwisniew thanks for reviewing and approving.  Is there a reason it was not automagically merged into develop?

cwisniew:
> @cwisniew thanks for reviewing and approving.  Is there a reason it was not automagically merged into develop?

No idea, other than GitHub has been a bit flaky/laggy for me last few days.

bubblobill:
@Baaaaaz 
Oi! 
this ->
```
16:27:15.160 (ThemeCssContext.java:148) [URL-Loader-1] WARN  - The theme "Aah(Large Print)" does not define a color for "Button.background".  Any corresponding CSS theme variable will be blank.
```
Considering I made ->
```
Button.startBackground   = if($Button.background, lighten($Button.background, 10%), $Button.default.startBackground)
Button.endBackground     = if($Button.background, darken($Button.background, 5%), $Button.default.endBackground)
```

Grumble. Don't encourage the FlatLaF

Baaaaaz:
I'm not sure exactly what are you are grumbling about?!  😉 

Should it not have logged anything at all?  Should it be using INFO or DEBUG instead?  Or do you think the text used in the message is incorrect?  Or something else?!

Doing some digging the AahLAF.properties file contains the `Button.background` key but does not assign anything to it.  i.e. 

```
Button.background
Button.foreground        = if($Button.background, contrast($Button.background, @col90, @col5), $Button.default.foreground)
Button.startBackground   = if($Button.background, lighten($Button.background, 10%), $Button.default.startBackground)
Button.endBackground     = if($Button.background, darken($Button.background, 5%), $Button.default.endBackground)
```

So the log message seems correct to me (the fact that there is no corresponding css variable for `Button.background` is because it is not directly being used in css button styling, but as a fallback in the `ButtonCssContext`).

Why have such as fallback in `ButtonCssContext`?  Well, there are a minority of themes such as "Cyan Light" and "Gray" which do not have color keys for `Button.startBackground` or `Button.endBackground`, but they do have a `Button.background` key color.  Prior to this PR in develop if you selected one of those themes MapTool would throw a wobbly and not generate the mt-theme.css at all...  e.g. if you run it at commit 30b2dd5b and select one of those themes and try and build the css you get:

```
(LoggingProviderImpl.java:257) [URL-Loader-1] WARN  - Unexpected error 
java.lang.NullPointerException: Cannot invoke "java.awt.Color.getRed()" because "color" is null
        at net.rptools.maptool.model.library.builtin.themecss.ThemeCssContext.formatColor(ThemeCssContext.java:90)
        at net.rptools.maptool.model.library.builtin.themecss.ButtonCssContext.<init>(ButtonCssContext.java:62)
        at net.rptools.maptool.model.library.builtin.themecss.ThemeCssContext.<init>(ThemeCssContext.java:73)
        at net.rptools.maptool.model.library.builtin.MapToolBuiltInLibrary.processHandlebarsTemplate(MapToolBuiltInLibrary.java:69)
```

So the currently solution gracefully handles some LAF file differences that I noticed during development, and helpfully notifies you of where it has done so and under which theme.  1.18.6 does not seem to throw the same wobbly though I could tell from looking just now, but my digging days are over on this PR! 💤

bubblobill:
For reliable rendering of gradients FlatCrap is (or was at some point) exclusive-or on the colours. So you provided either `Button.background` **or** `Button.startBackground` & `Button.endBackground`

`--mt-theme-button-background` should be `linear-gradient(to bottom, Button.startBackground Button.endBackground)`. 

`--mt-theme-button-background-colour-0: Button.startBackground`
`--mt-theme-button-background-colour-1: Button.endBackground`

If the theme does not provide gradient stops, then both values are the same

PR #5993, opened 2026-06-30
Handle more general geometries when converting AWT shapes to JTS polygons

### LINKED ISSUE #5992: [Bug]: Fog of War movement blocking edge cases
### Describe the Bug

There are still some FoW geometries that can't be processed correctly, resulting in these elements being discarded. The net effect is that tokens for non-GM players no longer pathfind properly as they think they are travelling through hard FoW even in cleared areas.

### To Reproduce

1. Open this campaign: [broken fow.cmpgn.zip](https://github.com/user-attachments/files/29518683/1.19.-.broken.fow.cmpgn.zip)
2. Enable _View_ > _Show As Player_
3. Try to drag the token through the cleared area with no Fog of War.
    - Notice that the movement distance is always 0, and no path is drawn.

### Expected Behaviour

We should handle these FoW cases so tokens can pathfind through the cleared areas.

### Screenshots

_No response_

### MapTool Info

1.18.6

### Desktop

Linux Mint 22.3

### Additional Context

_No response_

kwvanderlinde:

Fixes #5992

### Description of the Change

The particular problem in the linked issue was that precision reduction turned a tiny polygon into a point (represented as a `LineString`). This meant the `fixedPolygon` could not be represented as a `MultiPolygon` but was instead a `GeometryCollection`. We previously ignored those as they were not `Polygonal`, but this PR fixes that.

Now, after fixing and reducing the geometry, we flatten the result into a list. This allows us to pick out good geometry from an arbitrary `GeometryCollection`, even if these collections are nested. From that list, we discard any unusable geometry, keeping only the `Polygon` geometries we can use.

A new test cases ensures we are accurately pulling out the polygonal geometry.

### Possible Drawbacks

This will technically reduce pathfinding and vision performance, though I have not noticed anything significant there yet.

### Documentation Notes

N/A

### Release Notes

- Fixed a bug where certain fog of war geometries would wrongly block token movement through clear areas.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5993)


FullBleed:
Dropping an inquiry here only because it's loosely related to how MT must calculate what geometry is in a square to allow/disallow a token to move through it.  I've got a hacked method to calculate the amount of a given type of vbl/mbl in a square... but it seems like there might be a much more accurate/faster method already built into MT's system but not exposed to macroscript.  How hard would it be to give us a function to pull that value?  (trying to decide if I should put in a freq for it)

PR #5997, opened 2026-07-01
Tables window details view and other fixes

### LINKED ISSUE #2248: Alternative views for Tables in Table Window
Currently the only way to view tables - without a macro - is to open the Table Window and you see a grid of icons - 1 for each table. However, the table names are often longer then 12 characters - which can make it hard to find tables if you are using many of them. I can assign icons to each table - but it doesn't solve the issue and you then have to manage and remember many different icons. 

The table window should have a "details" view - which is similar to "Detail View" in Windows Explorer or ls -la in linux - basically a verticle list view of tables that shows the full name/title of them, date they were last modified, etc. etc (perhaps the columns can be configured?) 

I suppose a custom macro could be written to essentially replace the standard Table window - but I haven't implemented anything like that.

Baaaaaz:
closes #2248

### Description of the Change

Provide an optional details view of the tables window in a tabular format.  Views can be selected via new toggle buttons in the table window.  The details view shows a different set of columns for a GM than a player.  Default is the existing Image Panel.

Changes:
* `LookupTableFunction#processMutatedLookupTable()` - >  centralize handling of table updates from MTScript functions in the same class) to the server and refresh of the `LookupTablePanel`
* `LookupTableMacro` ->  updated logging to use I18N.
* `LookupTablePanel` ->  switch to use a CardLayout and addition of the `LookupTableDetailsTablePanel`.
* `LookupTablePanelView`  ->  Addition of toggle buttons (and associated new button icons) to switch between the image panel and the details table panel.
* `CampaignManager` ->  adding resetting the temp lookup table.
* `LookupTable` ->  new methods to get counts of entries, entries with values, and entries with images.
* `MapTool` -> reset and refresh the `LookupTablePanel` when setting a campaign (e.g. resolves stale `LookupTablePanel` when clients connecting to a server or when gm loads campaign with clients connected).
* `AppActions` -> 
  * only allow GMs to add default tables (otherwise players may override customised tables with the same names in a connected campaign).  
  * corrected issue where if a server was running and the gm added the default tables, they would not be show in connected clients' Table window.
* `EditLookupTablePanel` -> 
  * if the lookup table name changes (and is not a new table), delete the original lookup table on the server
  * addition of icons for show to players, allow lookup, and pick once (next to their associated checkboxes and labels)
* `RessourceManager` & `Icons` -> new Icons.  `misc/Zoom.svg` for Player Lookup also now used for Asset Panel Search.

New:
* `LookupTableDetailsTablePanelModel` -> Defines the table and (GM/Player) columns for the panel, inc. icons, values, renderers, widths, etc.
* `LookupTableDetailsTablePanel`  -> 
  * Implements a scrollable details table based on the underlying model.
  * Employs snapshot list of `LookupTable`s to avoid race conditions when updating the JTable.
* `LookupTableDetailsTableColumnRenderers` -> based on the column metadata specify which header renderer to use, one of:
  * `TableHeaderAlignedRenderer` (default) -> use default header renderer but with configurable alignment
  * `TableHeaderIconRenderer` -> use default header renderer but with configurable icon and/or text
* `LookupTableDetailsTableColumnRenderers` -> based on the column metadata specify which column renderer to use, one of:
   * `TableColumnAlignedRenderers` (default) - align columns
   * `TableColumnBooleanRenderer` - disable checkboxes
   * `TableColumnIconRenderer` - fit images to the table row height

| Role | Image View | Details View |
|------|------------|--------------|
| Player |     image       |      image        |
| GM |     image       |     image     |

### Possible Drawbacks
* Trying to find all the ways LookupTables get updated across clients was challenging - hopefully I have not missed any.  
* I may not have done things in the best way...  so would welcome feedback here!
* Players can no longer load default tables into a campaign (and overwrite any other tables with the same name)- pretty sure this is undesirable behavior.  The GM can still load them okay and now they will appear correctly on clients.

### Documentation Notes
* Table views can be toggled between the image and details views using the buttons at the top rightof the tables window.
* For Players the details tabular view will show the table image and the table name (good for tables with long names!).
* For GMs the details tabular view will show table image, table name, entries count, values count, images count; also **disabled** checkboxes for Show To Player, Allow Player Lookup, and Is Pick Once Table (providing an overview of tables without having to enter edit mode for each table).
* If a GM in tabular view, double-clicking on the table icon will roll on the table and double-clicking on any other column will open the lookup table edit panel.

### Release Notes
- Added the option to show table details in a tabular view in the Tables Window

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5997)


src/main/java/net/rptools/maptool/client/ui/lookuptable/EditLookupTablePanel.java
```diff
@@ -291,6 +293,10 @@ public boolean commit() {
     if (!name.equals(origname)) {
       // New name is not the same as the existing name
       MapTool.getCampaign().getLookupTableMap().remove(origname);
+      // if not a new table (i.e. create/duplicate) delete the previous named one from the server.
+      if (!isNew) {
+        MapTool.serverCommand().deleteLookupTable(origname);
```
kwvanderlinde, line 306: Good find - this was a regression in #5868.

A future improvement for us would be to have unique table IDs so renames are actually renames and not delete+create.

kwvanderlinde (commented):
I still have to go through the implementation, but the UI looks pretty nice on the whole.

I have some suggestions, based partly on the fact that I prefer a narrower *Tables* window docked to the side of the window. In that case, the details view would end up looking like this:
Image

- We can do without the **#Values** and **#Image** columns. This would also save some CPU cycles for those who run very large tables (not sure how big a deal this latter point will be).
- The column names can be a bit more concise. _Actually, icons or something else narrow would be nice for the checkbox columns, but that's a bigger ask._
    - **Show Table to Players** -> **Show to Players**
    - **Pick Once Table** -> **Pick Once**
    - **Allow Lookup by Players** -> … Actually, I don't have an idea for shortening this one.
- We should probably enforce some minimum size (determined by the column headings, perhaps), and allow the details table to scroll horizontally if that is exceeded. This would guarantee that at least the table structure is readable.

Baaaaaz:
> I still have to go through the implementation, but the UI looks pretty nice on the whole.

Thanks for taking an initial look!

I was debating whether to:
* add a third list view (basically what the Players version of the detail view currently shows but also available to GMs) and make the current details view available to GMs only, or
* or make the details columns GM configurable (but that seemed a lot more work).

Personally I like having all the summary table stats as it enables management of the tables at a glance (esp. when developing or importing tables or getting to grips with others frameworks).  Less useful when running a game.  I tested it with a several large data tables (with no images though) and had no noticeable drop in performance.  I suppose I could create some more with images.  Not sure how to add counters/timers to non-zone rendering things though - is that even possible/desirable for this?  

Columns icons are good idea, will revisit.  I did previously have ticks and cross icons, but the disabled check boxes were on a potential path to enabling them on that panel as well, suppose that could be done with icons though if that was desired.  

Will be more concise with the names... I did just use the existing entries for I18N tables for those, but will create some new keys.

Adding horizontal scroll and column sizing.  Agree, will add/revisit.

kwvanderlinde:
> I tested it with a several large data tables (with no images though) and had no noticeable drop in performance. I suppose I could create some more with images. Not sure how to add counters/timers to non-zone rendering things though - is that even possible/desirable for this?

Probably if there is no performance issue without images of should also be fine with images since these don't factor into the control flow when counting.

And it's totally possible (and easy!) to add timers anywhere. Figuring out how and when to report them is the hard part. I wouldn't worry about it for now.

Baaaaaz:
Addressed initial feedback and some other changes:

* Columns no longer get squashed on panel resizing
* Now comes with both scroll bars if needed.  
* Moved summary counts to the right (as less relevant during play)
* Added new icons for table boolean options to both the details panel and the edit panel:
  * image show to players 
  * image allow lookup
  * image pick once

Example output after changes:

image

kwvanderlinde:
I really like how this new table looks! Hoping to do a proper review of the code changes soon.

Baaaaaz:
Slight UI tweak...

Moved roll column to the right to improve visual scanning, e.g. 
* increase boolean icon columns' prominence 
* the roll column is next to summary counts for easier comparison
* the roll column is next to the pick once icon column for easier determination of empty roll values due to pick once (or not)

image

FullBleed:
Looks good... but why are there pipes around "Entries", "Values", and "Images"?  It crowds the text and looks weird, imo.

src/main/resources/net/rptools/maptool/language/i18n.properties
```diff
@@ -677,6 +677,16 @@ LookupTablePanel.error.saveFailed  = Could not save table "{0}".
 LookupTablePanel.info.saved        = Table saved as "{0}".
 LookupTablePanel.msg.titleEdit     = Edit Table
 LookupTablePanel.msg.titleNew      = New Table
+LookupTablePanel.countEntries         = |Entries|
+LookupTablePanel.countEntries.tooltip = A count of the number of table entries.
+LookupTablePanel.countValues          = |Values|
+LookupTablePanel.countValues.tooltip  = A count of the number of table entries with values.
+LookupTablePanel.countImages          = |Images|
+LookupTablePanel.countImages.tooltip  = A count of the number of table entries with images.
```
kwvanderlinde: Echoing @FullBleed's comment, let's remove the pipes.

Baaaaaz:
> why are there pipes around "Entries", "Values", and "Images"? 

Because vertical bars are a known cardinality notation for the number of elements in a set.  My mistake was just using a single bar where perhaps I could have instead used double bars to indicate a multiset (including duplicates) e.g. `||Entries||`. 

> It crowds the text and looks weird, imo.

Now you know what they means, maybe it will now not so look weird?!

src/main/resources/net/rptools/maptool/language/i18n.properties
Baaaaaz: Can address when I get other requested changes.

Or GMs who do not like the extra columns can adjust the column or panel widths so they remain out of sight!  :wink:

FullBleed:
>  My mistake was just using a single bar where perhaps I could have instead used double bars to indicate a multiset (including duplicates) e.g. `||Entries||`.
> 
> Now you know what they means, maybe it will now not so look weird?!

Nope.  Looks twice as weird.  ;)

If the headers need additional context then I'd suggest a tooltip.

src/main/resources/net/rptools/maptool/language/i18n.properties
kwvanderlinde: I'd say this PR is otherwise good to go!

Baaaaaz: Pipes removed.

kwvanderlinde (approved):

PR #5999, opened 2026-07-03
Fix `ChatTypingPanel` and `ChatActionLabel` notifications not displaying

### LINKED ISSUE #5907: [Bug]: Typing Notifications not displaying properly.
### Describe the Bug

GM running a server sees no client typing notifications.

Clients only see typing notifications if map is not visible.

### To Reproduce

Start Server and connect Client.
Make map Player Visible.
Type on Server and notice that the typing notification does not display on Client.
Make map not visible to player. 
Type on Server and notice that the typing notifications now DO display on Client.
Type on Client and observe that the typing notification does not show on Server (regardless of map visibility).

### Expected Behaviour

Typing notifications should display on Server.
Typing notifications should display on Client regardless of Map Visibility.

### Screenshots

_No response_

### MapTool Info

1.18.6

### Desktop

Win11

### Additional Context

Original bug reported :  https://discord.com/channels/296230822262865920/1446791445674590288

What should the **_Tool>Enforce Typing Notifications_** setting be doing?  Should it be remembered or be a server setting?

Baaaaaz:
fixes #5907

### Description of the Change
Set the Z-Order for the `ChatTypingPanel` and `ChatActionLabel` components so they both now appear on top of maps.   In full screen mode, when they appear they will appear over the top of any full screen tools (rather than under them as it was previously).  

### Possible Drawbacks
none anticipated

### Documentation Notes
none

### Release Notes
fixed an issue with chat typing and action notifications not displaying over the top of maps

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/5999)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/MapToolFrame.java
+++ b/src/main/java/net/rptools/maptool/client/ui/MapToolFrame.java
@@ -426,6 +426,10 @@ public class MapToolFrame extends DefaultDockableHolder implements WindowListene
     pointerToolOverlay = new PointerToolOverlay();
     zoneRendererPanel.add(pointerToolOverlay, PositionalLayout.Position.CENTER, 0);
 
+    // bring chat notifications to the front
+    zoneRendererPanel.setComponentZOrder(getChatTypingPanel(), 0);
+    zoneRendererPanel.setComponentZOrder(getChatActionLabel(), 0);
+
     // Put it all together
     setJMenuBar(menuBar);
     add(BorderLayout.NORTH, toolbarPanel);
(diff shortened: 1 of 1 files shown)


Baaaaaz:
This is ready for a review.

kwvanderlinde:
In fullscreen mode, is it possible to move the notification down below the tools instead of being overtop of them?

Baaaaaz:
Yeah I agree that would be better.  I did have a look into that but got a bit lost trying to offset stuff in the same positional area depending on whether the tools were enabled in full screen mode or not (i.e. toggled with `Ctrl+Alt+T`).

At the time I think I justified what I came up with to myself as being okay as:
* the typing notification is temporary
* users can turn the typing notification background off in their preferences
* (from my quick tests) I could click through the notification it to the full screen buttons underneath 

I can revisit if you want.  If I can crack that I may be able to group the chat action notification and chat typing notification into the same region as not sure why they are at different corners of the screen anyway?!

kwvanderlinde:
I'm fine with the changes as they are. Just wanted to see if you had any easy ideas on fixing it.

Probably we need to rethink the fullscreen layout a bit to better support these things.

kwvanderlinde (approved):

Issue #5279, opened 2025-02-21
[Feature]: Close frames and dialogs when switching campaigns


kwvanderlinde:
### Describe the Problem

I often write bespoke `frame5()` for various custom tools / UIs in my campaigns. These often reflect campaign state, and aren't applicable outside of the campaign they were written for. But when switching campaigns or creating a new campaign, these frames stick around as unwanted clutter. Also, if the frames use the Fetch API or similar to get campaign data, they may begin to log errors to chat as the expected libraries, etc., are no longer present.

### The Solution you'd like

Frames would be tagged with the context in which they were created. If created from a macro in the Global panel, they would be tagged as global frames. If created from chat, the Campaign panel, GM panel, a library, or any token macro they would be tagged as campaign-local frames.

When a campaign is closed, whether by opening another campaign or creating a new campaign, any campaign-local frames would be automatically closed while global frames would be allowed to live on.

The same logic would apply to dialogs. And I think overlays should be given the same treatment as well - today they are closed whenever a campaign is closed, even if created from the Global panel.

### Alternatives that you've considered.

Adding new parameters to the various HTML functions and roll options to explicitly mark them as campaign-local.

PR #6005, opened 2026-07-09
Use assets for custom html frame and dialog icons

### LINKED ISSUE #4018: [Feature]: Custom icons for HTML frames
### Describe the Problem

When even a handful of frames are being docked together, especially on a small screen, there isn't much room for the names to be shown. For built-in frames, this isn't a problem because they can (mostly) be distinguished by their icons. Custom frames on the other hand all have the same icon as the chat windows, so they aren't easy to distinguish.

### The Solution you'd like

A new `icon=` option to `frame()`, `frame5()`, `html.frame()` and `html.frame5()` to provide a custom icon. The custom icon would take the form of an `asset://` url that references the desired image to use as an icon. Some predefined keywords could also be allowed in case the frame fills a role similar to the built-in frames.

Examples:
```
[h: html.frame5("My Frame", "lib://Frames/macro/my_frame.html", "icon=asset://someassettoreference")]
[h: html.frame5("Custom Tables", "lib://Frames/macro/tables.html", "icon=tables")]
```

I could expand my frame docks so there's room for the title. In the same vein, I could avoid docking multiple frames on the left and right of the screen and instead prefer docking at the bottom of the screen. But I find left and right docking very convenient in other regards, and expanding those docks for the sake of seeing frame names would make them eat up too much screen real estate.

### Additional Context

Here's what my left side panel looks like on my laptop, where very little of the frame names are visible:
image

Baaaaaz:
closes #4018 

### Description of the Change
Icons for `DIALOG`, `DIALOG5`, `FRAME, `FRAME5` (and their `html.*` variants) can now be set by:
1. the frame/dialog property `icon=asset://...;`
2. in HTML using the `<link rel="icon" type="..." href="asset://...">` (requires all three attribute to be present).
3. by the functions `setFrameIcon()` or `setDialogIcon()` without having to do a full frame/dialog refresh.  However, if the frame/dialog is subsequently refreshed the icon will change to whatever has been specified according to methods 1 and 2 above.

**Code changes:**
* HTMLFrameFactory -> handling of the new `icon` parameter and new `getScaledImageIcon` method.
* HTMLActionEvent -> new `ChangeIconActionEvent` class.
* HTMLPane and HTMLWebViewManager -> additional handling of `<link rel="icon"` type tags and attributes.
* HTMLFrame and HTMLDialog -> manage updating the icon.

### Possible Drawbacks
* If setting via the HTML `<link>` tag, the property icon is set first (or set the default current icon if the property icon is not set), then the `<link>` tag icon is set.  The icon may flicker as it needs to wait for the HTML content to load first.  This flicker is similar to if setting the frame/dialog title via the `<title>` tag.  Had a go a cracking this, but was getting no where fast and I persevered for as long I could.

#### New functions:
`setFrameIcon(frameName, assetId)`
* `frameName` - the name of the frame
* `assetId` - the asset ID.  If blank (i.e. "" will reset the frame icon to the default)

ditto for dialogs

#### Enhanced functions:
`getFrameProperties()`: now returns an `icon` key with the assigned asset.
`getDialogProperties()`: now returns an `icon` key with the assigned asset.
`frame/frame5/dialog/dialog5/html.frame/html.frame5/html.dialog/html.dialog5` can now take an assetid for the `icon` parameter.

### Release Notes
- frame and dialog icons can now be customized to an image image.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/6005)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLFrameFactory.java
+++ b/src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLFrameFactory.java
@@ -341,9 +341,11 @@ public class HTMLFrameFactory {
         // if we have an asset, convert it to an icon and scale it
         if (asset != null) {
           ImageIcon imageIcon = new ImageIcon(RessourceManager.getImage(Images.BROKEN));
-          byte[] imageBytes = asset.getData();
-          if (imageBytes != null && imageBytes.length > 0) {
-            imageIcon = new ImageIcon(imageBytes);
+          if (asset.getType() == Asset.Type.IMAGE) {
+            byte[] imageBytes = asset.getData();
+            if (imageBytes != null && imageBytes.length > 0) {
+              imageIcon = new ImageIcon(imageBytes);
+            }
           }
           Image scaledImage = imageIcon.getImage().getScaledInstance(16, 16, Image.SCALE_SMOOTH);
           scaledImage.getWidth(null); // force completion
--- a/src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLFrameFactory.java
+++ b/src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLFrameFactory.java
@@ -15,13 +15,22 @@
 package net.rptools.maptool.client.ui.htmlframe;
 
 import com.google.common.eventbus.Subscribe;
+import java.awt.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.swing.*;
+import net.rptools.lib.MD5Key;
 import net.rptools.maptool.client.AppConstants;
 import net.rptools.maptool.client.MapTool;
 import net.rptools.maptool.client.events.ZoneActivated;
 import net.rptools.maptool.client.events.ZoneDeactivated;
+import net.rptools.maptool.client.ui.theme.Images;
+import net.rptools.maptool.client.ui.theme.RessourceManager;
 import net.rptools.maptool.client.ui.zone.SelectionModel;
 import net.rptools.maptool.events.MapToolEventBus;
 import net.rptools.maptool.language.I18N;
+import net.rptools.maptool.model.Asset;
+import net.rptools.maptool.model.AssetManager;
 import net.rptools.maptool.model.Token;
 import net.rptools.maptool.model.Zone;
 import net.rptools.maptool.model.zones.TokensChanged;
@@ -66,6 +75,7 @@ public class HTMLFrameFactory {
     boolean hasFrame = true;
     boolean closeButton = true;
     boolean scrollReset = false;
+    String iconRef = null;
 
     if (properties != null && !properties.isEmpty()) {
       String[] opts = properties.split(";");
@@ -153,10 +163,13 @@ public class HTMLFrameFactory {
           frameValue = value;
         } else if (keyLC.equals("tabtitle")) {
           tabTitle = value;
+        } else if (keyLC.equals("icon")) {
+          iconRef = value;
         }
       }
     }
     if (tabTitle == null) tabTitle = title; // if tabTitle not set, make it same as title
+
     if (frameType == FrameType.FRAME) {
       HTMLFrame.showFrame(
           name,
@@ -168,7 +181,8 @@ public class HTMLFrameFactory {
           scrollReset,
           isHTML5,
           frameValue,
-          htmlContent);
+          htmlContent,
+          iconRef);
     } else if (frameType == FrameType.DIALOG) {
       HTMLDialog.showDialog(
           name,
@@ -182,7 +196,8 @@ public class HTMLFrameFactory {
           scrollReset,
           isHTML5,
           frameValue,
-          htmlContent);
+          htmlContent,
+          iconRef);
     } else if (frameType == FrameType.OVERLAY) {
       MapTool.getFrame()
           .getOverlayPanel()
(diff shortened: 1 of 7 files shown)


Baaaaaz:
Some MTScript to have a play with toggling frame/dialog icons:

```
[h: vMacroArgs = macro.args]
[h, if (vMacroArgs == ""), code: {
	<!-- First run -->
	[h: doPropertyIcon = 0]
	[h: doLinkIcon = 0]
	[h: vMacroArgs = "{}"]
} ; {
	<!-- Subsequent runs via macrolinks -->
	[h: doPropertyIcon = json.get(vMacroArgs, "doPropertyIcon")]
	[h: doLinkIcon = json.get(vMacroArgs, "doLinkIcon")]
}]

[h: assert(listCount(getSelected())==1, "You must first select one and only one token")]

[h: macroLinkTogglePropertyIcon = macroLink("Toggle All Property Icons !" + doPropertyIcon + " (uses selected token image)", getMacroName()+"@"+getMacroLocation(), "all", json.set(vMacroArgs, "doPropertyIcon", 1-doPropertyIcon, "doLinkIcon", doLinkIcon))]
[h: macroLinkToggleLinkIcon = macroLink("Toggle All HTML Link Icons !" + doLinkIcon + " (uses selected token portrait)", getMacroName()+"@"+getMacroLocation(), "all", json.set(vMacroArgs,  "doPropertyIcon", doPropertyIcon, "doLinkIcon", 1-doLinkIcon))]

[h: vTokenImage = getTokenImage("", getSelected())]
[h: vTokenPortrait = getTokenPortrait("", getSelected())]

[h: vHtmlTypes = "FRAME5,FRAME,DIALOG5,DIALOG"]
[h: frameNameSuffix = ""]
[h: isTemp = 1]

[h: vProperty = strFormat("temporary=%s;", isTemp)]
[h, if(doPropertyIcon): vProperty = vProperty + strFormat("icon=%s; ", vTokenImage)]
[h, if(doLinkIcon): vHtmlHead = strFormat('<head><link rel="icon" type="image/x-icon" href="%{vTokenPortrait}"></head>') ; vHtmlHead = '']

[r, forEach(iHtmlType, vHtmlTypes), code: {
	[r: execMacro( strFormat('[%{iHtmlType}("%{iHtmlType}%{frameNameSuffix}", "%{vProperty}"): { %{vHtmlHead} <body>%{macroLinkTogglePropertyIcon}<br><br>%{macroLinkToggleLinkIcon}<hr>N.B. a link icon will override a propery set icon</body>}]'))]
}]

<h1>Frames</h1>
[r, forEach(iHtmlType, vHtmlTypes, "<br>"), code: {
	<pre>[r: json.indent(getFrameProperties(iHtmlType+frameNameSuffix))]</pre>
}]
```

kwvanderlinde:
Nice to see this being picked up! I like that HTML tags can be used to do this.

One general concern I have developed in the few years since opening the FR is that not all of "features" will keep working right if/when we tighten up the asset cache so that loaded assets need to be part of the campaign. One case I remember where this is a "no" is global macros that use assets as icons. I don't know what the answer would be for frames & dialogs. Not that it's really new for them, as they can already load assets in `` tags and the like. Related to that is #5279 for scoping frames to campaign, as that should side-step the issue.

Anyhow, I don't think anything needs to change in this PR regarding that point. I just wanted to make sure it's something we think about, and maybe we'll have to rethink a couple things in the future.

src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLFrameFactory.java
```diff
@@ -297,4 +312,45 @@ public static boolean isInternalOnly(String name) {
...
+        MD5Key assetId = new MD5Key(assetIdStr);
+        Asset asset = AssetManager.getAsset(assetId);
+        if (asset == null) {
+          asset = AssetManager.requestAssetFromServer(assetId);
+        }
+        // if we have an asset, convert it to an icon and scale it
+        if (asset != null) {
+          ImageIcon imageIcon = new ImageIcon(RessourceManager.getImage(Images.BROKEN));
+          byte[] imageBytes = asset.getData();
+          if (imageBytes != null && imageBytes.length > 0) {
```
cwisniew: Should check that the asset is an image with asset.getType()

src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLPane.java
```diff
@@ -315,6 +327,8 @@ void handleLinkTag(MutableAttributeSet attributes) {
           } catch (IOException | ExecutionException | InterruptedException e) {
             // Do nothing
           }
+        } else if (rel.toString().equalsIgnoreCase("icon")) {
```
cwisniew, line 330: rel could be null so need to check that it isn't first

cwisniew (changes requested):

src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLPane.java
Baaaaaz: Think that is already covered by the checks done at the parent nested level on line 308?

src/main/java/net/rptools/maptool/client/ui/htmlframe/HTMLFrameFactory.java
Baaaaaz: Added an asset type check.

kwvanderlinde (approved):

cwisniew (approved):

Issue #6003, opened 2026-07-07
[Maintenance]: Use Short and Display names throughout client UI


bubblobill:
### Describe the problem

Atrributes have an actual name, a short name, and a display name. These are only used for pop-up attribute sheets, when they should be presented to the user in all UI elements, e.g. the Token Property Editor. 

A good variable name rarely makes a good display name.

This added functionality should also be extended to other user definable campaign properties such as states, bars, lights, and sight.

### The improvement you'd like to see

Use the display name in the Token Property Editor and elsewhere.

Implement short and display names for all user definable campaign properties.

Make this information available (as appropriate) to Handlebars context data.

Macro Functions to look up names, e.g. `json.get(getPropertyNames("Mana Count"), "canonical")`.

### Expected Benefits

This would also permit more human readable entries in things like context menus.

Pop-up attribute sheets will be able to display state/bar image names more compactly.

### Additional Context

As the bar/state name is not normally shown on the map, people tend to use descriptive names that can be long. This can conflict with my layout constraints to make things ugly. It would be preferred to have the option of using a short name

Issue #5944, opened 2026-03-12
[Feature]: Enhanced input() function


Baaaaaz:
### Describe the Problem

Input is nice for quick data entry/maintenance, but the current range of input types and options is limited.

### The Solution you'd like

New input types such as:
- SPINNER: for managed number input
- SLIDER: for managed number input
- TEXTAREA: for long text input
- CHECKLIST: for grouping related checkboxes

New control types such as:
- TITLE: to be able to customize the input dialog title
- PANEL: à la html fieldsets for grouping input types and which are placeable
- WIZARD: alternate TAB implementation, so behaves like an input wizard with previous/next buttons

Other improvements:
- allow TABs to be placed above/below/left/right
- allow a customizable end prompt character (currently a colon)
- ability to set default values relevant to the input type
- input validation on ok/next

### Alternatives that you've considered.

Just build stuff in html5 dialogs/frames.

Baaaaaz:
e.g. some WIP...

<img width="1444" height="1475" alt="Image" src="https://github.com/user-attachments/assets/d86b40c4-9ca8-4ac1-9649-13d7e73612c5" />

FullBleed:
These improvements would be very welcome!

Thanks for taking this on.

Baaaaaz:
WIP RATING:

<img width="688" height="1484" alt="Image" src="https://github.com/user-attachments/assets/7d6af0d3-4181-4450-9199-50314844fca0" />

Baaaaaz:
WIP SPINNERPLUSMINUS

<img width="326" height="260" alt="Image" src="https://github.com/user-attachments/assets/b5396cd7-5acb-41c8-b11d-7612a7ee9940" />

Baaaaaz:
WIP BAR (with number entry field on focus)

<img width="370" height="595" alt="Image" src="https://github.com/user-attachments/assets/a83d33bf-50a8-46f2-9dec-917092416be4" />

Baaaaaz:
WIP CAROUSEL / CARD LIST

<img width="1460" height="1412" alt="Image" src="https://github.com/user-attachments/assets/c34486d1-e44e-4479-b4b1-18102ef421b7" />

Baaaaaz:
WIP WIZARD

<img width="559" height="735" alt="Image" src="https://github.com/user-attachments/assets/95a13f44-c9b8-4549-b10d-84b6da994db7" />

Baaaaaz:
[...]

Issue #5446, opened 2025-05-16
[Feature]: Add customisable Token Types - beyond PC/NPC


bubblobill:
### Describe the Problem

When there are a zillion objects on the map, it is impossible to find things in the Map Explorer because everything is an NPC including the trees and the rocks.

When playing team games, it is very hard to implement two or more sets of players because they are all PCs and can see everything the other guys are doing, or they can only see themselves and not their team-mates.

### The Solution you'd like

Expand the dichotomy to recognise stamps as something other than PCs. This would also make sifting through token stacks easier in the code.

Add the ability to have player groups/teams.

### Alternatives that you've considered.

There aren't any that are not a horrid kludge.

FullBleed:
When the map explorer is updated to support different token types, it would be nice if we could choose to have it sort by more than just "types"... labels and property types at least... maybe shapes and sizes, too.

kwvanderlinde:
Sounds to me like two entirely separate features are being requested here (types vs teams). Splitting them up would allow more focused discussion.

Issue #4960, opened 2024-09-30
[Feature]: It should be possible to let players see the stat sheets of NPC tokens they do not own


fishface60:
### Describe the Problem

I have been designing my stat sheets as a way to present information to players about the monsters they face.

I was surprised to learn they could not see them despite setting the token properties such that the values were visible on the stat sheet to players and not GM or Owner specific.

After digging through the code to figure out why this might be I found:

```java
if (MapTool.getPlayer().isGM()
    || AppUtil.playerOwns(token)
    || token.getType() != Type.NPC) {
  statSheet.setContent(
      event.token(),
      ssManager.getStatSheetContent(ssId),
      ssRecord.entry(),
      ssProperties.location());
}
```

It was confusing to learn that this is hard-coded after being able to have properties that are visible to non-owner and non-GM led me to believe that properties being visible to players did not extend to NPC tokens.

### The Solution you'd like

I would like stat sheets of NPCs to be visible.

Unconditionally showing them would work for me, but potentially not for others.

### Alternatives that you've considered.

1. Make all my NPC tokens into PC tokens so the stat sheets are displayed and keep the existing behaviour, though that mixes my NPC tokens in with my PC tokens in the map view.

2. Have a separate flag for whether stat sheets should be visible on NPCs, though if this is per-token that may be cumbersome.

3. Guess whether a stat sheet should be displayed by whether it has any properties.

   This has the problem of if you have a stat sheet for a token type that doesn't have properties it wouldn't be visible when you might want it to (e.g. publicly visible notes), and if the properties are accessible for other purposes such as macros then you might not want to have that cause the sheet to be displayed.

bubblobill:
@cwisniew change behaviour or add a checkbox?

cwisniew:
> @cwisniew change behaviour or add a checkbox?

I think the behaviour change is undesirable without toggling it on/off (defaulting to how it is now). It's been forever, and a day that players can't see NPC token pop-ups they don't own. I suspect there would be some real surprises for GM's if that behaviour suddenly changed. We can probably add it at the property type level (rather than individual properties) something along the lines of "Players can see NPC Stat-Sheet" option

fishface60:
I would be very happy with a checkbox in the property type to toggle this on.

timgstewart:
[...]

PR #6015, opened 2026-07-17
Permissions - TokenProperty - Stat-Sheets - Token Editor

### LINKED ISSUE #5943: [Bug]: Impossible to display stats from unowned NPC tokens on custom stat sheets
### Describe the Bug

On custom stat sheets, but not the old default stat sheet, if a token has the type NPC, and is unowned by the players, properties that are supposed to show on the stat sheet to non-owners and non-GMs will not display. This also prevents the portrait from displaying. With the default stat sheet, as long as one property is visible to players, the portrait is also visible.

### To Reproduce

Use a custom stat sheet. Designate a property as appearing on the stat sheet while leaving GM and owner unchecked. Log in with a player client that does not own the token, and mouse over the token.

### Expected Behaviour

The expected behavior is that stats designated to show to non-GM non-owners will in fact show to non-GM non-owners.

### Screenshots

_No response_

### MapTool Info

1.18.6

### Desktop

any

### Additional Context

I believe this is due to the condition in  statsheetlistener.java, 

```
 if (MapTool.getPlayer().isGM()
            || AppUtil.playerOwns(token)
            || token.getType() != Type.NPC) {
          statSheet.setContent(
```

I believe if that entire condition is deleted, then the custom stat sheet will behave correctly like the default stat sheet.

In case anyone likes for this bug to persist because they don't want to reveal NPC properties to players, they could just use an NPC property type and restrict the property to GMs, so there's really no rational basis to not fix this bug.

bubblobill:
fixes #5943 #4960 #4774 #1641 #296 #4396 #1373
progresses #1762 #6003 #5446 #5944
and probably a bunch of others.

### Description of the Change
Born out of the desire to enable the new pop-up attribuite-sheets to be shown to players for NPCs.
Previously the visibility of a property was settable through combining booleans as:
NONE -> GM(s) -> OWNER(s) -> ALL

Added a new enum Permissions and changed TokenProperty to use Permissions instead.
Permissions currently has the following entries;

- NONE
- GM
- OWNER = GMs + Owners
- ALLIES = GMs + Owners + Allies
- ALL
--
- OWNER_ONLY 
- ALLIES_ONLY
- OWNER_ALLIED_ONLY
- OPPONENTS

Allies are currently implemented simply. For example; a property with this visibility on a player's token would be visible to other players with the same GM status, i.e. Non-GM players can see this on each others tokens.
This resolves the current problem with stat-sheets not being able to show NPCs attributes. They can be set to be visible to all, or whoever is on the same team. This is more explicit and builds on the current setup.

The first batch of values are cumulative. Each level expands the scope of who has permission.
The second batch are exclusive.

For TokenProperty, Permissions have been assigned to:
- Stat-Sheet visibility,
- Token Editor visibility, and
- Token Editor editability.

Campaign Token Properties editor
image

With these permissions, you can set it so a player can see what values are, but not allow them to edit the values. This will help framework developers who require values be changed through another mechanism.
Other advantages is the ability to tailor information more contextually, such as the GM excluding themselves from seeing things they don't care about.

The Token Editor
image

This is the GM view. You can see where Defense was set as not being player visible, it now appears separately. Similarly, non-Type properties can be seen by the GM under the Misc group.

I also threw in a search bar whilst I was at it.

Future work will be the use of permissions for other aspects, like states and bars.

### Possible Drawbacks
This is a big change, and whilst I tested it on old campaigns, the testing was not thorough.

### Release Notes
- Made Stat-Sheets visible for NPCs
- Added permissions for player editing of Token properties.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/6015)


Baaaaaz:
@bubblobill Nice!  Can you include some screenshots of the `MTMultilineStringCellEditor ` in action? 

I like what you are starting do so with the different syntax types, something that I was looking at for the [input function enhancements](https://github.com/RPTools/maptool/issues/5944#issuecomment-4929944129).  I've attached below where I got up with a custom control for with that, as it handles more syntax types, parses some syntax types after typing, can toggle minifying and prettifying json, etc  Just in case it inspires you to enhance yours more (just recalling what you did to me with Halos! 😁 ) and then I can just perhaps re-use yours as you got in first!

[InputControlTextEditor.java.txt](https://github.com/user-attachments/files/30133788/InputControlTextEditor.java.txt)

What you could also do was automatic selection of the syntax type if the property name had a specific file type suffix.  That would be neat.

bubblobill:
@Baaaaaz  `MTMultilineStringCellEditor `
Is what you see when you edit a token property in the token editor. 
image

It used to be buried as a private class. I just pulled it out so it can be accessed elsewhere. 
The combo box could probably use with being extracted too.

I'm falling out of like with RSyntaxtTextArea. It's theming is not very compatible with FlatLaF.

kwvanderlinde:
I haven't gone through the code at all (it's a bit big for the time I have atm). For now I've just taken a gander at the linked issues and PR description. I want to make sure we're closing off / progressing the right ones:
- #4774 is a random fix-up PR, not related. Not sure which one was intended.
- #296 doesn't seem to be done, at least not according to the PR description.
- I'm not sure what about #5944 is being progressed.

bubblobill:
> * [Go back to calling MapTool.setCampaign() #4774](https://github.com/RPTools/maptool/pull/4774) is a random fix-up PR, not related. Not sure which one was intended.

Quite correct. Edited it to #4775. This allows fancy stat-sheets on non-player owned tokens.

> * [Expose Internal Token Properties On Edit Token Panel #296](https://github.com/RPTools/maptool/issues/296) doesn't seem to be done, at least not according to the PR description.
1. All properties set on a token are now use to populate the "Edit Token" -> "Properties" table
2. This produces 2 sets of properties; i.e.
- Those that do form part of the Token-Type Properties, and
- Those that don't.

3. Those that do form part of the Token-Type properties also belong to sub-sets based on how they have been set as player viewable & player editable. 
- Player editable enables/disables the edit field.
- Player viewable properties show as normal under the "\<Property-Type>" category.
- Non-viewable properties are only visible to the GM in the "\<Property-Type>(GM)"  category. 

4. Those that don't are only visible to the GM and are displayed under the "Misc" category.


> * I'm not sure what about [[Feature]: Enhanced input() function #5944](https://github.com/RPTools/maptool/issues/5944) is being progressed.

From comment above, making components reusable, e.g. the MultiLineStringEditor ->https://github.com/RPTools/maptool/pull/6015#issuecomment-5005869329

PR #6020, opened 2026-08-01
New function base64.encode.asset

### LINKED ISSUE #2981: Way to get raw image data of an asset
There is currently no way to get the raw or base 64 data of an asset image in a macro.

I would like for something like getAssetProperties() to be able to return a string of an image's data.
I would use this to send image data to image hosting APIs.

Tried to find a way to get the data from local files directly into a macro but there is no file input functionality in macros.
Using REST to run a python script on a server which would fetch local file data and then send to APIs, but this solution would be cumbersome and unreliable and not what I want to use.

Baaaaaz:
closes #2981

### Description of the Change
new MT script function `base64.encode.asset`

### Possible Drawbacks
Currently only does assets that are images.
No equivalent decode function provided as either natively handled in html (but webview performance on doing this sucks), or decoding could/should be handled via the `createAsset()` function instead (which may need to be adjusted to handle data URI prefixes).

### Documentation Notes
new MT script function``base64.encode.asset` for converting asset images to base64, params are:
1. asset id
2. optional true/false add data URI prefix (default=false) i.e. "data:MIME type/MIME subtype;base64,"

### Release Notes
New base64.encode.asset function

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/6020)

### DIFF
--- a/src/main/java/net/rptools/maptool/client/functions/Base64Functions.java
+++ b/src/main/java/net/rptools/maptool/client/functions/Base64Functions.java
@@ -14,16 +14,23 @@
  */
 package net.rptools.maptool.client.functions;
 
+import static net.rptools.maptool.client.functions.TokenImage.getMD5Key;
+
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 import java.util.List;
+import net.rptools.lib.MD5Key;
 import net.rptools.maptool.language.I18N;
+import net.rptools.maptool.model.Asset;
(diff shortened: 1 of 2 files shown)


kwvanderlinde (approved):
Looks good. Would be nice to extend this to all assets, but I think we need a bit of extra work first on the asset management side. E.g., have `Asset` keep track of its detected media type so it can be used in `base64.encode.asset` regardless of type.

PR #6029, opened 2026-08-07
Fix Suppress Tooltips behaviour

### LINKED ISSUE #6028: [Bug]: Suppress ToolTips for MacroLinks is broken
### Describe the Bug

Hyperlinks show the link URL as a tooltip regardless of preference.
**Suppress ToolTips for MacroLinks** acts on all links, whether they are **macro:** urls or not.
Tooltips suppressed (default) shows the URL as a tooltip. 
The fallback tooltip expects to return the title from a "span" element. With no span it returns the URL as a tooltip. Exactly the same as when it is suppressed.
HTML spec expects the "title" attirbute to be used as a tooltip. Deviating from this probably screws up screen-readers.

### To Reproduce

1. create a hyperlink with a "title" attribute
2. stick it in chat and hover over the result.
3. try it with a "lib://" URL
4. try it with a "macro://" URL
5. try it with a "http://" URL
6. try it with a "#id" URL

### Expected Behaviour

Both  "macro:" and "lib:" formats should be subject to the preference **Suppress ToolTips for MacroLinks**.
**Only**  "macro:" and "lib:" formats should be subject to the preference **Suppress ToolTips for MacroLinks**.
Before any weird \<SPAN\> workarounds, the hyperlink "title" attribute should be used as a tooltip.

### Screenshots

_No response_

### MapTool Info

_No response_

### Desktop

_No response_

### Additional Context

_No response_

bubblobill:
fixes #6028

### Description of the Change
TooltipView:
- Rearranged logic and renamed variables to avoid double negatives and make things clearer
- Added decision fork to catch "lib:" uris containig "/macro/".
- When suppressing anti-cheat tooltip, or not a macro link the hyperlink "title" attribute is thte first choice as tooltip 

i18n.properties
-  Reworded the appalling tooltip text

### Possible Drawbacks
I unwrapped the double negatives poorly.
People using span tags for tooltips that also have a title attribute on their links will get weird results.

### Documentation Notes
n/a

### Release Notes
n/a

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/6029)


src/main/java/net/rptools/maptool/client/swing/TooltipView.java
```diff
@@ -38,35 +38,56 @@ public TooltipView(Element elem, boolean macroLinkToolTips) {
...
+          // not using anti-cheat tooltip, or not a macroLink, i.e. not suppress tooltip
+          if (title != null) {
+            return title;
+          }
+        } else if (href.toLowerCase().startsWith("macro:")) {
+          // use anti-cheat tooltip, i.e. suppress normal tooltip
           return MacroLinkFunction.getInstance().macroLinkToolTip(href);
+        } else if (href.toLowerCase().startsWith("lib:")) {
+          // just show the URL
+          return href;
```
kwvanderlinde, line 77: _If_ I'm understanding the feature correctly (that's a big if), the whole point is to avoid revealing the macro link URL. So if `lib:/.../macro/...` also counts as a macro link, we should also avoid showing that URL and instead show details similar to what is done for `macro:` URLs.

bubblobill: @kwvanderlinde  You are correct. 
The FR is to fix the tooltip suppression on non-macroLink URLs and to also catch `lib:/.../macro/...` links.
Sadly the function `macroLinkToolTip` will not work on the `lib:` URL, so I had to choose between showing the link or letting it fall through. I opted for returning the link as this prioritises accessibility over presentation.

kwvanderlinde (approved):

kwvanderlinde:
FYI, I added #6050 so we can keep track of the fact that `lib://` URLs don't yet have similar tooltip formatting as `macro://` URLs.

PR #6033, opened 2026-08-15
Table groups, metadata, and editor UI memory

### LINKED ISSUE #6032: [Feature]: Tables and table editor enhancements
### Describe the Problem

1. unable to find or sort tables by something other than its name.
2. unable to store metadata relating to the table co-located with the table, without fudging it into a -1 entry row or similar.
3. the table editor always forgets that I have re-positioned or re-sized it, meaning UI monotony each time I want to re-edit a table.

### The Solution you'd like

1. Ability to create groups, assign a group to a table, and find/sort tables by said groups.
2. Ability to store additional table metadata with the table
3. Ability for the edit table dialog to remember user positioning and sizing during the session.

None.

### Additional Context

_No response_

Baaaaaz:
resolves #6032

#### EditLookupTablePanel
Added group ComboBox
image

Move table entries into a tab on a new tabbed pane
image

Added a metadata syntax editor into a tab on the new tabbed pane
image

Caches the panel from the factory, so retains the size and position when it is reopened rather than creating a new one each time from the DialogFactory.

#### LookupTableDetailPanel & Model
Added new column for Group visible to GMs and Players
Added new column for Metadata visible to GMs only - which displays the syntax type if there is metadata
image

Modified MTScript functions:
`getTableNames(delim, group)`  - new group optional parameter to get all/visible table names in a named group

New MTScript functions:
`getTableGroups()` - get groups from all (GM) /visible (Player) tables

New MTScript trusted functions:
`getTableGroup(tblName)` - get the named table's group
`setTableGroup(tblName, group)` - set the named table's group
`getTableMetadata(tblName)` - get the named table's metadata
`setTableMetadata(tblName, metadata)` - set the named table's metadata

+ associated i18n text entries

### Possible Drawbacks
Grouping only visible in the table details panel, not table image panel
Otherwise none foreseen

### Release Notes
* Table groups, metadata, and editor UI memory

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/6033)


FullBleed:
Excellent update.

I do wish there was a representation under the Table Image display, too, though.

But to make that complete I suppose we'd need to be able to set a `group image` and be able to drag-n-drop tables into the groups in the table image panel.

Perhaps in a separate PR?

Baaaaaz:
> I do wish there was a representation under the Table Image display, too, though.

Yeah I get that.  I did look into this and the Table window image panel uses a thing that is also used for the Library window image panel - so I did not want to change that within this request.  b.t.w. I was also wanting to add optional Player / GM Notes for tables to show in the table's tooltip but the current image panel does not handle that nicely so I left those out for now.

> But to make that complete I suppose we'd need to be able to set a `group image` and be able to drag-n-drop tables into the groups in the table image panel.
 
I was thinking something akin to how macro groups are shown on macro panels, i.e. a border and a label for each table group.  I would see dragging tables between groups being a GM only feature.  

> Perhaps in a separate PR?

Indeed!  Other table group enhancements could be rename a table group and to export/import a group of tables.

kwvanderlinde (changes requested):
Code changes themselves look good to me.

For the UI, I would suggest removing the "Metadata" column since:
1. it doesn't display the metadata (nor should it)
2. just the media type is not a very helpful bit of information, and so I would say it's not a good use of horizontal space in an already-lengthy table.

Baaaaaz:
Thanks for reviewing @kwvanderlinde. 

I'd like to keep the metadata type in the details view (albeit with different name as you pointed out) as the type is shown only *if* the table has metadata so it is helpful (to me at the very least!) when importing tables or during table development as users will be able to see at glance which tables have metadata (and type) without having to go hunting for this information (either opening each table in turn or writing macros to fill this gap).

I would agree that such information is less helpful when running a game but users can resize both the table window and the details view columns so it can be tailored to suit.  Even though the table is getting wider, this column has been added at the end of the GMs details view so its prominence in terms of table design is low and players do not even see this column at all.

Following this initial review I've made some other changes, the main things are:
* renamed the details view column header to "Metadata Type"
* created an Enum in LookupTable to centralize supported table metadata types
* changed the EditLookupTablePanel JComboBox to use the Enum rather than Strings
* created MTScript trusted get/set functions for a table's metadata type (with the setter validating against the Enum)
* I18N entries for MIME types are displayed in the UI (table details view and JComboBox) instead of the MIME type (reusing existing I18N keys used in the token editor notes/gmnotes panel or adding new ones if absent)

Maybe the columns displayed in the tables GM's details view could be configurable via user preferences - that could be a separate ticket or do you want it included in this PR?

PR #6036, opened 2026-08-18
Bring events together in one location

### LINKED ISSUE #6035: [Maintenance]: Make EventBus "Event" classes easier to find
### Describe the problem

Currently there are events everywhere, usually in the package they were created in to serve a related class.
It becomes nearly impossible to work out what events exist and what might be related to what you are working on.

This creates a bad user experience and extra work when a release goes out and someone complains that X isn't showing what it should.

Bring all existing events used by MapToolEventBus into one package where they will be easy to find and determine their relevance to a current project.

Alternately, add a common interface that can be applied to event classes. Coders can then search for implementations to find them all.

### Expected Benefits

Better experience for the user, the coder, and less hasty fixes post-release.

### Additional Context

_No response_

bubblobill:
closes #6035

### Description of the Change
Brought MapToolEventBus and all the events I could find into one package

### Possible Drawbacks
I may not have found them all

### Documentation Notes
n/a

### Release Notes
n/a

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/6036)


cwisniew:
No, this goes against the moves to start grouping things together. It's how the code base ended up in the mess it's in now where you have to check all over the code base to find code related to something, like states or tokens for example

bubblobill:
@cwisniew How about attatching an empty interface to all the events instead?
`x implements MTEvent`
Then you could at least go to the interface and look for implementations.

cwisniew:
> @cwisniew How about attatching an empty interface to all the events instead?
> `x implements MTEvent`
> Then you could at least go to the interface and look for implementations.

Yeah that works

PR #6037, opened 2026-08-19
Refactoring -

### LINKED ISSUE #6035: [Maintenance]: Make EventBus "Event" classes easier to find

bubblobill:
closes: #6035 

### Description of the Change
- Created new interface MTEvent.
- Implemented interface on existing events.

Extra Bonus
- Added static getter for MapToolEventBus 

### Possible Drawbacks
I may not have found them all

### Documentation Notes
n/a

### Release Notes
n/a

- - -
This change is [Reviewable](https://reviewable.io/reviews/RPTools/maptool/6037)

### DIFF
--- a/src/main/java/net/rptools/maptool/events/MapToolEventBus.java
+++ b/src/main/java/net/rptools/maptool/events/MapToolEventBus.java
@@ -25,7 +25,16 @@ public class MapToolEventBus {
   /**
    * Returns the main MapTool {@link EventBus}.
    *
-   * @return the main Maptool {@link EventBus}.
+   * @return the main MapTool {@link EventBus}.
+   */
+  public static EventBus get() {
+    return mainEventBus;
+  }
+
+  /**
+   * Returns the main MapTool {@link EventBus}.
+   *
+   * @return the main MapTool {@link EventBus}.
    */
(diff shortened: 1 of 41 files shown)


src/main/java/net/rptools/maptool/events/MapToolEventBus.java
```diff
@@ -25,7 +25,16 @@ public class MapToolEventBus {
   /**
    * Returns the main MapTool {@link EventBus}.
    *
-   * @return the main Maptool {@link EventBus}.
```
kwvanderlinde, line 28: Let's remove the `get()` method as it's redundant and never got unused.

